PR #2, opened 2015-10-18
Fix a typo in "Bootstrap customizations"

TimMoore:
Changed "you should not switch to using hooks" to "you should now switch to using hooks".


TimMoore:
(I'm assuming this is a typo at least.)

keithpitt:
Ha ha, yes it is indeed a typo. Thanks for fixing!

ALSO, OMG FIRST CONTRIBUTION!! You win the internet today @TimMoore!

pic

PR #3, opened 2015-10-21
Update ubuntu.md.erb

eminetto:
Fix path

### DIFF
--- a/pages/agent/ubuntu.md.erb
+++ b/pages/agent/ubuntu.md.erb
@@ -27,7 +27,7 @@ See the [Agent SSH Keys](/docs/agent/ssh-keys) documentation for more details.
 You can run as many parallel agents on the one machine as you wish by copying the service configuration file.
 
 ```shell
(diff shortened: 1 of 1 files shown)


keithpitt:
Hey @eminetto, thanks for the contribution! So that line there is to copy the ubuntu upstart configuration, so it'll automatically boot up another agent on startup (and allows you to run `sudo service buildkite-agent-2 start` to start another agent on the same machine). You don't need to copy/paste the `cfg` file, because each agent can use the same configuration (unless there's something specifically you need to change) - does that make sense? :)

toolmantim:
Ah, good catch @eminetto. Like Keith said, the instructions were correct but the wording was terrible. I've just updated the Ubuntu and Debian instructions in 2a2179f52f5f6a02c70387a7f0f3c327340dab3b and 438594cc214826949c824d8e2bb33f39236dc5ed to clarify! Sorry for the confusion.

PR #4, opened 2015-11-06
Removing duplicate line

ae0000:


keithpitt:
Ah! Great find! Thanks @ae0000 !

PR #9, opened 2015-11-25
Fix documentation about BUILDKITE_PULL_REQUEST

floatplane:
The docs say that it has a value of `true` when building a pull request, but it actually contains the pull request number.


keithpitt:
Ah, excellent find! Thanks @floatplane!! :dancers:

PR #13, opened 2015-12-30
Bring docker-compose syntax up to date

daveslutzkin:


pages/guides/docker_containerized_builds.md.erb
```diff
@@ -45,12 +45,12 @@ app:
     - ./:/go/src/github.com/buildkite/agent
     - /usr/bin/buildkite-agent:/usr/bin/buildkite-agent
   environment:
-    - BUILDKITE_AGENT_ACCESS_TOKEN
-    - BUILDKITE_JOB_ID
-    - BUILDKITE_BUILD_ID
-    - BUILDKITE_BUILD_NUMBER
-    - GITHUB_RELEASE_ACCESS_TOKEN
-    - DOCKER_HUB_TRIGGER_TOKEN
+    - BUILDKITE_AGENT_ACCESS_TOKEN: $BUILDKITE_AGENT_ACCESS_TOKEN
```
keithpitt: Hey @daveslutzkin, thanks for making these changes! I remember not needing to define the extra `: $...` part of a ENV if you just wanted the value to be copied to the container. Just define the name of the ENV and docker compose would do the rest, did they deprecate that functionality?

pages/guides/docker_containerized_builds.md.erb
```diff
@@ -16,9 +16,9 @@ app:
   volumes:
     - ./:/app
   links:
-    - db:db
-    - redis:redis
-    - memcache:memcache
+    - db
```
keithpitt: Ah, this is much cleaner :)

toolmantim: Nice, back in the day you'd end up with only `redis_1` in `/etc/hosts` but I just checked and they now include a non-numbered one for the first service with that name:

```bash
$ cat docker-compose.yml 
app:
  image: ubuntu
  links:
    - redis
redis:
  image: redis
$ docker-compose run app bash -c 'grep redis /etc/hosts'
172.17.0.1	redis_1 8cb1d3e600ae composetest_redis_1
172.17.0.1	composetest_redis_1 8cb1d3e600ae
172.17.0.1	redis 8cb1d3e600ae composetest_redis_1
172.17.0.1	composetest_redis_1
172.17.0.1	composetest_redis_1.bridge
```

:+1: no need for `redis:redis` anymore

pages/guides/docker_containerized_builds.md.erb
toolmantim: Pretty sure they're still supported… I couldn't see any word on the Github issues about it going away https://docs.docker.com/compose/compose-file/#environment

daveslutzkin: @toolmantim: yeah, looks like you're right. Not sure where I got that from!

keithpitt:
Nice, thanks! :dancers:

PR #15, opened 2016-01-30
Fix YML syntax for pipeline upload block step syntax

tommeier:
The example was invalid yml, now it would work :D


keithpitt:
Ah, thanks :) In in this case, the name was actually supposed to be under the `block` command. `block` commands can have their named changed so instead of saying "Continue" in the pipeline, it'll say what's in the file :)

keithpitt:
I've made this change in master now :)

PR #16, opened 2016-02-01
Fix link for Mac OS X installation documentation

irfansharif:
Previous link incorrectly pointed to debian, renamed Mac to Mac OS X for consistency with name in side panel.


keithpitt:
Ah, good catch! Thanks so much for the fix!

PR #20, opened 2016-03-01
Update the firewall install docs with nginx proxy info

toolmantim:
docs


sj26:
Nice. :+1:

keithpitt:
Yeah sweet, you smashed it!

PR #21, opened 2016-03-02
Fix nginx config

jeromegn:
My bad, I had not tested that part quite enough.

This fixes the regex.


keithpitt:
Cool, thanks for fixing that! :tada:

PR #23, opened 2016-03-24
Add CORS info to API overview

toolmantim:
I've also updated the CodePen code to be jQuery-less, and uses Github's window.fetch polyfill


keithpitt:
:+1:

PR #27, opened 2016-04-05
Parameter clarification

sj26:
There has been a little confusion around how to supply parameters to requests, so let's clarify!

I've also renamed "properties" to "parameters". `http://.../builds?state=passed` has a parameter to me, not a property. Parameters seem like a superset of properties. Thoughts?

/cc @keithpitt

### DIFF
--- a/pages/api.md.erb
+++ b/pages/api.md.erb
@@ -18,9 +18,9 @@ curl https://api.buildkite.com
 <%= present_json response: "Hello World" %>
 ```
 
-## Parameters
+## Parameters and Properties
 
-Some API methods take parameters. Some have parameters within the path of the URI, such as 'my-org' and 'my-pipeline' in the following request:
+Some API methods take parameters and/or properties. Many have parameters within the path of the URI, such as 'my-org' and 'my-pipeline' in the following request:
 
 ```bash
 curl "https://api.buildkite.com/v2/organizations/my-org/pipelines/my-pipeline/builds"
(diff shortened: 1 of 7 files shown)


sj26:
We decided not to change "properties".

pages/api.md.erb
```diff
@@ -18,6 +18,31 @@ curl https://api.buildkite.com
 <%= present_json response: "Hello World" %>
 ```
 
+## Parameters and Properties
+
+Some API methods take parameters and/or properties. Many have parameters within the path of the URI, such as 'my-org' and 'my-pipeline' in the following request:
```
toolmantim: I like the idea of this section! 

"query string parameters" is what we use to refer to query string parameters elsewhere. I looked up the HTTP language for it when I chose it back in the day… it makes it really clear that it’s the query string.

What about we split this out into two separate sections, and then link to them from the individual API docs whenever we use the word "query string parameters" and "properties"?

## Query String Parameters

Some API endpoints accept query string parameters which are added to the end of the URL. For example, the [builds listing APIs](https://buildkite.com/docs/api/builds#list-all-builds) can be filtered by `state` using the following `curl` command:

```
curl "https://api.buildkite.com/v2/organizations/my-org/pipelines/my-pipeline/builds?state=passed"```
````

## Request Body Properties

Some API requests accept JSON request bodies for specifying data. For example, the [build create API](https://buildkite.com/docs/api/builds#create-a-build) can be passed the required properties using the following `curl` command:

```
curl -X POST "https://api.buildkite.com/v2/organizations/my-org/pipelines/my-pipeline/builds`` -d '{
  "key: "value"
}'
````

The data encoding is assumed to be `application/json`. Unless explicitly stated by an API endpoint, you can not post data as `www-form-urlencoded` or `multipart/form-data`.

sj26: I like it :ok_hand:

PR #28, opened 2016-04-05
Links on how to download artifacts outside of a running build

keithpitt:
It can be a little confusing because you think you might be able to use the `buildkite-agent artifact download` command outside of a running build. This is just to have a presence on the other pages to point peeps in the right direction.


toolmantim:
Great stuff! I just did a quick fix. :rocket:

PR #30, opened 2016-04-05
Improved the pipeline documentation

toolmantim:
This splits up the pipelines documentation into two documents. One is a guide, that provides more general advice and complete examples. One is the technical docs in the agent section alongside the artifact and meta data command, and it covers the example properties that you can use to configure the pipelines.


toolmantim:
Actually, the agent yml has unreleased agent plugin stuff in it

toolmantim:
@keithpitt this has been updated! It's now two sections and I've just put an inline example yml file in there, rather than linking to GitHub

keithpitt:
👍

PR #31, opened 2016-04-08
Update the Heroku guide to support automatic deploys

toolmantim:
deploying-to-heroku


keithpitt:
Oh this is cool!

Is there something we need to mention about pull requests and it only happening on `master` or something?

toolmantim:
Cool, I've just updated it to mention Review Apps which is the pull request stuff. The branch stuff is explained in the Heroku docs… there's literally nothing they need to do in Buildkite aside from ensuring the build commit statuses are being posted.

PR #33, opened 2016-04-21
fixed 404

lukeaus:
https://buildkite.com/docs/guides/parallelizing-builds
the link for 'Agents API' points to
https://buildkite.com/docs/guides/docs/api/agents
but should point to:
https://buildkite.com/docs/api/agents


keithpitt:
Oh whoops! Thanks for catching that!

PR #37, opened 2016-05-25
Updated documentation for branch filtering, bang branch name needs to…

justinhennessy:
… use quotes


keithpitt:
Thanks for the PR! If you define the branch config via the Buildkite UI, you don't need quotes, so we may need to tweak the text here to make that clear :) But we can do that!

justinhennessy:
Of course, so would a (needs quotes if uploading a pipeline via `buildkite-agent pipeline upload`)?

keithpitt:
Yeah, that works :)

justinhennessy:
Done!

keithpitt:
Thanks!

PR #40, opened 2016-05-29
Added systemd commands to run multiple agents

deanpcmad:
Hey, here I've added some commands to run multiple agents with systemd.
I assume this is the right way to do it? It's working fine for me :)

### DIFF
--- a/pages/agent/ubuntu.md.erb
+++ b/pages/agent/ubuntu.md.erb
@@ -32,6 +32,13 @@ sudo service buildkite-agent-2 start
 sudo tail -f /var/log/upstart/buildkite-agent-2.log
 ```
 
+If you are using systemd (15.04 and above) this is different.
+
+```shell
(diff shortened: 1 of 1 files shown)


sj26:
Hiya @deanperry!

This would work! systemd generally recommends only using `/etc/systemd` for customizations on the local system, so you could copy it there instead with a new name, or even just symlink it.

I ended up doing this recently on one of our build servers and used a slightly different approach. I put both of these in `/etc/systemd/system`

https://gist.github.com/sj26/bccfac9142bf804113b4c1f4026d105a

The `buildkite-agent.service` overrides the one distributed with Buildkite and is just a central placeholder for controlling all agents at one. The `buildkite-agent@.service` lets you `systemctl start buildkite-agent@[n]` to start as many instances of the service as you like, which is what the placeholder service does.

The neat thing about having a single service tying them all together is that `systemctl restart buildkite-agent` etc will work on the whole group!

What do you think?

deanperry:
Yeah that sounds good 👍

sj26:
Consolidated into #56, thanks!

PR #44, opened 2016-07-19
Fix: change equals to not-equals sign

sander-m:
In this example, to only allow the `script/deploy` command to run, it should be `!=` instead of `==`.


keithpitt:
Ah, great catch! Thanks for this!

PR #46, opened 2016-08-05
Update examples to use systemd service@ files.

tekacs:
Only after pushing this did I notice #40, but this essentially updates the docs with a more systemd-esque way of launching multiple agents that allows running any number of agents without having to copy init files over and over.

This is a subjective change, as it might be 'best' for buildkite-agent itself to simply install a `buildkite-agent@.service` file out of the box, so that people can do this without a manual copy and the service file can be kept up to date.


sj26:
Consolidated into #56, thanks!

PR #49, opened 2016-08-30
Fix agent config env variable prefix

gkze:
@toolmantim where else would this change need to be made? Not sure...

### DIFF
--- a/pages/agent/configuration.md.erb
+++ b/pages/agent/configuration.md.erb
@@ -33,10 +33,10 @@ debug=true
 
 ## Environment Variables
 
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks for taking the time to submit this @gkze! Turns out it's kind of tricky, and they're not entirely consistent (although they should be!). I've just documented each one in 18e02116a9ab29887ab7cf5cea6db21aedc341d4 and that should be going live any minute now :) Thanks again!

gkze:
Thanks Tim!

PR #52, opened 2016-09-09
Add a guide on concurrency

toolmantim:
We don't have any guide on how to control concurrency, and @sj26 wanted to link to one from the Terraform article he's writing. So this adds a basic one that at least tells people that it's organisation wide, the syntax to use, and an example.


sj26:
Beautiful, and so speedy! Thanks!

toolmantim:
🎸

PR #55, opened 2016-10-11
Add environment variable doco

e-r-w:
Suggest adding some documentation like so around the buildkite pipeline upload process regarding environment variables in pipelines. I had to dig into the [agent source code](https://github.com/buildkite/agent/blob/master/agent/pipeline_parser_test.go#L107) to figure out what was going on, and there's a great deal of useful things you can do with the plain pipeline yaml (instead of dynamically generating the pipeline). 

For example, we're using pipelines that do things like:

``` yaml
- command: docker run -w /tmp -v $PWD:/tmp some-container
```

Which would use the working directory on the agent that originally uploaded the pipeline. If the step is not run on the same agent, this would cause docker not to volume the current directory. The solution was to escape the variables:

``` yaml
- command: docker run -w /tmp -v $$PWD:/tmp some-container
```

But this doesn't seem to be documented anywhere?

### DIFF
--- a/pages/agent/build_pipelines.md.erb
+++ b/pages/agent/build_pipelines.md.erb
@@ -74,7 +74,42 @@ A command step that uses all the possible configuration keys:
   concurrency_group: 'serial_tests'
   timeout_in_minutes: 60
 ```
+ 
+## Environment variables
 
+Environment variables are automatically interpolated when you upload your build, for example:
+      
+```yml
+- command: echo 'building ${BUILDKITE_PIPELINE_SLUG}...'      
+```
+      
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks for the contribution, and the well-written docs! Environment variable interpolation isn't available in the 2.x agent, so once 3.x is released as stable we'll be able to merge this one in.

e-r-w:
Didn't know that wasn't in the 2.x agents, that makes a lot of sense now. Thanks!

lox:
I'm going to close this one as it's now included in our v3 docs.

PR #56, opened 2016-10-14
Recommend systemd templates & instances

sj26:
Consolidate #40 and #46 with the units we currently ship with the agent. We'll probably start shipping a template like buildkite/agent#363 and start recommending that eventually.

/cc @keithpitt


keithpitt:
Legit 👍

PR #59, opened 2016-11-04
Updated the compose variables in the uploading pipeline docs.

wolfeidau:
Couple of small updates based on slack discussion and my misadventures with docker compose.


keithpitt:
Oh right! Yeah we changed that a while ago. Thanks for the heads up!

PR #63, opened 2016-11-13
Update Build Location for buildkite-agent

grepme:
I'm not positive if this is the default file location for builds across all Linux distributions, but it was certainly the case for Ubuntu:
```
ubuntu@ciders:/var/lib/buildkite-agent$ buildkite-agent -v
buildkite-agent version 2.3, build 1382
ubuntu@ciders:/var/lib/buildkite-agent$ ls
builds
```

Any feedback is appreciated.

### DIFF
--- a/pages/agent/_apt_locations.md.erb
+++ b/pages/agent/_apt_locations.md.erb
@@ -1,6 +1,6 @@
 * Configuration: `/etc/buildkite-agent/buildkite-agent.cfg`
 * Hooks: `/etc/buildkite-agent/hooks/`
-* Builds: `/var/buildkite-agent/builds/`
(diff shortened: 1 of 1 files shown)


keithpitt:
Ah yes! In both our 3.0 beta and the 2.3 stable, both of the locations are set to `/var/lib`

https://github.com/buildkite/agent/blob/master/packaging/linux/root/usr/share/buildkite-agent/buildkite-agent.cfg#L35
https://github.com/buildkite/agent/blob/2-3-stable/packaging/linux/root/usr/share/buildkite-agent/buildkite-agent.cfg#L25

Thanks for making the fix! Looks like it was a typo on our end. Cheers!

PR #65, opened 2016-11-18
Docs for unblock prompt

harrietgrace:
Updating the docs for the unblock prompt... putting things in tables to see how it looks. 

Turns out that tables are great! Also added a link to the sample pipeline for unblock prompts.


harrietgrace:
## Before:
unblock-docs-before

## After
unblock-docs-after

pages/agent/cli_pipeline.md.erb
```diff
@@ -115,49 +115,103 @@ available for `master` builds:
...
+<table>
+  <tr>
+    <th>Attribute</th>
+    <th>Example</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>name</td>
+    <td>Release \:rocket\:</td>
+    <td>the text that is displayed for the step in the pipeline</td>
```
keithpitt: Should this start with a capital letter? /shrug

pages/agent/cli_pipeline.md.erb
```diff
@@ -115,49 +115,103 @@ available for `master` builds:
...
+  </tr>
+  <tr>
+    <td>branches</td>
+    <td>master</td>
+    <td>
+      scope the block step to a set of branches <br>
```
harrietgrace: Yeah, it looks awkward... I'm going to change it to just be one. What's the syntax for doing more than one branch? Just a comma?

pages/agent/cli_pipeline.md.erb
harrietgrace: I was thinking that... it was more of a visual choice not to. But it feels weird without. I'll put them back and see how it looks.

pages/agent/cli_pipeline.md.erb
keithpitt: Spaces - we've got a doc you can link to: https://buildkite.com/docs/pipelines/branch-configuration

harrietgrace:
The latest main change was updating this item:
screen shot 2016-12-08 at 10 49 18 am

Also added full stops to the end of all the table descriptions.

keithpitt:
Ah coolio! We also need to change `optional: true` to `required: false`

harrietgrace:
Yep! Did that :smile:

pages/agent/cli_pipeline.md.erb
```diff
@@ -115,49 +115,102 @@ available for `master` builds:
...
-  prompt: "Fill out this form!"
-  fields: 
-    - select: "What team should be notified after this deployment?"
-      key: "notify-team"
+    - text: Release notes
+      key: release-notes
+      hint: Enter your release notes here \:sparkles\:
+    - text: Release name
+      key: release-name
+      optional: true
```
keithpitt: Forgot one here ;)

harrietgrace: 💥👌🏻 many thanks!

toolmantim:
Ok… here's a few improvements I think you be good to see before this goes out. Is this the best place for it?

---

This page in the docs still feels really hard to find? We have a "Pipelines" section in the docs now, but you have to go to "Uploading Pipelines" and then "Further Documentation". And then you're deep within the cli documentation. Would it be crazy to bring all the different step types into their own docs files, living under Pipelines?

step-types

This might also help to fix these Click to Unblock sub-sections being awkwardly the same level as Command Steps and Waiter Steps:

toc-2

And if they have their own page, we can put example screenshots, link to new step types from announcements (like the upcoming "trigger" step), etc.

Looking at the "build attributes" and "fields attributes" in detail, I can see that documenting this type of YML config is HARD. Stuff like the fields example being the word "text":

fields-text

There's probably ways of structuring this we can steal from the API docs, but only if these are all on their own pages. I think it'd be too big if we tried that on this one page.

Does any of that sound like a good idea? I can help on the page content tomorrow.

PR #74, opened 2016-12-23
Fix a typo in webhooks.md.erb

maria-v:


keithpitt:
Oh yeah, good catch! I probably had too many ☕️'s that day and was smashing the keyboard :)

Thanks so much for the PR!

PR #75, opened 2016-12-27
Add "blocked" webhook event documentation

maschwenk:
Not sure where to edit the `webhook_description` but I believe this is now part of the officially supported webhooks


keithpitt:
@maschwenk thanks for the PR! We haven't added `build.blocked` to webhooks just yet, but you're right - we totally need to add it.

We haven't done it yet since we need to push out a new version of our webhooks out (to add `blocked` as a state to builds and jobs and as an event) but we haven't gotten around to it yet. We've got a PR internally to finish this off in the new year though!

PR #76, opened 2017-01-05
Add command and wait step documentation

harrietgrace:
- [x]  Command step bit
- [x]  Wait step bit
- [x] @toolmantim makes the Troubleshooting bit pretty 
- [x]  Screenshots!
- [x]  Concurrency doc update
- [x]  Glob path pattern addition 
- [x]  Review!
- [ ]  🚀


harrietgrace:
There wasn't anything to say about the wait step really... so I just added a bit more to the intro sentences. It's a little repetitive, so I might remove it later 💁🏻

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
 
 <%= toc %>
 
-A *command* step runs a command on an agent. The following is the most simple *command* step:
+
+A *command* step runs a command on an agent. The command can be the name of a script, or <cli things>
```
toolmantim: Might need to axe `The command can be the name of a script, or <cli things>`?

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+
+## Command Step Attributes
+
+_Optional attributes:_
+
+<table>
+  <tr>
+    <td><code>name</code></td>
+    <td>
+      The text and/or emoji that will be displayed in the pipeline visualisation in Buildkite.<br>
```
toolmantim: I wonder if we should use `text and/or emoji` elsewhere too? Or perhaps we add "(supports emoji)" to any pipeline.yml field that supports emoji, and then use a word like "label"?

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+      <em>Example:</em> <code>"\:hammer\: Tests"</code><br>
+    </td>
+  </tr>
+  <tr>
+    <td><code>branches</code></td>
+    <td>
+      The <a href="/docs/pipelines/branch-configuration#branch-pattern-examples">branch pattern</a> defining which branches will include this command step in their builds.<br>
```
toolmantim: This line blew my brain a little bit. Would `The branch pattern defining which branches will include this step` be too brief? Or perhaps just removing the word "command" might be enough?

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+  </tr>
+  <tr>
+    <td><code>agents</code></td>
+    <td>
+      A list of <a href="/docs/agent/cli-meta-data">meta-data</a> key value pairs to <a href="docs/agent/cli-start#agent-targeting">target specific agents</a> for this step. <br>
+      <em>Example:</em> <code>npm: "true"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>plugins</code></td>
```
toolmantim: Plugins are in beta and aren't publicly released yet, which is why there's no page to link to yet. Let's leave this off!

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+    <td>
+      A list of plugins that you require for this step. <br>
+      <em>Example:</em> <code>docker-compose</code><br>
+      <em>Example:</em> <code>harrietgl/myplugin</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>artifact_paths</code></td>
+    <td>
+      The path or paths where <a href="/docs/agent/cli-artifact">artifacts</a> from this step will be uploaded. <br>
```
toolmantim: Just a heads up on this one, as it's definitely a quirky design. It's a list of glob path patterns, but it's separated using bash semi-colons rather than a normal YAML list.

Really we should be able to specify a list here… such as:

```
artifact_paths:
  - "logs/**/*"
  - "coverage/**/"
```

For the pattern syntax itself, I don't think we mention "glob path patterns" anywhere in our docs yet really, aside from the glob property in the [Artifacts API docs](https://buildkite.com/docs/rest-api/artifacts). Would linking to the [Glob wikipedia page](https://en.wikipedia.org/wiki/Glob_(programming)) be a bad idea? Behind the scenes it uses [Golang's Glob/Match function](https://golang.org/pkg/path/filepath/#Match) which has terrible docs itself! Ugh.

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+      <em>Example:</em> <code>"logs/**/*;coverage/**/*"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>parallelism</code></td>
+    <td>
+      The number of jobs that will run this step <a href="/docs/builds/parallelizing-builds#job-parallelism">simultaneously</a>. <br>
```
toolmantim: This is a tricky one! Because steps aren't "run", but jobs are.

Some rough ideas…

`The number of <a href="/docs/builds/parallelizing-builds#job-parallelism">parallel jobs</a> that will be created based on this step`

Or something around the "number of copies"

Hrmm. Maybe the original is good!

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+    <td><code>concurrency</code></td>
+    <td>
+      The <a href="/docs/builds/controlling-concurrency#concurrency-limits">maximum number of jobs</a> that can be run inside this concurrency group at the same time as this job. <br>
+      <em>Example:</em> <code>1</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>concurrency_group</code></td>
+    <td>
+      A nickname for the concurrency value that you set with the <code>concurrency</code> attribute. This can also be the name of an existing <a href="/docs/builds/controlling-concurrency#concurrency-groups">concurrency group</a>. <br>
```
toolmantim: There's a few quirks with this one that are probably worth documenting.

The concurrency value is organization wide, and I think people don't realise this because it's defined in a specific pipelines's pipeline.yml. So `deploy` in one pipeline will block a `deploy` in another pipeline.

An idea…

> The name of the group used to apply <a href="/docs/builds/controlling-concurrency#concurrency-groups">concurrency limits</a> when using the `concurrency` attribute. (something about it being org wide)

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,113 @@
...
+      <em>Example:</em> <code>"serial_tests"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>timeout_in_minutes</code></td>
+    <td>
+      The number of minutes this step should wait for an agent before timing out. When the step hits this timeout without finding an agent, the step will fail. <br>
```
toolmantim: The way this one works is…

> The number of minutes the job is allowed to run. If the job does not finish within the timeout limit, it will be automatically cancelled and the build will fail.

pages/pipelines/wait_step.md.erb
```diff
@@ -2,7 +2,11 @@
 
 <%= toc %>
 
-A *waiter* step waits for all previous steps to complete before continuing, for example:
+A *wait* step waits for all previous steps to complete before allowing following jobs to continue.
```
toolmantim: Based on #77 it might be worth saying "to have successfully completed" here?

pages/pipelines/command_step.md.erb
harrietgrace: The <cli things> bit was waiting for more info - I wasn't really sure what to call it. You can put commands in there that you would normally run on the command line, right? So like, you could cd somewhere and then do some things? I wanted to mention that you could do that there and that it doesn't have to just be the name of a script.

pages/pipelines/command_step.md.erb
harrietgrace: Sounds like a plan. I was hesitant about label for a bit, but I'm coming around to it. I wanted to be super clear that it can be just emoji, or just text, or both, which I wasn't sure was super clear if we just say label? But I'm keen to test drive it and see how it goes.

pages/pipelines/command_step.md.erb
toolmantim: What about something like?

> A *command* step runs one or more shell commands on an agent. Each command can be a shell command, such as `npm install`, or an executable file or script, such as `build.sh`

Unfortunately you can't use a normal YAML "list" (like we support for a lot of other things):

```yaml
steps:
  - name: ":hammer:"
    command:
      - "build.sh"
      - "test.sh"
```

Instead you have to do one of these:

```yaml
steps:
  - name: ":hammer:"
    command: "build.sh && test.sh"
  - name: ":hammer:"
    command: "build.sh\ntest.sh"
```

But that's a bit sucky! Let me see if I can do a sneaky pull request this week to support the proper YAML list syntax?

Should we also support `commands` as a pseudonym as well?

```yaml
steps:
  - name: ":hammer:"
    commands:
      - "build.sh"
      - "test.sh"
```

pages/pipelines/command_step.md.erb
toolmantim: Cool bananas!

pages/pipelines/command_step.md.erb
harrietgrace: Do we want to support that list syntax before shipping this? This doc update isn't urgent.
We could add our own version of that glob description to our artifact page? And then link to that?

toolmantim: > Do we want to support that list syntax before shipping this?

I'll have a look at it when I do the command list stuff this afternoon!

> We could add our own version of that glob description to our artifact page? And then link to that?

Yeah!

pages/pipelines/command_step.md.erb
harrietgrace: Considering that we don't have a 'cancelled' state, I'd like to avoid using that word... but I can't come up with anything better haha. Maybe something like 

If the job does not finish within this time limit
it will automatically fail
it will fail, therefore failing the build
it will fail, which causes the build to fail
the job will be marked as failed

Or something

pages/pipelines/command_step.md.erb
harrietgrace: I think people are probably fine with using && in there, but it would be nice to support the list as well. And yeah, allowing plural commands would be great too!

pages/pipelines/command_step.md.erb
toolmantim: Ah, jobs do have `cancelling` and `canceled` states, but that's for when you hit the "Cancel" button in the web UI. And it causes the agent to terminate the running command on the agent machine.

I thought this was how timeouts worked on the agent machine too… hence why I wanted to mention it. Because people will probably want to know "What happens to the running job when the timeout is reached?"

I thought if the command times out, the agent will forcefully terminate the running command on the agent machine, the job will then be marked as `timed_out`, and the build will be `failed`.

But after looking at the code, my assumptions are actually wrong :( The agent never terminates the job on the machine when the timeout is reached. The web UI marks the job as `timed_out` and the agent goes along it's merry way, still running the job. I think this might be a bug.

I might need @sj26’s input on this one.

toolmantim: @sj26 the relevant lines: https://github.com/buildkite/agent/blob/0d579d308c59a4d09e94e3cbd9bbb834fcc5970b/agent/job_runner.go#L276 https://github.com/buildkite/buildkite/blob/master/app/models/queue_manager/job_state_expirer.rb#L69 https://github.com/buildkite/buildkite/blob/master/app/models/queue_manager/job_state_expirer.rb#L97

pages/pipelines/command_step.md.erb
sj26: The pattern I've seen for multi commands is a yaml block string:

```yaml
steps:
  - label: ":hammer:"
    command: |
      build.sh
      test.sh
```

pages/pipelines/command_step.md.erb
sj26: @toolmantim Oh my, I think you're right! That job status updater should be looking for the timing_out state and actioning it. I'll look into it.

pages/pipelines/command_step.md.erb
toolmantim: Ok, this change is now live! You can use YML arrays/lists for commands 🎉

pages/pipelines/command_step.md.erb
toolmantim: Ok, this change is going live now! You can use YML arrays/lists for `artifact_paths` 🎉

pages/pipelines/command_step.md.erb
sj26: Ah, no, it's translated into cancelled here:

https://github.com/buildkite/buildkite/blob/master/app/controllers/agent/jobs_controller.rb#L43-L49

pages/pipelines/command_step.md.erb
harrietgrace: Do we want to be encouraging people to reuse existing groups here? Or to create new ones? I assume it's the former - if so I might put the group attribute first, and the concurrency value one second. Then we can talk about existing group names first, then about how to create new ones.

Akkuma:
What does `npm: true` do exactly? I understand that `queue` targets a specific agent with that particular name, but haven't seen anything explaining what this would do.

ticky:
@Akkuma it means that job will target agents which have a metadata key of `npm` whose value is `true` - `queue` is not actually special, it’s just the metadata key we tend to use for this kind of targeting! That metadata is set with the `meta-data` flag on the [agent command line](https://buildkite.com/docs/agent/configuration#configuration-settings), or `BUILDKITE_AGENT_META_DATA` environment variable.

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,61 @@
...
+
+B) You want to run one step multiple times, and are using the `parallelism` attribute to create 5 duplicate jobs each time a build is created. These jobs access a shared resource, so you don't want them to run at the same time. As a concurrency limit of 1 was set on the step, these jobs will run one after each other, even if there are agents available to run them all in parallel.
+
+<%= image "concurrency-limits.png", width: 1092/2, height: 214/2, alt: "Screenshot of a step's Global Concurrency Limit field" %>
+
+You can add concurrency limits to steps either through Buildkite, or your `pipeline.yml` file. Defining concurrency in a `pipeline.yml` file also requires the use of the `concurrency-group` attribute. See [Concurrency Groups](/docs/builds/controlling-concurrency#concurrency-groups) for more information.
+
+<section class="Docs__troubleshooting-note">
+    <h1>I'm getting ___ error when my pipeline upload runs</h1>
```
keithpitt: I think this would be "Missing concurrency_group_id for ..."

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,61 @@
 
 Some tasks need to be run with very strict concurrency rules to ensure they don’t collide with each other. Common examples for needing concurrency control are deployments, app releases and infrastructure tasks.
 
-To help you control concurrency Buildkite provides two primitives: concurrency groups, and concurrency limits.
+To help you control concurrency Buildkite provides two primitives: concurrency limits, and concurrency groups. While these two primitives are interdependant, they are not explicitly linked.
```
keithpitt: Should the commas here be:

To help you control concurrency**,** Buildkite provides two primitives: concurrency limits[removed] and concurrency groups. While these two primitives are interdependant, they are not explicitly linked.


Note: I failed english pretty hard.

harrietgrace: Defs yes to the extra one, but I'm a pretty intense fan of the oxford comma, so the one before the 'and' is sticking around.

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,61 @@
...
 <%= toc %>
 
+## Concurrency Limits
+
+Concurrency limits define the number of jobs that are allowed to be running at any one time. These limits are set per-step and only apply to jobs that are based on that step. 
+
+```yml
+steps:
+- command: "tests.sh"
+  concurrency: 1
```
keithpitt: This example will blow up with the "Missing concurrency_group_id for ..." error, maybe it needs a concurrency group?

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+    <td><code>parallelism</code></td>
+    <td>
+      The number of <a href="/docs/builds/parallelizing-builds#job-parallelism">parallel jobs</a> that will be created based on this step. <br>
+      <em>Example:</em> <code>3</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>concurrency</code></td>
+    <td>
+      The <a href="/docs/builds/controlling-concurrency#concurrency-limits">maximum number of jobs</a> created from this step that are allowed to run at the same time. If you use a concurrency limit, you must also define a label for it in the concurrency-group attribute.<br>
```
keithpitt: should be `concurrency_group` (no dash)

keithpitt: should be `concurrency_group` (no dash)

keithpitt:
Probably should mention dynamic concurrency groups as well

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,47 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+- there are two wildcards available that match non-separator characters:
+  + `*` to match a sequence of characters
+  + `?` to match a single character
+- character ranges surrounded by `[]` support the `^` as a negator
+- special characters can be escaped with `\\`
+- multiple paths are separated with `;`
+- surround the pattern with quotes
+
+The source path of your artifact is the one that will be used at the destination as well. If you just want the item, and not the full path name, make sure you `cd` into the directory containing your artifact before running the upload.
```
keithpitt: Here's some rando examples:

```
# Upload a specific file
buildkite-agent artifact upload "log/test.log"

# To upload all the jpegs and pngs, in all folders and subfolders
buildkite-agent artifact upload "*/**/*.jpg;*/**/*.jpeg;*/**/*.png"

# Upload all the log files in the log folder
buildkite-agent artifact upload "log/*.log"

# Upload all the files and folders inside the `coverage` directory
buildkite-agent artifact upload "coverage/**/*"
```

keithpitt:
```
# To upload a file names, with special characters: e.g "hello??.html"
buildkite-agent artifact upload "hello\?\?.html"
```

keithpitt: When you run 

```
buildkite-agent artifact upload "log/test.log"
```

Buildkite will show the file as `log/test.log` in the user interface. If you'd rather it be labeled as just "test.log", then you'll need to change the working directory of the `buildkite-agent` before running upload like so:

```
cd log
buildkite-agent artifact upload "test.log"
```

harrietgrace:
@keithpitt I've added some more bits and all your examples - could you please take another look at this? 😊 There's the concurrency page itself, the concurrency bits of the command step doc, and the glob path section.

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
 
 Some tasks need to be run with very strict concurrency rules to ensure they don’t collide with each other. Common examples for needing concurrency control are deployments, app releases and infrastructure tasks.
 
-To help you control concurrency Buildkite provides two primitives: concurrency groups, and concurrency limits.
+To help you control concurrency, Buildkite provides two primitives: concurrency limits and concurrency groups. While these two primitives are interdependant, they are not explicitly linked.
```
keithpitt: But they are linked! At the hip even!

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
 
 <%= toc %>
 
+## Concurrency Limits
+
+Concurrency limits define the number of jobs that are allowed to be running at any one time. These limits are set per-step and only apply to jobs that are based on that step. 
+
+For example:
+
+You have set a concurrency limit of `1` on the first step in your pipeline, to ensures that no two jobs created from that step will run at the same time. You'll find this affects a couple of different scenarios:
```
keithpitt: "to ensures" - I think that should be "to ensure"

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
+
+A) Two builds are started on the same pipeline and are running in parallel. This means that in the pool of started jobs, there are two that have been created from the same step - one in each of the running builds. As a concurrency limit of 1 was set on the step, these jobs will not run at the same time, even if there are agents available to do so.
+
+B) You want to run one step multiple times, and are using the `parallelism` attribute to create 5 duplicate jobs each time a build is created. These jobs access a shared resource, so you don't want them to run at the same time. As a concurrency limit of 1 was set on the step, these jobs will run one after each other, even if there are agents available to run them all in parallel.
+
+<%= image "concurrency-limits.png", width: 1092/2, height: 214/2, alt: "Screenshot of a step's Global Concurrency Limit field" %>
+
+You can add concurrency limits to steps either through Buildkite, or your `pipeline.yml` file. Defining concurrency in a `pipeline.yml` file also requires the use of the `concurrency-group` attribute. See [Concurrency Groups](/docs/builds/controlling-concurrency#concurrency-groups) for more information.
```
keithpitt: Should be "concurrency_group" (no dash, underscore party)

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
+
+<section class="Docs__troubleshooting-note">
+    <h1>I'm seeing an error about a missing concurrency_group_id when I run my pipeline upload</h1>
+    <p>This error is caused by a missing concurrency group label. If you set a concurrency limit in a pipeline.yml file, you must also set a group label.</p>
```
keithpitt: I think `pipeline.yml` file here should be in a code block since it was before.

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
+
+Concurrency groups guarantee that jobs will be run in the order that they were created in. Jobs inherit the creation time of their parent. Parents of jobs can be either a build or a pipeline upload job. As pipeline uploads add more jobs to the build after it has started, the jobs that they add will inherit the creation time of the pipeline upload rather than the build.  
+
+If you have multiple pipeline uploaders within a single pipeline that each contain steps with the same concurrency group: Buildkite doesn't know about the steps inside a pipeline upload until it's running and the jobs are created. If one upload runs more quickly than another, its jobs may start running before the rest have even been created. If there are steps that rely on correct ordering, take advantage of [wait](/docs/pipelines/wait-step) or [block](/docs/pipelines/block-step) steps so that you can be sure that they will run correctly each time.
```
keithpitt: I have a memory of us talking in person about ripping this paragraph out...or I may be making it up in my head.

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
+
+<drawing/gif example here>
```
keithpitt: Such drawing

pages/builds/controlling_concurrency.md.erb
```diff
@@ -2,27 +2,59 @@
...
 
-Note: you can’t set a custom concurrency group label using the web interface, only a `pipeline.yml`. If you set concurrency in the web interface a label is automatically generated for that step.
-
-## Concurrency Limits
-
-Concurrency limits define the number of build jobs that are allowed to be running at any one time.
-
-For example, a concurrency limit of `1` ensures that no two jobs with the same concurrency group will run at the same time.
+<section class="Docs__note">
+	<p>When you create a step using the Buildkite web interface, you can’t set a custom concurrency group label. A label will be automatically generated for you based on that step. If you want to use group labels, you can define your steps in a pipeline.yml file. </p>
```
keithpitt: `pipeline.yml` should be in code blocks here as well

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+      - "npm install && npm test" 
+      - "moretests.sh"
+      - "build.sh"
+```
+
+_Optional attributes:_
+
+<table>
+  <tr>
+    <td><code>name</code></td>
```
keithpitt: We prefer `label`

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+    <td>
+      The <a href="/docs/pipelines/branch-configuration#branch-pattern-examples">branch pattern</a> defining which branches will include this step in their builds.<br>
+      <em>Example:</em> <code>"master stable/*"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>env</code></td>
+    <td>
+      A list of <a href="/docs/builds/environment-variables">environment variables</a> for this step. <br>
+      <em>Example:</em> <code>FOO: "bar"</code>
```
keithpitt: This should probably include a multi-line example to showcase the format the ENV vars take:

```
steps:
  - command: "test.sh"
    env:
      RAILS_ENV: "test"
      RACK_ENV: "test"
```

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+
+## Command Step Attributes
+
+_Required attributes:_
+
+<table>
+  <tr>
+    <td><code>command</code></td>
+    <td>
+      The shell command/s to run during this step. This can be a single line of commands, or a yaml list. <br>
```
keithpitt: yaml = YAML

keithpitt: (as in, it should be all capitals)

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+    <td><code>agents</code></td>
+    <td>
+      A list of <a href="/docs/agent/cli-meta-data">meta-data</a> key value pairs to <a href="/docs/agent/cli-start#agent-targeting">target specific agents</a> for this step. <br> 
+      <em>Example:</em> <code>npm: "true"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>artifact_paths</code></td>
+    <td>
+      The <a href="/docs/rest-api/artifacts#glob-path-pattern">glob path</a> or paths where <a href="/docs/agent/cli-artifact">artifacts</a> from this step will be uploaded. This can be a single line of paths separated by semicolons, or a yaml list.<br>
```
keithpitt: YAML CAPITAL PARTY HERE AS WELL

pages/pipelines/command_step.md.erb
```diff
@@ -2,26 +2,142 @@
...
+steps:
+  - name: "\:hammer\: Tests"
+    command: "npm install && tests.sh"
+    branches: "master"
+    env:
+      FOO: "bar"
+    agents:
+      npm: "true"
+      queue: "tests"
+    plugins:
```
keithpitt: Perhaps `plugins` should be removed from this example since it's not officially out yet?

pages/pipelines/wait_step.md.erb
```diff
@@ -2,7 +2,11 @@
 
 <%= toc %>
 
-A *waiter* step waits for all previous steps to complete before continuing, for example:
+A *wait* step waits for all previous steps to have successfully completed before allowing following jobs to continue.
+
+<%= image "wait-step.png", width: 944/2, height: 364/2, alt: "Screenshot of a basic wait step" %>
```
keithpitt: This image looks to be borked / not there!

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,90 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+- there are two wildcards available that match non-separator characters:
+  + `*` to match a sequence of characters
+  + `?` to match a single character
+- character ranges surrounded by `[]` support the `^` as a negator
+- special characters can be escaped with `\\`
+- multiple paths are separated with `;`
+- surround the pattern with quotes
+
+The source path of your artifact is the one that will be used at the destination as well. If you just want the item, and not the full path name, make sure you `cd` into the directory containing your artifact before running the upload. When you run:
```
keithpitt: This sentence lost me a little :P

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,90 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+```bash
+buildkite-agent artifact upload "log/test.log"
+```
+
+Buildkite will show the file as `log/test.log` in the user interface. If you'd rather it be labeled as just `test.log`, then you'll need to change the working directory of the buildkite-agent before running upload like so:
```
keithpitt: Oh right, you explain the `cd` bit here. You could probably tweak the previous paragraph, or just get rid of it all together!

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,90 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+Uploading a file name with special characters (e.g hello??.html)
+
+```bash
+buildkite-agent artifact upload "hello\?\?.html"
+```
+
+### Downloading artifacts
+
+The path patterns for downloading artifacts are a little different, as they are translated into SQL LIKE statements to query our database. To enable this translation, the path pattern only supports text matching and the single wildcard syntax `*`.
```
keithpitt: I'd probably not mention the SQL gear, we may change how we do it in the future:

> The path patterns for downloading artifacts are different to uploads. Buildkite only supports a `*` wildcard character something something.

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,90 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+### Downloading artifacts
+
+The path patterns for downloading artifacts are a little different, as they are translated into SQL LIKE statements to query our database. To enable this translation, the path pattern only supports text matching and the single wildcard syntax `*`. 
+
+e.g. The path `foo/*/bar/hello.txt` would be translated to `foo/%/bar/hello.txt` on the backend to query our artifact database.
+
+If you don't want the full path of your download source to be replicated in your destination directory, use wildcards rather that specific directory names in your source path. This will replicate unix-style file access.
+
+Path merging will occur if you have the same directory names in the download source and destination paths.
```
keithpitt: Good explanation!

pages/rest_api/artifacts.md.erb
```diff
@@ -24,6 +24,90 @@ An artifact is a file uploaded by your agent during the execution of a build’s
...
+
+e.g. The path `foo/*/bar/hello.txt` would be translated to `foo/%/bar/hello.txt` on the backend to query our artifact database.
+
+If you don't want the full path of your download source to be replicated in your destination directory, use wildcards rather that specific directory names in your source path. This will replicate unix-style file access.
+
+Path merging will occur if you have the same directory names in the download source and destination paths. 
+
+e.g. If you're downloading `dist/my-file.txt` to `Desktop` then it will create a directory called `dist` and your file will end up in `Desktop/dist/my-file.txt`. However, if you are downloading `dist/my-file.txt` to `dist`, then the downloader will merge those two path names and your file will end up in `dist/my-file.txt` on your destination machine.  
+
+<section class="Docs__troubleshooting-note">
```
keithpitt: I think this should be in the "Uploading artifacts" part of the page?

keithpitt:
@harrietgrace just did another 👁👁

pages/builds/controlling_concurrency.md.erb
harrietgrace: I feel like we still need to mention something about this case... right? Or is it no longer a thing that can happen anymore? I remember you pushed a fix for something, but I can't remember what 😅

pages/pipelines/command_step.md.erb
harrietgrace: There's the full example at the end... do you think we need another inline?

keithpitt: Oh yeah, I made that comment without reading the rest of the page. 👍

PR #77, opened 2017-01-05
Explain that a step considered "complete" when exit code is 0

koresar:
I hope I understand the "step completeness" correctly. If not, please explain what's a "complete" step.
Thanks!


toolmantim:
Thanks @koresar for taking the time to submit a PR… those docs about step completeness are definitely confusing. We'll get this fixed up in #76 for you!

PR #82, opened 2017-01-23
Agent actually supports EC2 config as env vars

diist:
See https://github.com/buildkite/agent/blob/master/clicommand/agent_start.go#L116-L124


keithpitt:
Oh yeah, good find. Thanks for that! 🚀

PR #87, opened 2017-02-06
Improve Writing Build Scripts docs

harrietgrace:
Improves the [Writing Build Scripts](https://buildkite.com/docs/builds/writing-build-scripts) documentation.


toolmantim:
@harrietgrace I gave this a big once-over too! I reckon this is pretty much ready to 🚀 whenever you’re ready.

pages/builds/writing_build_scripts.md.erb
```diff
@@ -1,90 +1,140 @@
...
-Most build scripts should set the `-e` and `-o pipefail` options to ensure if any command fails your entire build script fails. For example:
+Bash’s built-in `set` command with a `-` argument is used to enable options (the `+` argument is used to disable options).
+
+For example, the following script enables the most commonly used options for build scripts:
 
 ```bash
 #!/bin/bash
 
-set -eo pipefail
+set -euo pipefail
```
toolmantim: @lox @sj26 @keithpitt @ticky is this too controversial a suggestion? I know the `set -u` option has caused us grief in the past because quite a few other build tools (rvm comes to mind) blow up with this option enabled. But it's super handy, and I use this as my default all the time.

ticky: Maybe documenting that caveat would be worthwhile? Something like

> **A note about Bash options**
> In these examples, we set three Bash options. These are;
>* `-e` (`errexit`), which causes the script to exit if it encounters any error in execution,
>* `pipefail`, which causes errors in piped commands not to be swallowed, and
>* `-u` (`nounset`), which treats the dangerous operation of substituting an unset variable as an error
>
>Note that while `nounset` is useful, it can cause issues with some tools like `rvm`, which rely on the default Bash behaviour. If you encounter unexpected errors with this, remove the `u` from the `set` command in your script.

Additionally, I’m always a fan of fully qualified options, is there a reason not to use these options by name rather than (hard to search for!) flag?

toolmantim: Thanks for the feedback @ticky!

That note for `-u` is really good, I've added that in. I've also included the exact error message ("unbound variable") that Bash will output, to help people googling "Buildkite rvm unbound variable error" type thing.

unbound-variable-error

As for using the shorthand or longhand version of the options, in general I'd agree! I tried it on for size in the docs, updating the table and the recommended default (`set -o errexit -o nounset -o pipefail`) but it felt wrong (most people don't use those versions), and the default combo felt unmemorable (`set -euo pipefail` vs `set -o errexit -o nounset -o pipefail`). It seems it's more defecto standard to use the shorter ones, similar to `ls -la`. I think I'll leave it as the short one for now.

pages/builds/writing_build_scripts.md.erb
```diff
@@ -1,90 +1,145 @@
...
 #!/bin/bash
 
-set -x
+echo "$PATH"
+
+some_command
 ```
 
-## Capturing Exit Status
+As a last resort, you can execute `env` to print out all the environment variable names and their values. If you use `env` you should ensure you don't leak private keys or other information, by filtering the output using a tool such as `grep` (or `egrep`).
```
harrietgrace: Maybe something like 'If you require more environment information' or something rather than 'as a last resort'? It sounds like they'll have no other options if they can't solve their problem after executing 'env'. I think the second sentence conveys the 'this isn't the best option' idea that I guess you were going for with 'as a last resort'?

pages/builds/writing_build_scripts.md.erb
```diff
@@ -1,90 +1,145 @@
...
-    <th><code>-x</code></th>
-    <td>Print each expanded command before executing (useful for debugging)</td>
+    <th><code>x</code></th>
+    <td>Expand and print each command before executing. See <a href="/docs/builds/writing-build-scripts#debugging-your-environment">Debugging your environment</a> for more information. </td>
   </tr>
   </tbody>
 </table>
 
-Most build scripts should set the `-e` and `-o pipefail` options to ensure if any command fails your entire build script fails. For example:
+Bash’s built-in `set` command with a `-` argument is used to enable options (the `+` argument is used to disable options).
```
toolmantim: Good catch. Re-reading that now, it is totally confusing. How's this latest version?

> Bash’s built-in `set` command can be used to enable and disable options. To enable an option use the `-` argument, and to disable an option use the `+` argument.

pages/builds/writing_build_scripts.md.erb
```diff
@@ -1,90 +1,145 @@
...
-```
+run_tests
+
+clean_up
+``` 
+
+Running this script will exit with the status returned by the final command, `clean_up`. However, what we really care about is the exit status of the first command, `run_tests`.
 
-You can also set the `-x` option which will print out each command, fully expanded, before executing it.
+By using a variable to store the exit status of `run_tests`, we can run additional commands (like `clean_up`) whilst still returning the original exit status. For example:
```
toolmantim: Cool, I fixed this one up!

> By using a variable to store the exit status of `run_tests`, we can run additional commands while still returning the original exit status. For example:

pages/builds/writing_build_scripts.md.erb
toolmantim: "If you require more environment information" is heaps better. Last resort sounds so dramatic!

toolmantim:
Thanks for the edits and help y’all! 🙌🏼

pages/builds/writing_build_scripts.md.erb
ticky: I don’t think `-` quite counts as an “argument” by itself. Might be clearer to phrase it more like this;

> Bash’s built-in `set` command can be used to enable and disable options.
> Options are enabled by prefixing them with a `-`, and disabled by prefixing them with a `+`.

Perhaps evan add an example of disabling an option afterwards, just for clarity!

toolmantim: I was never happy with the wording there. I've just improved it again in #131 and assigned you to review. Thanks!!

PR #90, opened 2017-02-08
Improve the Defining Your Pipeline Steps documentation

harrietgrace:
Improves the [Defining Your Pipeline Steps](https://buildkite.com/docs/pipelines/defining-steps) documentation.

- [x] Update show-example-test.png to have the correct emoji for the pipeline upload step
- [x] Do a retina version of the new screenshot

PR #93, opened 2017-02-10
Update documentation

nhuray:
Hey guys,

Here is a small correction in the documentation.

Nicolas


harrietgrace:
Good catch! Thanks for submitting 🙌🏻 I've been working on a broad update for this page, so I've added your find in the latest commit on #90.

PR #96, opened 2017-02-15
remove unnecessary word

liweinan0423:


keithpitt:
Good find, thanks @liweinan0423!

PR #106, opened 2017-04-21
Path to id_rsa and id_rsa.pub file were incorrect.

devacto:
Hi, I am using Buildkite on a Kubernetes cluster and it's been awesome. I followed your documentation. Everything went well except for the placement of the `id_rsa` and `id_rsa.pub` files which are supposed to be in the `/root/.ssh/` folder. When I deployed using the new folder it works. Thanks!


toolmantim:
Sorry that was incorrect, and thanks so much for the docs PR! ❤️

PR #108, opened 2017-04-26
Fix gcloud docs for ssh-key mounts.

fotinakis:
- The current `volumeMounts` config does not work, `mountPath` only accepts a folder and not a file unless you use "file projection" secrets. Also, in kubernetes YAML you can use octal `defaultMode: 0400` to be clearer about permissions (instead of `defaultMode: 254` which is required with kubernetes JSON).
- Also added a quick note about how to do container startup scripts.

### DIFF
--- a/pages/agent/gcloud.md.erb
+++ b/pages/agent/gcloud.md.erb
@@ -175,20 +175,14 @@ spec:
   containers:
   - ...
     volumeMounts:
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks again for the PR! Much appreciated. Those changes look good.

PR #116, opened 2017-05-19
Add GraphQL documentation

toolmantim:
This adds documentation for our GraphQL API, which I wanted to link to from an updated Scheduled Builds documentation page:

unknown-3-3-3

PR #117, opened 2017-05-19
Improvements to schedules docs

toolmantim:
Once #116 is merged, it'd be nice to tell people _How_ they can actually create schedules. This updates the Scheduled Builds documentation to include that, as well as some readability tweaks to the header/interval section.


toolmantim:
unknown-4-1-1-1-1-1-1

toolmantim:
An image would be nice. Let's just 🚀 this improvement for now though I think!

PR #120, opened 2017-05-24
Document pipeline.yml environment interpolation and caveats

toolmantim:
Our current [trigger documentation](https://buildkite.com/docs/pipelines/trigger-step) uses environment variable interpolation in the examples, but that's not supported in Buildkite Agent 2.0, only in the new Buildkite Agent 3.0 betas. Which can cause confusion (buildkite/feedback#260).

It is very common to want to use trigger steps in this way, so instead of just modifying those examples not to use interpolation, this adds documentation to the `pipeline upload` command, and a note below the trigger examples. The idea is that people will upgrade to 3.0 if necessary, or use the workaround if they can't upgrade their agent.

- [x] First cut
- [x] Update `pipeline upload` docs for `$VAR`, `${VAR}` and `${VAR:-default}` support etc

### DIFF
--- a/pages/agent/cli_pipeline.md.erb
+++ b/pages/agent/cli_pipeline.md.erb
@@ -44,7 +44,7 @@ Options:
 
 ## Pipeline format
 
-The pipeline can be written as YAML or JSON, but YAML is more commonly for its readability. There are two top level properties you can specify:
+The pipeline can be written as YAML or JSON, but YAML is more common for its readability. There are two top level properties you can specify:
 
 * `env` - A map of <a href="/docs/builds/environment-variables">environment variables</a> to apply to all steps
 * `steps` - A list of [build pipeline steps](/docs/pipelines/defining-steps)
--- a/pages/agent/cli_pipeline.md.erb
+++ b/pages/agent/cli_pipeline.md.erb
@@ -49,6 +49,31 @@ The pipeline can be written as YAML or JSON, but YAML is more commonly for its r
 * `env` - a hash of build-wide environment variables
 * `steps` - an array of build steps to be run (a mix of script, [`wait`](/docs/pipelines/wait-step) and [`block`](/docs/pipelines/block-step) steps)
 
+## Environment variable interpolation
 
+As of the Buildkite Agent 3.0 beta, the `pipeline upload` command interpolates all $ environment variables in your pipeline.yml files.
 
+For example:
 
+```yml
+- trigger: "$BUILDKITE_PIPELINE_SLUG-deploy"
+  label: "\:rocket\: Deploy"
+  branches: "master"
+  async: true
+  build:
+    message: "$BUILDKITE_MESSAGE"
+    commit: "$BUILDKITE_COMMIT"
+    branch: "$BUILDKITE_BRANCH"
+```
+
+You can also set defaults, for example:
+
+```
+"${SOME_ENV_VAR:default-value}"
+```
+
+Although we do not recommend it, if you need to emulate this behavior in older 2.0 agents you can use the following workaround:
+
+```shell
+echo $(eval echo $(cat pipeline.yml)) | buildkite-agent pipeline upload
+```
--- a/pages/agent/cli_pipeline.md.erb
+++ b/pages/agent/cli_pipeline.md.erb
@@ -6,7 +6,7 @@ See the [Uploading Build Pipelines](/docs/pipelines/uploading-pipelines) guide f
 
 <%= toc %>
 
-## Uploading Pipelines
+## Uploading pipelines
 
 ```
 Usage:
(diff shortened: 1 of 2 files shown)


toolmantim:
I had a read over the agent tests and source, and put together these docs.

@keithpitt as the author of said source code, could you please give this a quick check for accuracy?

@harrietgrace can I get you to check if there's anything you see that could be fixed, or any glaring problems?

unknown-123123123

Thanks!

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,64 @@ Options:
...
 
-## Pipeline Format
+## Pipeline format
 
 The pipeline can be written as YAML or JSON, but YAML is more commonly for its readability. There are two top level properties you can specify:
 
-* `env` - a hash of build-wide environment variables
-* `steps` - an array of build steps to be run (a mix of script, [`wait`](/docs/pipelines/wait-step) and [`block`](/docs/pipelines/block-step) steps)
+* `env` - A map of <a href="/docs/builds/environment-variables">environment variables</a> to apply to all steps
+* `steps` - An list of [build pipeline steps](/docs/pipelines/defining-steps)
```
harrietgrace: *A list

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,64 @@ Options:
...
+
+If an environment variable has not been set its default value will be a blank string. You can set your own default value using the syntax `${VAR:-default}`.
+
+For example, the following step will run the command `deploy.sh staging` if the `DEPLOY_ENV` environment variable has not been set or is blank (i.e. `""`):
+
+```yaml
+- command: "deploy.sh \"${DEPLOY_ENV:-staging}\""
+```
+
+If you need to substitute blank values (i.e. `""`) you can use the syntax `${VAR-default}`.
```
harrietgrace: This bit about default values is a little confusing. Up on 71 it says you can set your own default values (also, what's the scope of those?), then here it says you can use default to put in a blank value? It sounds like it will be a blank string if you don't update the default value - possibly worth mentioning here that you can sub in a blank, only if you haven't changed the default? Or is that not how this works 😅

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,64 @@ Options:
...
+
+For example, if the `DEPLOY_ENV` environment variable is set to `""` the following step will run the command `deploy.sh ""`, and if it has not been set it will run the command `deploy.sh "staging"`:
+
+```yaml
+- command: "deploy.sh \"${DEPLOY_ENV-staging}\""
+```
```
harrietgrace: Might be good to split this example up a bit. An idea:
Have the 'for example' bit first explaining what the code would run, then the example.
Then after that, say what would have happened if the env var isn't set.

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,64 @@ Options:
...
+
+You can substitute a range of characters from an environment variable by specifying a start and end position using the syntax `${VAR:start:end}`.
+
+For example, the following step will echo the first 7 characters of the `BUILDKITE_COMMIT` environment variable:
+
+```yaml
+- command: "echo ${BUILDKITE_COMMIT:0:7}"
+```
+
+If the environment variable has not been set, the range will return a blank string.
```
harrietgrace: Might be cool to give an example of a use case for this one? Not super necessary though

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,64 @@ Options:
...
+
+### Buildkite Agent 2.x support
+
+If you want to use environment variable substitution, but are unable to update your Buildkite Agent to version 3.0 or above, it is possible (but not recommended) to emulate the behavior using your shell’s built-in environment variable substitution:
```
harrietgrace: I think they probably already want to use env vars if they've made it this far down the page 😄 
We could start this one with the 'if you're unable to update your agent to 3.0' bit. It's a long sentence at the moment with a lot of different things in there, you end up a little lost by the end.

pages/agent/cli_pipeline.md.erb
toolmantim: Good catch!

toolmantim:
I think we're ready to go:

unknown-4-123123123123123

Will await @keithpitt’s eyes…

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,80 @@ Options:
    --debug-http                                 Enable HTTP debug mode, which dumps all request and response bodies to the log [$BUILDKITE_AGENT_DEBUG_HTTP]
 ```
 
-## Pipeline Format
+## Pipeline format
 
 The pipeline can be written as YAML or JSON, but YAML is more commonly for its readability. There are two top level properties you can specify:
```
keithpitt: Typo "but YAML is more commonly for its readability"

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,80 @@ Options:
...
 
-* `env` - a hash of build-wide environment variables
-* `steps` - an array of build steps to be run (a mix of script, [`wait`](/docs/pipelines/wait-step) and [`block`](/docs/pipelines/block-step) steps)
+* `env` - A map of <a href="/docs/builds/environment-variables">environment variables</a> to apply to all steps
+* `steps` - A list of [build pipeline steps](/docs/pipelines/defining-steps)
 
+## Environment variable substitution
 
+The `pipeline upload` command in Buildkite Agent versions 3.0 and above supports environment variable substitution using the syntax `$VAR` and `${VAR}`.
```
keithpitt: "Agent versions 3.0 and above supports environment" shouldn't it just be "support"?

toolmantim: Ah, I've reworded it to avoid confusion.!

pages/agent/cli_pipeline.md.erb
toolmantim: Nice catch. Fixed!

pages/agent/cli_pipeline.md.erb
```diff
@@ -42,13 +42,80 @@ Options:
...
+    <tr><th>Environment Variables</th><th>Syntax</th><th>Result</th></tr>
+  </thead>
+  <tbody>
+    <tr><td><code></code></td><td><code>"${SERVER-staging}"</code></td><td><code>"staging"</code></td></tr>
+    <tr><td><code>SERVER=""</code></td><td><code>"${SERVER-staging}"</code></td><td><code>""</code></td></tr>
+    <tr><td><code>SERVER="staging-5"</code></td><td><code>"${SERVER-staging}"</code></td><td><code>"staging-5"</code></td></tr>
+  </tbody>
+</table>
+
+### Character ranges
```
keithpitt: Might wanna mention "escaping" as well:

```
steps:
  - command: "$$FOO"
  - command: "\$FOO"
```

Will result in a commands of "$FOO" being sent to the agent (including the $)

keithpitt: Also, you can set "required" environment variables:

```
steps:
  - command: "${REQUIRED_VAR?}"
```

The `pipeline upload` command will exit with a status code of `1` if it hasn't been set.

keithpitt:
Nice, there's a few extra nuggets of into that you may wanna mention. Aside from that, 🚀

pages/agent/cli_pipeline.md.erb
toolmantim: Woo, I didn't catch these ones! What does `${REQUIRED_VAR?}` do for blank values? (e.g. `REQUIRED_VAR=""`). Does it still exit 1, or is blank allowed?

keithpitt: `REQUIRED_VAR?` the `? says it "must be set" - blank values are A-OK

toolmantim:
Ok, all done!

PR #125, opened 2017-06-05
Fix artifacts link

nullobject:
This PR fixes the artifacts link on the buildkite-agent artifacts page.


toolmantim:
Oh, thanks for the catch!

PR #126, opened 2017-06-08
unlocked => unblocked

mvines:
Minor typo (I think) that I noticed while reading docs


harrietgrace:
Thanks for that @mvines! Much appreciated 😊

PR #127, opened 2017-06-14
Add example for using an environment variable from steps

toolmantim:
If you've been using a pre-3.0 Buildkite Agent, with a pipeline that uses `$SOME_ENV` in it's commands, it can be a bit of trial and error to figure out exactly how you need to update your pipeline.yml files to make them work with the new environment variable interpolation.

Because this is probably quite common, this PR updates the environment variable section with an example of doing exactly this. Which should, fingers crossed, help people who are snagged by this.

We should also look at updating our "Pipeline Steps" documentation with this type of example too.

Before:

before

After:

after


toolmantim:
@asceresini awesome. Thanks for the review!!

PR #128, opened 2017-06-16
Fix errors and clarify artifact upload docs

toolmantim:
This aims to clean up a few mistakes in our artifact documentation that are tripping people up, and tries to make the artifact upload/download CLI documentation a little clearer. It covers:

* Fixes the incorrect `--job` argument to artifact download (the correct argument is `--step`)
* Avoids using the word "glob" to describe the download wildcard syntax
* Adds download examples that match the upload examples (as often, that's what you want to do… upload in one step, download in another)
* Fixes the broken links to the old `/docs/rest-api/artifacts#glob-path-pattern` URL
* Removes the link to our source code implementation of glob matching
* Removes the mentioning of the `%` SQL LIKE function
* Puts the upload/download examples before the technical/full definition of the pattern

### DIFF
--- a/pages/agent/cli_artifact.md.erb
+++ b/pages/agent/cli_artifact.md.erb
@@ -162,16 +162,24 @@ Downloading a specific file into a specific directory (note the trailing slash):
 buildkite-agent artifact download build.zip tmp/
 ```
 
-Downloading all the files uploaded to `log` into a local `log` directory (note that local directories will be created to match the uploaded file paths):
+Downloading all the files uploaded to `log` (including all subdirectories) into a local `log` directory (note that local directories will be created to match the uploaded file paths):
 
 ```bash
 buildkite-agent artifact download "log/*" .
 ```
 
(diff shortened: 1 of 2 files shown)


pages/agent/cli_artifact.md.erb
```diff
@@ -33,51 +33,24 @@ Example:
...
-- multiple paths are separated with `;`
-- surround the pattern with quotes
-
-The source path you supply to the upload command will be replicated exactly at the destination. If you run:
-
-```bash
-buildkite-agent artifact upload "log/test.log"
-```
-
-Buildkite will store the file at `log/test.log`. If you want it to be stored as `test.log` without the full path, then you'll need to change into the file's directory before running your upload command:
```
keithpitt: I thought this was important information in the docs - since it's a quirk of the uploader and downloader that you wouldn't expect - but the path the file is stored at is important for downloading.

toolmantim: It's just moved down into the examples!

keithpitt: Ah I missed it, soz!

PR #130, opened 2017-06-21
Fix incorrect env for token config

lox:
As pointed out by @mistydemeo, this is referring to the wrong env var. `BUILDKITE_AGENT_ACCESS_TOKEN` is the temporary token given to the agent by Buildkite, I'm pretty sure we mean `BUILDKITE_AGENT_TOKEN` here.

Fixes https://github.com/buildkite/agent/issues/490.


toolmantim:
loxs-first-docs-pr

PR #144, opened 2017-07-26
Correct name of envvar in documentation

kcollasarundell:
The buildkite agent does not appear to recognise the buildkite_meta_data
env var and it is referred differently above


toolmantim:
Oh, thanks for catching this @kcollasarundell!

PR #150, opened 2017-08-21
Fix a parsing issue causing weird formatting

ticky:
Comments starting with `##` in fenced code blocks are, for some reason, turned into HTML headings, then processed by the syntax highlighter. What the heck.

Anyway, this turns them into real Markdown headings, which sidesteps the problem entirely.

I’m gunna sneaky-merge this because it’s broken rn!


harrietgrace:
👍🏻👌🏻🙌🏻👏🏻

Issue #102, opened 2017-04-10
Trigger step access control isn’t mentioned


ticky:
It’d be helpful if the trigger step documentation included some mention that the user who starts the triggering build must have access to build in the triggered build.

joscha:
@ticky I think that would be good, yes, I just stumbled over this after I did a quite substantial refactor of our pipeline config, e.g.:

```

2017-08-17 06:13:27 WARN   POST https://agent.buildkite.com/v3/jobs/6b1020ca-37c7-4bfb-a2e0-bf85fd20470a/pipelines: 422 Builds cannot be triggered from this pipeline as there is no Buildkite user linked to this build. Ensure that a user has added "username@users.noreply.github.com" as a verified email address on their account. (Attempt 5/5 Retrying in 1s)
--
  | 2017-08-17 06:13:28 FATAL  Failed to upload and process pipeline: POST https://agent.buildkite.com/v3/jobs/6b1020ca-37c7-4bfb-a2e0-bf85fd20470a/pipelines: 422 Builds cannot be triggered from this pipeline as there is no Buildkite user linked to this build. Ensure that a user has added "username@users.noreply.github.com" as a verified email address on their account.
```

The pipeline that build is coming from is part of our private account, so there is no way it would be triggered in the first place if the user wasn't authorized to do so - hence all downstream pipelines are okay to be triggered. Is there a way to instruct BK to trigger that pipeline anyway or use a different method to trigger it?

harrietgrace:
Added in #151 🎉

PR #151, opened 2017-08-22
Update trigger docs

harrietgrace:
Adding some words about access control to the Trigger step doc re: https://github.com/buildkite/docs/issues/102

I'm not sure if these additions are super accurate - @ticky would you mind having a quick squiz and let me know if anything needs to be added or changed? 😊

### DIFF
--- a/pages/pipelines/trigger_step.md.erb
+++ b/pages/pipelines/trigger_step.md.erb
@@ -9,11 +9,17 @@ steps:
   - trigger: "another-pipeline"
 ```
 
-The author of the initial build will be the author of all builds created from a trigger step. 
+All builds created by a trigger step will be authored by the same user that created the initial build. This user must:
 
-If you have [Teams](/docs/pipelines/permissions) enabled in your organisation, ensure that this author has Build permission on all of the pipelines that will be triggered by this step.
+* be a member of your organisation
+* have a verified email address
+
+If you have [Teams](/docs/pipelines/permissions) enabled in your organisation, the authoring user must also have:
+
+* 'Build' permission on *every* pipeline that will be triggered
 
 If the author does not have the required access, the initial build will fail, and builds on subsequent pipelines will not be triggered.
+
 
 ## Trigger Step Attributes
 
--- a/pages/pipelines/trigger_step.md.erb
+++ b/pages/pipelines/trigger_step.md.erb
@@ -9,7 +9,7 @@ steps:
   - trigger: "another-pipeline"
 ```
 
-All builds created by a trigger step will be authored by the same user that created the initial build. This user must:
+All builds created by a trigger step will have the same author as the parent build. This user must:
 
 * be a member of your organisation
 * have a verified email address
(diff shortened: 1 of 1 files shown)


pages/pipelines/trigger_step.md.erb
```diff
@@ -9,6 +9,12 @@ steps:
   - trigger: "another-pipeline"
 ```
 
+The author of the initial build will be the author of all builds created from a trigger step. 
+
+If you have [Teams](/docs/pipelines/permissions) enabled in your organisation, ensure that this author has Build permission on all of the pipelines that will be triggered by this step.
+
+If the author does not have the required access, the initial build will fail, and builds on subsequent pipelines will not be triggered.
```
ticky: This is good, but it’s missing the requirement that the author be part of the org!

Maybe something like

>Any builds created by a trigger step will have the same author as the initiating build.
>
>The author of the build containing the trigger step must be a member of your organisation with a verified email address in order to trigger further builds. If you have [Teams](/docs/pipelines/permissions) enabled in your organisation, the author must have Build permission on the pipelines that will be triggered by this step.
>
>If the author does not have the required access, the trigger step will fail, and the requested builds are not created.

harrietgrace:
Thanks for the help @ticky! I added in the bit about orgs, but it was feeling a bit wordy. This bit is kind of important, so we don't want people skipping over the paragraph. To make it a bit punchier, I changed it to have dot points. What do you think? It doesn't read as smoothly, but I think it highlights the vital information better.

pages/pipelines/trigger_step.md.erb
```diff
@@ -9,6 +9,18 @@ steps:
   - trigger: "another-pipeline"
 ```
 
+All builds created by a trigger step will be authored by the same user that created the initial build. This user must:
```
ticky: “will be authored by” still feels very awkward to me

pages/pipelines/trigger_step.md.erb
```diff
@@ -9,6 +9,18 @@ steps:
...
+
+* be a member of your organisation
+* have a verified email address
+
+If you have [Teams](/docs/pipelines/permissions) enabled in your organisation, the authoring user must also have:
+
+* 'Build' permission on *every* pipeline that will be triggered
+
+If the author does not have the required access, the initial build will fail, and builds on subsequent pipelines will not be triggered.
```
ticky: “initial” build feels like kind of confusing wording

pages/pipelines/trigger_step.md.erb
lox: "All builds created by a trigger step will be created by the user that triggered the original build, or the user that unblocked the last preceding block step. That user must be:"

lox: Possibly with "If the build was triggered by a commit, then the user will be inferred from the commit author"?

pages/pipelines/trigger_step.md.erb
lox: "If the author doesn't have sufficient access to the pipeline being triggered, the trigger step will fail and cause the original build to fail as well"?

harrietgrace:
Eep, this pr really slipped through the cracks for me. Thanks for the feedbacks @ticky and @lox. 
I removed the 'will be authored by' and 'inital' bits, how does it feel with the revisions?

lox:
Sounds good to me!

harrietgrace:
Going to merge this one - happy to do a V2 sometime in the future if people are getting confused with the wording.

PR #153, opened 2017-09-05
WIP: Agent 3.0 docs

lox:


pages/agent/v3/installation.md.erb
```diff
@@ -0,0 +1,34 @@
...
+* [Linux](linux)
+* [Docker](docker)
+
+Alternatively you can install it manually using the instructions below.
+
+## Manual Installation
+
+If you need to install the agent on a system not listed above you'll need to perform a manual installation using one of the binaries from [buildkite-agent’s releases page](https://github.com/buildkite/agent/releases).
+
+Once you have a binary you simply create a bin and builds directory in `~/.buildkite-agent` and copy the binary and `bootstrap.sh` file into place:
```
harrietgrace: Random question: 
@lox is this still the case with the bootstrap file? We have a question in the issues here (bk/docs) about where the bootstrap file comes from, so I was having a poke around and saw something about the shell version of bootstrap being removed from v3?

pages/agent/v2.md.erb
```diff
@@ -45,7 +45,3 @@ The agent has a standard configuration file format on all systems to set meta-da
 ## Customising with Hooks
 
 The agent’s behaviour can be customized using hooks, which are just shell scripts that exist on your build machines or in each pipeline’s code repository. Hooks can be used to set up secrets as well as overriding default behavior. See the [hooks](agent/hooks) documentation for full details.
-
-## Signal Handling
```
keithpitt: This seems to also be the case in `v2` of the agent, so we can probably keep this in @lox? https://github.com/buildkite/agent/blob/2-6-stable/process/process.go#L276

keithpitt:
I've copied these changes into a new PR, so I'll close this one for now.

PR #160, opened 2017-09-26
Update build environment variables

toolmantim:
This PR adds some missing build environment variables, and documents a few of the news ones.


harrietgrace:
Are all documented now? There are a couple of outstanding/old issues asking about undocumented env vars, I'd love to be able to reply to them and say that they're all here now 😄

toolmantim:
Cool, yep these are all the public ones that you can rely on. There may be other environment variables present, due to the way the buildkite-agent works, but shouldn't be relied upon.

PR #163, opened 2017-10-15
Add some notes about securing the agent when building untrusted commits.

solemnwarning:


lox:
I totally missed this! Thanks @solemnwarning!

PR #173, opened 2017-11-23
Add allowed character set docs for block step key

keithduncan:
I ran into this from the API in a pipeline upload, this adds the error the API gives to the docs ahead of time 😸


harrietgrace:
Rad, thanks! Could you please put your additions before the 'example' lines? 😊

Issue #83, opened 2017-01-24
Proper way to handle node_modules across steps?


Akkuma:
My company is tarring up the node_modules and uploading this as an artifact. We then have to check in each build step if a node_modules exists and if not to download the artifact & untar it. None of the examples presented are even close to real world, since 1 step isn't close to how people use CI, which makes it quite difficult to understand whether this is the proper way to handle it.

ticky:
Tarring as an artefact is the way to do it if you’re not using docker!

This part of the setup is definitely something we hope to address in upcoming docs improvements!

lox:
I'm going to close this one in favour of the more general #177.

PR #177, opened 2017-11-29
Update docs on artifacts and meta-data

harrietgrace:
RE: https://github.com/buildkite/docs/issues/83 - a request for information on sharing data in builds. 

This is a draft of a new page that provides an overview of how/why to do data sharing in BK builds. Also updates the existing pages on meta-data, and maybe the one on artifacts too. 

Things to be done:
- [x] Update meta-data page
- [x] Expand dot points on data sharing page

Not doing for this PR:
- Add more use cases
- Add more code examples somewhere?
- Add note about google cloud only being available in 3.x agents


pages/agent/cli_meta_data.md.erb
```diff
@@ -8,7 +8,7 @@ See the [Using Build Metadata](/docs/builds/build-meta-data) guide for a step-by
 
 ## Setting Data
 
-Use this command in your build scripts to save simple string data to the build for retrieval in future steps.
```
harrietgrace: I know that in this instance 'simple string data' is referring to the type of data that you want to receive - is it actually different to referring to 'string data'? Are we specifically requesting that they use a subset of string's capabilities?

lox: Nope, you are correct, it's just a string.

lox:
I was thinking about this the other day, lots of new customers don't get our approach to making builds repeatable and stateless. I wonder if we should cover this in a getting started to tie it together? 

Also "Sharing data" wasn't immediately obvious to me what it would be about, will have a think about it 🤔

harrietgrace:
> I was thinking about this the other day, lots of new customers don't get our approach to making builds repeatable and stateless. I wonder if we should cover this in a getting started to tie it together?

@lox Totally! That would be great. I just added a bit to mention this idea briefly in the intro, but I think it would be good to talk about it more in a Getting Started guide. 

> Also "Sharing data" wasn't immediately obvious to me what it would be about, will have a think about it 🤔

The page itself is called "Sharing data between steps", does that make it any better? I could put the whole title in the sidebar if it does. Otherwise, I was thinking it could be something like 'reusing data in builds' or 'managing build data'? Something that is about the data that is created in a build, and also about doing things with that data 🤔

harrietgrace:
@lox I've just merged in master to this branch. Could you please have a look over what I've added so far? If you're happy with it/it's accurate, I'll add the new bits to the v3 artifact and meta-data docs.

lox:
Thoughts on "Sharing state across steps" vs "Sharing data..." ?

Other than that, I think this is a great change.

harrietgrace:
I like it! I'll change the title now, want me to add the changes to the v3 version of this doc too?

lox:
Actually, shall we scope this down to just the "Sharing state across steps" page and leave the agent docs customizations for later? 

I reckon we could get this shipped pretty quick if so.

harrietgrace:
@lox done! I'll keep a note that the v3 docs need to be updated too, I think I'll be able to get to it later today. This can ship as-is though!

pages/builds/build_meta_data.md.erb
```diff
@@ -1,25 +1,51 @@
...
+
+To set meta-data in the meta-data store, use the `set` command with a key/value pair:
 
 ```bash
 buildkite-agent meta-data set "release-version" "1.1"
 ```
 
+This command results in the value "1.1" being associated with the key "release-version" in the meta-data store. 
+
+To set meta-data on an agent, use the `meta-data` flag on the `start` command:
```
lox: Given this page is on Build meta-data, I reckon this might just confuse folks. I'd possible have a note about this confusion rather than this line?

E.g Note: The agent also has the concept of "meta-data" (renamed to "tags" in v3) for targeting builds to specific agents (e.g `queue=default`). This shouldn't be confused with the scoped key/value meta-data that the `buildkite-agent meta-data` command refers to!

pages/builds/sharing_state.md.erb
```diff
@@ -0,0 +1,50 @@
+# Sharing State Across Steps
+
+Sharing state across the steps of your build removes repetion and speeds up your build time. Passing data like npm packages and docker containers between steps can help keep your steps single-purpose (and cuts down the time each step takes and how much processing it has to do).
```
lox: .. and allow you to run lots of steps in parallel to speed up your builds.

pages/builds/sharing_state.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+Sharing state across the steps of your build removes repetion and speeds up your build time. Passing data like npm packages and docker containers between steps can help keep your steps single-purpose (and cuts down the time each step takes and how much processing it has to do).   
+
+There are two ways to share state across steps in a build: artifacts and meta-data. Both of these data stores are accessible across your organisation and pipelines.  
+
+<%= toc %>
+
+## Build Artifacts
+
+Artifacts are any files that are generated during a build step. Buildkite can host these files for you in the Buildkite artifact store.
```
lox: I view artifacts as products of a build that are useful to humans, or an input for a later build step.

pages/builds/build_meta_data.md.erb
harrietgrace: I was thinking about just removing it entirely... I left it in because I assumed that people searching for either one of the meta-datas in v2 would come to the meta-data page, and I wanted all the info to be there. 
Saying that, I'm happy to remove it, or add in your note. What do you think?

harrietgrace: Also, there's already a note in there to that effect, just after the code example. Should we remove mention of the flag and the example, but leave in the note?

pages/builds/sharing_state.md.erb
harrietgrace: Artifacts can be anything though, right? And we recommend that people only keep useful things/things they want to use later? I'd like to leave the first sentence there, and then add a recommendation about use. Yeah/nah?

pages/builds/build_meta_data.md.erb
lox: Yup, I reckon remove, or just a note as above to act as a disambiguator for v2.

pages/builds/sharing_state.md.erb
lox: Yeah, thats ok.

harrietgrace:
@lox updates from the most recent two commits:
- removed reference to the agent meta-data from the Build meta-data page
- added a line about how we recommend that people use artifacts for things they want to use later and things that will be useful to them
- added the new bits into the v3 versions of the artifact and meta-data docs (this was surprisingly short!)

harrietgrace:
@lox I pulled the 'sharing state' doc and its references out into a new PR - what are your feels about shipping this one with just the minor updates to the artifacts and metadata pages?

harrietgrace:
Sweet, thanks @toolmantim!

PR #181, opened 2018-01-05
Fix environment variable substitution Buildkite Agent 2.x support

ZimbiX:
The old command swallows all newlines, resulting in invalid YAML.


lox:
Thanks very much!

PR #183, opened 2018-02-07
Update S3 access URL environment variable

dblandin:
Not sure if this was ever `baseUrl` but it looks like it should be `BUILDKITE_S3_ACCESS_URL` based on the current agent code and https://github.com/buildkite/agent/pull/261.


harrietgrace:
Great catch @dblandin, thanks a bunch for the PR! 🙏🏻😄

PR #185, opened 2018-02-15
Documentation on latest retry changes

keithpitt:
I've started on docs on the new syntax I've added for retries.

Here's a rundown of all the changes I've made over the last few weeks. Not all of them in the docs, maybe you can fit them in somewhere else?

- Introduced a new `retry` configuration for steps.
  - Manual retries can be disabled
  - You can optionally provide a `reason` why manual retries are disabled
  - `permit_on_passed` allows manual retries to be performed on passed jobs
  - You can configure the job to automatically retry depending on exit status
  - You can configure how many times a job can be retried (max of 10) depending on exit status
  - If a job gets an exit status of (`-1`) that means we lost connection with the agent. By default, we automatically retry on this exit status as well. So if you `kill -9` an agent half way through a job, it'll automatically retry when our connection with the agent times out. Useful for spot instances that terminate jobs half way though.
- You can now manually cancel a job mid way through, and then retry it right away. This is useful if you know a job is going to fail, or if it hangs. The caveat here is you can only retry a job that was individually cancelled. If you cancel the entire build, no jobs can be retried anymore. We consider a build cancel to be "the last stop, nothing else can happen on this build"
- Instead of deleting logs/artifacts when retrying a job, we create a new job right next to the old one. By default, we hide the retried jobs in the build header. You need to hit the eyeball on the right hand side to toggle them on/off in that view. Retried jobs are still visible in the details section below the build header. We also now show who retried the job in the UI.
- Retried jobs are excluded from the REST API and web hooks.

### DIFF
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -161,6 +161,96 @@ _Optional attributes:_
   </tr>
 </table>
 
+## Retry Attributes
+
+<div class="Docs__note">
+<p class="Docs__note__heading">At least one of the following attributes is required.</p>
+<p>You don't have to set both <em>automatic</em> and <em>manual</em> set on your retry attribute, but you can choose to set both.</p>
+</div>
+
+<table>
+  <tr>
+    <td><code>automatic</code></td>
+    <td>
+      Whether to allow a job to retry automatically. This field accepts a boolean value, individual retry conditions, or a list of multiple allowed retry conditions.<br> If set to <code>true</code>, the retry conditions are set to the default value.<br>
+      <em>Default value:</em><br><code>retry:</code>
+      <br><code>  automatic:</code>
+      <br><code>    exit_status: "*"</code>
+      <br><code>    limit: 2</code>
+      <em>Example:</em> <code>true</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>manual</code></td>
+    <td>
+      Whether to allow a job to be retried manually. This field accepts a boolean value, or a single retry condition.<br>
+      <em>Default value:</em> <code>true</code>
+      <em>Example:</em> <code>false</code>
+    </td>
+  </tr>
+</table>
+
+```yml
+steps:
+  - label: "Linting"
+    command: "lint.sh"
+    retry:
+      manual: false
+  - label: "Tests"
+    command: "tests.sh"
+    retry:
+      automatic: true
+```
+
+
+## Automatic Retry Attributes
+
+_Optional Attributes_
+
+<table>
+  <tr>
+    <td><code>exit_status</code></td>
+    <td>
+      The exit status number that will cause this job to retry. 
+      <em>Example:</em> <code>"*"</code>
+      <em>Example:</em> <code>2</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>limit</code></td>
+    <td>
+      The maximum number of times this job can be retried.
+      <em>Example:</em> <code>3</code>
+    </td>
+  </tr>
+</table>
+
+```yml
+steps:
+  - label: "Linting"
+    command: "lint.sh"
+    retry:
+      automatic:
+        limit: 3
+  - label: "Linting"
+    command: "lint.sh"
+    retry:
+      automatic:
+        limit: 3
+  - label: "Tests"
+    command: "tests.sh"
+    retry:
+      automatic:
+        - exit_status: 5
+          limit: 2
+        - exit_status: "*"
+          limit: 4
+```
+
+## Manual Retry Attributes
+
+## Examples
+
 ```yml
 steps:
   - label: "\:hammer\: Tests"
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -180,18 +180,22 @@ _Optional Attributes_
     <td>
       The exit status number that will cause this job to retry. <br>
       <em>Example:</em> <code>"*"</code><br>
-      <em>Example:</em> <code>2</code><br>
-      <em>Example:</em> <code>-1</code>
+      <em>Example:</em> <code>2</code>
     </td>
   </tr>
   <tr>
     <td><code>limit</code></td>
     <td>
-      The maximum number of times this job can be retried.<br>
+      The number of times this job can be retried. The maximum value this can be set to is 10.<br>
       <em>Example:</em> <code>3</code>
     </td>
   </tr>
 </table>
+
+<div class="Docs__note">
+<p class="Docs__note__heading">-1 Exit Status</p>
+<p>An exit status of -1 results when a job loses connection with the agent running it. Jobs that exit with -1 will be automatically retried, unless you have overridden the retry behaviour in the <code>pipeline.yml</code> file.</p>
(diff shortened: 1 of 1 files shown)


harrietgrace:
This looks rad! I'll try adding those other bits in somewhere, and will stick that in another commit on here if that's alright with you 😊

keithpitt:
@harrietgrace please do! I'm happy to hand this PR over to you to do with what you will. If there's any other info or clarification of how stuff works lemme know.

harrietgrace:
Sorry I didn't get to finish this off today @keithpitt! If you want to ship it as-is, I can update it next week with some more bits?

pages/pipelines/command_step.md.erb
```diff
@@ -140,12 +179,22 @@ steps:
       - "coverage/**/*"
     parallelism: 1
     timeout_in_minutes: 60
+    retry:
+      automatic:
+        - exit_status: 1
```
DazWorrall: It might be useful to document the `-1` use case - where there was actually no exit code reported, likely because of an infrastructure issue.

pages/pipelines/command_step.md.erb
```diff
@@ -120,12 +120,51 @@ _Optional attributes:_
...
+      <code>retry:</code><br>
+      <code>&nbsp;&nbsp;automatic: true</code><br/>
+
+      <br/>
+      <em>Example (different automatic retry rules depending on exit status):</em><br>
+      <code>retry:</code><br>
+      <code>&nbsp;&nbsp;automatic:</code><br/>
+      <code>&nbsp;&nbsp;&nbsp;&nbsp;- exit_status: 5</code><br/>
+      <code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;limit: 2</code><br/>
+      <code>&nbsp;&nbsp;&nbsp;&nbsp;- exit_status: "*"</code><br/>
```
harrietgrace: Does `exit_status` always need to have a `-`in front of it?

keithpitt: Yeah, in this case, we're defining a list of "automatic retry rules". There are 2 rules defined:

```
- exit_status: 5
   limit: 2
- exit_status: "*"
   limit: ...
```

The `-` means its a YAML array. Just like:

```
steps:
  - command: "..."
     foo: "bar"
```

pages/pipelines/command_step.md.erb
keithpitt: Ah yeah, good call! @harrietgrace where do you reckon we put that?

pages/pipelines/command_step.md.erb
```diff
@@ -120,12 +120,51 @@ _Optional attributes:_
...
+      <code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;limit: 4</code><br/>
+
+      <br/>
+      <em>Example (disabling manual retries):</em><br>
+      <code>retry:</code><br>
+      <code>&nbsp;&nbsp;manual: false</code><br>
```
harrietgrace: @keithpitt does `manual` default to true? And can a "manual" retry come from the API as well as the app, or only from the button in the app?

keithpitt:
@harrietgrace did you still want to ship as-is? Or did you wanna tweak first? I'm happy for you to merge as soon as you're ready.

pages/pipelines/command_step.md.erb
keithpitt: That's right, `manual: true` is the default. A "manual" retry is a retry initiated by a user, either via the API or UI. There's no difference to BK between an API retry or a UI retry.

harrietgrace:
@keithpitt I'm shuffling things around right now, so I'll merge it this afternoon once I'm done 😊

keithpitt:
@harrietgrace radical

pages/pipelines/command_step.md.erb
harrietgrace: Okay cool. So how about `limit`? It looks like that one can be standalone and doesn't need a `-`?

pages/pipelines/command_step.md.erb
harrietgrace: I'm adding in an 'optional attributes' section for `automatic`, I'll put a note in there!

pages/pipelines/command_step.md.erb
```diff
@@ -120,12 +120,51 @@ _Optional attributes:_
...
+
+      <br/>
+      <em>Example (allowing retries on passed jobs):</em><br>
+      <code>retry:</code><br>
+      <code>&nbsp;&nbsp;manual:</code><br>
+      <code>&nbsp;&nbsp;&nbsp;&nbsp;permit_on_passed: true</code><br/>
```
harrietgrace: Are there any other options available under `manual`? Or is it just `permit_on_passed`?

pages/pipelines/command_step.md.erb
keithpitt: `automatic` can accept either a single retry rule (no `-`) or multiple retry rules (which need a `-`).

So:

```
# Changes the default automatic retry rule to have a limit of 4.
automatic:
  limit: 4

# Changes the default automatic retry rule to only look at jobs with an exit status of 23.
automatic:
  exit_status: 23

# When defining an array of exit status rules, both `exit_status` and `limit` are required. So this will raise an error...
automatic:
  - exit_stauts: 23

# As will this
automatic:
  - limit: 2
```

keithpitt: Oh, and a single retry rule looks like this:

```
automatic:
  limit: 4
  exit_status: 2
```

If they want to add another rule, they'll need to convert it to a list:

```
automatic:
  - limit: 4
    exit_status: 2
  - limit: 2
    exit_status: 5
```

pages/pipelines/command_step.md.erb
keithpitt: There's `allowed: true/false`, which you'll probably use along side the optional property `reason`

```
manual:
  allowed: false
  reason: "No retries allowed on deploy steps"
```

In the below example, the `reason` has no affect and will do nothing:

```
manual:
  allowed: true
  reason: "No retries allowed on deploy steps"
```

Also in this example, `permit_on_passed` will have no effect since we've turned off manual retries anyway.

```
manual:
  allowed: false
  permit_on_passed: true
```

harrietgrace: Is the `reason` field displayed somewhere in the UI?

keithpitt: Yeah, the "Retry" button will be disabled in the UI with a ⚠ icon next to it. When you over over the button a tooltip while show you why it can't be retried (in this case, the user provided reason will be shown)

keithpitt: screen shot 2018-02-19 at 12 25 18 pm

harrietgrace: One more question 😄does `permit_on_passed` have a default value?

harrietgrace:
@keithpitt could you please have a look over the things I've added? Just to make sure I haven't added in anything wrong, or missed something important 😊

keithpitt:
@harrietgrace looks good to me!

PR #191, opened 2018-03-06
Fix typo in retry docs

shamanskyh:
I noticed this while I was checking out the docs on the automatic retry attributes. I'm assuming this is a typo, but please let me know if I'm mistaken :)


harrietgrace:
Hey @shamanskyh, yes it's definitely a typo! Thanks a bunch for taking the time to PR an update 😊

PR #193, opened 2018-03-12
Add an example for patch semver number

uriahcarpenter:
It took me a few attempts to figure out that you can have multiple wildcards.


lox:
Thanks @uriahcarpenter!

PR #196, opened 2018-03-19
Copy all of agent v2 docs into a v3 namespace

keithpitt:
This is the last preparation change before going deep into the v3 specific docs.

This change just copies all of the v2 docs into a v3 namespace, at which point we can start chopping changing so we've got a proper diff of all the changes that were made going from v2 to v3.


lox:
👍

PR #199, opened 2018-03-21
Remove shebangs from hooks

### LINKED ISSUE #79: Hooks that export variables or change directories should not have a #!/bin/bash line.
I'm looking at: 
https://buildkite.com/docs/agent/hooks

An example hook is given as:

    #!/bin/bash
    set -eu
    echo '--- :house_with_garden: Setting up the environment'
    export GITHUB_RELEASE_ACCESS_KEY='xxx'

The first line  "#!/bin/bash"  implies this is a script which will get executed.   If it is executed, the export won't do anything.     A similar issue exists with the  pre-command hook.
It's a small picky issue, but caused me some confusion and extra work when someone asked me to create an env hook.

lox:
Hooks are sourced, not executed, so shebangs are misleading.

Closes #79.

### DIFF
--- a/pages/agent/v2/hooks.md.erb
+++ b/pages/agent/v2/hooks.md.erb
@@ -42,10 +42,6 @@ Hooks are bash scripts you can use to execute commands and export environment va
 The following is an example of a custom `environment` hook which exports a GitHub API key for the pipeline’s release build step:
 
 ```bash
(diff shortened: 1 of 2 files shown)


harrietgrace:
👌🏻👌🏻👌🏻

PR #204, opened 2018-03-26
Install on windows docs

lox:
Document how to use the new install script to install via Powershell, also covers how to run as a service.

### DIFF
--- a/pages/agent/v3/install/_windows.md.erb
+++ b/pages/agent/v3/install/_windows.md.erb
@@ -1,4 +1,4 @@
 1. Download the latest Windows beta from <a href="https://github.com/buildkite/agent/releases">Buildkite Agent releases on GitHub</a>
-2. Extract the files to a directory of your choice
+2. Extract the files to a directory of your choice (we recommend C:\buildkite)
 3. Edit `buildkite-agent.cfg` and add your token
 4. Run `buildkite-agent start` from a command prompt
--- a/pages/agent/v3/install/_windows.md.erb
+++ b/pages/agent/v3/install/_windows.md.erb
@@ -1,4 +1,4 @@
 1. Download the latest Windows beta from <a href="https://github.com/buildkite/agent/releases">Buildkite Agent releases on GitHub</a>
-2. Extract the files to a directory of your choice (we recommend C:\buildkite)
+2. Extract the files to a directory of your choice (we recommend `C:\buildkite-agent`)
 3. Edit `buildkite-agent.cfg` and add your token
 4. Run `buildkite-agent start` from a command prompt
--- a/pages/agent/v3/windows.md.erb
+++ b/pages/agent/v3/windows.md.erb
@@ -40,7 +40,7 @@ The configuration file is located at `C:\buildkite-agent\buildkite-agent.cfg`. S
 
 ## Upgrading
 
-Simply rerun the install script.
+Rerun the install script.
 
 ## Configuration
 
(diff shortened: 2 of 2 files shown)


pages/agent/v3/windows.md.erb
```diff
@@ -13,16 +22,44 @@ The Buildkite Agent can be installed on both 64bit and 32bit editions of Windows
...
+* Builds: `C:\buildkite\builds`
+* SSH keys: `%USERPROFILE%\.ssh`
+
+## Configuration
+
+The configuration file is located at `C:\buildkite\buildkite-agent.cfg`. See the [configuration documentation](/docs/agent/v3/configuration) for an explanation of each configuration setting.
+
+## Upgrading
+
+Simply rerun the install script.
```
harrietgrace: Can we remove the 'simply' here? It would be even more to the point as 'Rerun the install script' 😊

pages/agent/v3/windows.md.erb
```diff
@@ -4,7 +4,16 @@ The Buildkite Agent can be installed on both 64bit and 32bit editions of Windows
 
 <%= toc %>
 
-## Installation
+## Installing with Powershell
```
keithpitt: We should also update the docs here: https://github.com/buildkite/frontend/blob/master/app/components/docs/07-windows.mdx

pages/agent/v3/install/_windows.md.erb
```diff
@@ -1,4 +1,4 @@
 1. Download the latest Windows beta from <a href="https://github.com/buildkite/agent/releases">Buildkite Agent releases on GitHub</a>
-2. Extract the files to a directory of your choice
+2. Extract the files to a directory of your choice (we recommend C:\buildkite)
```
keithpitt: On linux we use `~/.buildkite-agent` so shouldn't the folder be `Buildkite Agent` or something?

pages/agent/v3/windows.md.erb
lox: Will do!

pages/agent/v3/install/_windows.md.erb
keithpitt: Or put it in "UserData"?

lox: Possibly. Lots of tools like Git for Windows and Golang go for a simpler approach of a top-level directory that is simply named. When I asked what best practice was from some Windows folks, they indicated a simple top-level dir was preferable. 

Whether it's `C:\buildkite-agent` or `C:\buildkite` is another question. I was going for naming it after the company, but I can see the arguments for either, especially seeing as we aren't namespacing `C:\buildkite\agent` or similar.

keithpitt: > Possibly. Lots of tools like Git for Windows and Golang go for a simpler approach of a top-level directory that is simply named. When I asked what best practice was from some Windows folks, they indicated a simple top-level dir was preferable.

Yeah cool, what ever the Windows folks say!

> Whether it's C:\buildkite-agent or C:\buildkite is another question

I'd aim for `buildkite-agent` with parity with the Linux one, and `buildkite-agent` is the name of the software.

pages/agent/v3/windows.md.erb
lox: Copy/pasted from our other docs, but agree this is a good change and will make it across the board.

pages/agent/v3/install/_windows.md.erb
lox: Ok, I'll PR the agent install script to correct this and then update these docs to reflect it.

lox: The linux installer script creates `~/.buildkite-agent/bin` to put the binary in. Do we want to replicate this?

PR #205, opened 2018-03-26
Security guide updates for v3

keithpitt:


keithpitt:
I'd really like to write some words about what the ramifications of turning it off are, and what the ramifications of having it on by default are.

Do you think that'd be OK to add? Or do you think this guide assumes that they're experts on this stuff?

harrietgrace:
I think that would be really useful, I'm sure there will be plenty of non-experts reading this doc. And it would be a good refresher for experts anyway 😄

lox:
Yup, this is awesome. 

Is it worth explaining some of the why? E.g.

> Automatically scanning keys for ssh hosts uses a "trust on first use" model, which is generally the approach that git/ssh takes. Unfortunately in environments where hosts are being added and removed frequently, there are more opportunities for man-in-the-middle attacks to intercept these requests and modify the response. If this is a concern for your use-case, this feature lets you manage it yourself.

lox:
Beyond that, GitHub publishes their key fingerprints:

https://help.github.com/articles/github-s-ssh-key-fingerprints/

If I were using the `no-ssh-keyscan` feature, I'd probably have an `authorized_hosts` file setup with those fingerprints hardcoded into it.

keithpitt:
(I did a sneaky deal with @lox where I did some data analysis for him and he'll take over this PR and finish the SSH keyscan docs :P)

harrietgrace:
@keithpitt I think this is ready to go?

lox:
I said I'd write some more words for it, but let's get this merged and then I will follow up if we have time.

harrietgrace:
@lox I added in the extra bits that keith requested, was there more on top of that?

PR #213, opened 2018-04-02
Fixe the agent communication SVG on small screens

toolmantim:
In #211 the SVG image was inlined into the HTML, but didn't handle smaller screens, or namespaced the ids. This PR should (if I'm not mistaken) fix these two things.

Before:

before-1

After:

after-1


toolmantim:
I might have to merge this @ticky sorry. But if you spot anything, feel free to go ahead and fix it ☺️

ticky:
Ah geez I didn’t expect svgo to nuke the viewbox. Should’ve checked!

PR #214, opened 2018-04-04
Add instructions on how to open powershel with admin

keithpitt:
I was playing with Windows stuff, and I wasn't sure how to open a powershell with Admin privelages. It's required, so I figured some sneaky words couldn't hurt!

### DIFF
--- a/pages/agent/v3/windows.md.erb
+++ b/pages/agent/v3/windows.md.erb
@@ -6,7 +6,9 @@ The Buildkite Agent can be installed on both 64bit and 32bit editions of Windows
 
 ## Automated install with Powershell
 
-You'll need to run the follow commands within Powershell with Administrative Privileges. You can do this (in Windows 8) by:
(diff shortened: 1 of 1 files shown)


pages/agent/v3/windows.md.erb
```diff
@@ -4,7 +4,15 @@ The Buildkite Agent can be installed on both 64bit and 32bit editions of Windows
 
 <%= toc %>
 
-## Installing with Powershell
+## Automated install with Powershell
+
+You'll need to run the follow commands within Powershell with Administrative Privileges. You can do this (in Windows 8) by:
```
harrietgrace: 'following' rather than 'follow'?

harrietgrace: And it might be worth clarifying a little at the end with 'To start a Powershell session as an Administrator (in Windows 8):'

PR #219, opened 2018-04-09
Fix curl url for artifacts/download.

papaver:
Looks like the curl urls for both get and downloading an artifact are missing the job information.


keithpitt:
Ah yes, looks like those routes were from v1 of our API. Thanks for the fix!

PR #220, opened 2018-04-10
Add an out-of-date agent warning to v2 docs pages

toolmantim:
This adds a warning to the top of v2 agent docs, directing people to the v3 version:
v2-notice

Handy if they've arrived there by Google, or old links, and adds to the internal page rank for the v3 pages.

- [x] First cut, for words feedback
- [ ] Add to all v2 agent docs pages

### DIFF
--- a/pages/agent/v2/aws.md.erb
+++ b/pages/agent/v2/aws.md.erb
@@ -1,5 +1,10 @@
 # Running Buildkite Agent on AWS
 
+<section class="Docs__troubleshooting-note">
+  <p>This page references the out-of-date Buildkite Agent v2.</p>
+  <p>For docs referencing the Buildkite Agent v3, <a href="/docs/agent/v3/aws">see the latest version of this document</a>.
+</section>
+
 The Buildkite Agent can be run easily on AWS using our Elastic CI Stack for AWS.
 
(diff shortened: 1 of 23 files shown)


harrietgrace:
I had another go at some words for this:
image
I think it makes the link a bit more obvious and clickable. I'm not sure if we should draw more attention to the out-of-date-ness of the v2 agent, I only really left it out because it didn't flow as well. I can totes add it back in though.

lox:
I reckon you could consolidate that down to:

```
!! This page describes how to work with Buildkite Agent v3
For docs referencing Buildkite Agent v3, see the latest version of this document
```

lox:
I reckon we should just ship as-is. I think there is some value in keeping "out-of-date" in the wording.

harrietgrace:
How about we meet in the middle with a tiny update to the original that keeps the 'out-of-date' wording, but rephrases it a little and has a shorter link? 😄 

> This page references the out-of-date Buildkite Agent v2.
> For docs referencing the Buildkite Agent v3, see the latest version of this document.

lox:
Works for me!

toolmantim:
Noice! 🚀

PR #225, opened 2018-04-24
Document the `buildkite-agent meta-data exists` command

haines:
This command seems to be undocumented online at the moment, but it is listed in `buildkite-agent meta-data --help`.


harrietgrace:
Great addition, thanks for taking the time to PR @haines! 🙌🏻🚀

PR #227, opened 2018-04-27
Fix environment variable for disabling plugins

shaun-whitely:
The environment variable for disabling plugins seems to be incorrectly documented. [Here is the exact location in `buildkite/agent` where the environment variable is defined](https://github.com/buildkite/agent/blob/adee22c9d9f102239f2f6271315208f797f89196/clicommand/agent_start.go#L250).


lox:
Sorry about this, and good catch @shaun-whitely! We'll get this updated ASAP.

PR #234, opened 2018-04-29
Make this match the agent v3 code

petemounce:
https://github.com/buildkite/agent/blob/master/clicommand/bootstrap.go#L217-L221

I'm not sure this is the correct way around to fix this, but at the moment, the docs and code seem to be at odds.

I would find it clearer if the variable indicated positively which way around the boolean should be interpreted. So `BUILDKITE_WILL_FETCH_GIT_SUBMODULES` or similar. But, naming is hard :)

### DIFF
--- a/pages/builds/environment_variables.md.erb
+++ b/pages/builds/environment_variables.md.erb
@@ -67,7 +67,7 @@ The following environment variables are automatically provided to every job, and
     <td>The commit of the build. <p class="Docs__api-param-eg"><em>Example:</em> <code>"83a20ec058e2fb00e7fa4558c4c6e81e2dcf253d"</code></p></td>
   </tr>
   <tr>
-    <th><code>BUILDKITE_DISABLE_GIT_SUBMODULES</code></th>
+    <th><code>BUILDKITE_GIT_SUBMODULES</code></th>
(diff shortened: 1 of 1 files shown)


lox:
This is a confusing aspect of our code that I've long been trying to figure out how to reconcile. The `buildkite-agent start` command reads env in the negative https://github.com/buildkite/agent/blob/f1e729b038ee71c1ea017953c80fb6b2fde4b499/clicommand/agent_start.go#L260 and then converts them to a positive phrasing here https://github.com/buildkite/agent/blob/f1e729b038ee71c1ea017953c80fb6b2fde4b499/clicommand/agent_start.go#L367, and then sets them as environment for the running job here https://github.com/buildkite/agent/blob/f1e729b038ee71c1ea017953c80fb6b2fde4b499/agent/job_runner.go#L232.

That is why the bootstrap is expecting the env in the form `BUILDKITE_GIT_SUBMODULES`. It ends up being an implementation detail though, that inversion shouldn't be relevant to scripts that are run by the bootstrap.

What problem are you trying to solve @petemounce?

petemounce:
When I posted from docs to my golang binary to harvest environment variables onto build log lines being sent to logstash, this was unset. I thought to make it match what the agent set. 

I'll double check to make sure it wasn't legitimately unset, as opposed to mismatch of name between docs and agent.

lox:
Going to close this one as I think it's a misunderstanding, let us know if not @petemounce.

Issue #230, opened 2018-04-28
Contradictory statements for Artifact stage


baank:
Referring to this page:
https://buildkite.com/docs/agent/v3/cli-artifact#uploading-artifacts

Within the "Uploading artifacts" section it shows this:
`  $ export BUILDKITE_S3_ACCESS_KEY_ID=xxx
   $ export BUILDKITE_S3_SECRET_ACCESS_KEY=yyy
   $ export BUILDKITE_S3_DEFAULT_REGION=eu-central-1 # default is us-east-1
   $ export BUILDKITE_S3_ACL=private # default is public-read
   $ buildkite-agent artifact upload "log/**/*.log" s3://name-of-your-s3-bucket/$BUILDKITE_JOB_ID
`

But then further down in the "Using your own private AWS S3 bucket" section it says:

"To do so, you’ll need to export the following environment variables using an environment agent hook (this can not be set via the Buildkite web interface, API, or during pipeline upload):"

`export BUILDKITE_ARTIFACT_UPLOAD_DESTINATION="s3://name-of-your-s3-bucket/$BUILDKITE_JOB_ID"
export BUILDKITE_S3_DEFAULT_REGION="eu-central-1" # default: us-east-1
`

Do I need to use an environment agent hook or can I just export variables from the pipeline code ?

harrietgrace:
Thanks for picking that up @baank, and for taking the time to let us know 🙏🏻

In the docs section "Using your own private AWS S3 bucket", I've clarified that there are multiple ways to export the variables, but we recommend using hooks as a best practice. I've also removed the parenthetical about not being able to set the environment variables in some places - I ran some tests just now and they no longer apply 😊

PR #238, opened 2018-05-02
Add example and clarify artifact downloads outside builds

### LINKED ISSUE #229: Meaning of download artifacts section is unclear
In regards to this section:
https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts-outside-a-running-build

It isn't clear what this means. Does it mean:

(1) Downloading artifacts outside the current build
(2) Downloading artifacts outside any build

harrietgrace:
Added @lox's explanatory words about artifact downloads, plus a little example. Does this sound accurate @lox?

Also fixed up the S3 buckets bit - removed an out-of-date caveat about not being able to set env vars in places, added a note that using hooks is best practice, and updated some syntax in the env vars example so that the `env` is a proper hash. 

Fixes #229 and #230

### DIFF
--- a/pages/agent/v3/cli_artifact.md.erb
+++ b/pages/agent/v3/cli_artifact.md.erb
@@ -198,7 +198,9 @@ There is no need to escape characters such as `?`, `[` and `]`.
 
 ## Downloading artifacts outside a running build
 
-The `buildkite-agent artifact download` command only works within the context of a running build.
+The `buildkite-agent artifact download` command only works within the context of running builds. It relies on environment variables that are set by the agent while a build is running. 
+
+For example, executing the `buildkite-agent artifact download` command on your local machine would return an error about missing environment variables. However, when this command is executed as part of a build, the agent has set the required variables, and the command will be able to run. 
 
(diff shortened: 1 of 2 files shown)


pages/agent/v3/cli_artifact.md.erb
```diff
@@ -198,7 +198,9 @@ There is no need to escape characters such as `?`, `[` and `]`.
 
 ## Downloading artifacts outside a running build
 
-The `buildkite-agent artifact download` command only works within the context of a running build.
+The `buildkite-agent artifact download` command only works within the context of running builds. It relies on environment variables that are set by the agent while a build is running.
```
lox: I wonder if a clearer way to frame this might be "... command relies on environment variables that are set by the agent while a build is running."

lox: "within the context of running builds" seems confusing to me, and not actually accurate.

harrietgrace: Yeah, I like it! Added in 👌🏻

PR #239, opened 2018-05-03
Add a doc with info about sharing state

harrietgrace:
Previously part of PR #177, this is a high level doc about sharing data/state in pipelines.

### DIFF
--- a/pages/builds/sharing_state.md.erb
+++ b/pages/builds/sharing_state.md.erb
@@ -1,0 +1,50 @@
+# Sharing State Across Steps
+
+Sharing state across the steps of your build removes repetion and speeds up your build time. It also helps you use the same state for multiple steps running in parallel. Passing data like npm packages and docker containers between steps can help keep your steps single-purpose, decreases the amount of processing each step has to do, and cuts down the time they take to complete.
+
+There are two ways to share state across steps in a build: artifacts and meta-data. Both of these data stores are accessible across your organisation and pipelines.  
+
+<%= toc %>
+
+## Build Artifacts
+
+Artifacts are any files that are generated during a build step. Buildkite can host these files for you in the Buildkite artifact store. We recommend only storing artifacts that you need in subsequent build steps, or files like logs that would be useful at a later date.
+ 
+You can upload and download from the store using our REST API. There are limits to how much you can upload, and to how long your files can be kept for. You can also use our command-line tool to upload your artifacts to a personal store.  
+
+Using a personal store like Amazon S3 is recommended when:
+
+- you need to persist the artifact longer than the build
+- your artifacts contain sensitive information
+
+If you're using the v3 Agent, you're also able to use the artifact [plugin](/docs/agent/plugins).  
+
+For a short introduction to uploading and downloading from the Buildkite artifact store, see the [Using Build Artifacts](/docs/builds/artifacts) page.
+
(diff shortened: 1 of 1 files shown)


ticky:
Jotting down notes on this, in no particular order, as I won’t be able to attend the chat about it tonight!

1. I’m excited to have this well-documented, but I think this info needs to be more up-front. I believe it needs at least a brief explanation as early as [Pipelines](https://buildkite.com/docs/pipelines), and likely needs to be mentioned in [Defining Steps](https://buildkite.com/docs/pipelines/defining-steps) and the [Command Step](https://buildkite.com/docs/pipelines/command-step).
    1. That brief explanation should at least note:
        - that steps are not expected or guaranteed to run on the same agent
        - that shared state like `node_modules` or other local installation or compiled assets need to be passed as artifacts

3. This new document should also be linked to in context by [buildkite-agent artifact](https://buildkite.com/docs/agent/v3/cli-artifact), [Using Build Artifacts](https://buildkite.com/docs/builds/artifacts), [buildkite-agent meta-data](https://buildkite.com/docs/agent/v3/cli-meta-data) and [Using Build Metadata](https://buildkite.com/docs/builds/build-meta-data).

4. Looking over the introduction here, I feel like while we make this call because it will make things faster and able to be done in parallel, it’s not something you can ignore, so we probably also need to center these restrictions upfront and justify them as the not-so-secret sauce allowing for the growth and expansion in accomplishing complex testing and deployment tasks our bigger customers enjoy on the platform.

5. This document should also probably mention Docker and the Docker-compose plugin as additional options for sharing state which might be useful if users already have a Docker-based workflow.

6. If we have a recommended method of working around the “no guarantee of running on the same agent” restriction, we should probably document that too, as a last resort (I think it involves manually setting queues for builds at the step upload point, though?)

I’m excited to see this making progress! ✨

harrietgrace:
@ticky @lox - @toolmantim and I had some quick chats about this, and this is what we came up with for a start 😄There will totes be more steps after this, feel free to add more thoughts and feels to this thread.

First steps for improving this doc: 
- separate out into two docs: guide to caching, and intro to stateless jobs
- the guide will go into the tutorials section, and the intro will be in pipelines
- links to the guide will go into the meta data and artifacts existing docs
- links to the intro will go into some of the other pipelines docs like command step and defining steps
- add stuff about docker, pre-existing caching stuff like s3 usage, single agent pipelines, etc. to the guide

filipesilva:
Heya, was there any more development regarding the `node_modules` caching usecase? That one is specially interesting for me.

filipesilva:
Hello!

Just wanted to ping you all again on this topic. We've been trying different approaches to caching on nodejs projects and settled on have a [rolling docker image as cache](https://github.com/angular/angular/pull/27990) on the target branch for simplicity. 

This isn't great in the sense that docker images are quite big. But it does make it easier since we cached docker layers are reused on the small machine pool we have.

What's the current thinking/recommendations on BuildKite caching strategies?

harrietgrace:
Hey @filipesilva! I'm sorry we didn't get back to your message, thanks for bumping the thread 🙏🏻We've been having lots of internal discussions about caching, but haven't settled on anything we'd consider 'best practice' yet. 

@lox @plasticine - I think you folks have opinions on caching? Also, check out @filipesilva's caching for node projects 👀

PR #245, opened 2018-05-15
Fix old links to v2 agent docs

toolmantim:


lox:
Awesome, I wonder if this will fix the google issues we have with v2 still showing up. Nice find, perplexed how I missed this initially!!

toolmantim:
Fingers crossed @lox!

PR #246, opened 2018-05-15
Add BUILDKITE_TRIGGERED_FROM_BUILD_ID to env vars

toolmantim:


pages/builds/environment_variables.md.erb
```diff
@@ -142,6 +142,10 @@ The following environment variables are automatically provided to every job, and
     <th><code>BUILDKITE_TIMEOUT</code></th>
     <td>The number of minutes until Buildkite automatically cancels this job, if a timeout has been specified. <p class="Docs__api-param-eg"><em>Example:</em> <code>"15"</code> for 15 minutes, or <code>"false"</code> if no timeout is set</td>
   </tr>
+  <tr>
+    <th><code>BUILDKITE_TRIGGERED_FROM_BUILD_ID</code></th>
+    <td>The UUID of the build containing the <a href="/docs/pipelines/trigger-step">trigger step</a> that was used to trigger this build. <p class="Docs__api-param-eg"><em>Example:</em> <code>"5aa7c894-c8c0-435b-bc17-13923b90f163"</code>, or <code>""</code> if the build was not triggered from another build.</td>
```
harrietgrace: Could we have the example a just the values? The words about the empty string case could go in the body of the description, and we could have the empty string example code as a 'default' case?

toolmantim: I just checked this out again, and tried to have a stab at it, and then realised that I was following the pattern from many of the other envs, such as `BUILDKITE_BUILD_ID `, `BUILDKITE_PULL_REQUEST_REPO`. They follow the same format.

I wonder if the description on this was making it worse though? I've just updated it from:

image

to:

image

What do you reckon? Or should we change it (and maybe the others?)

toolmantim:
Okay cool! Sounds like a good plan.

PR #250, opened 2018-05-15
Call out the top level yaml overrides

toolmantim:
This behaviour is counteruntuitive, and we'd love to change it for future pipeline.yml versions, but for now maybe let's call it out more strongly?


toolmantim:
Oh whoops. It was a mistake in the other PR, but that's merged now so I'll close this. Thanks for picking that up @harrietgrace 🙏🏼

PR #251, opened 2018-05-15
Link to the trigger step doc from the env variables doc

toolmantim:
Also cleans up some of the step type pages to make them a little more consistent.

### DIFF
--- a/pages/builds/environment_variables.md.erb
+++ b/pages/builds/environment_variables.md.erb
@@ -227,25 +227,7 @@ Defining an environment variable at the top of your yaml file will set that vari
 
 **Setting variables in a Trigger step**
 
-Environment variables set in the `env` attribute of a trigger step end up as build-level environment variables on the newly created build.
-
-If you're using a 3.x agent, you can use variable interpolation to set variables for a trigger step that are already set elsewhere in the parent build. For example:
-
-```
-env:
-  FOO: bar
-
-steps:
-  - trigger: some-pipeline
-    build:
-      env:
-        FOO: $FOO # triggered build will have `FOO: bar`
-```
-
-<div class="Docs__note">
-<p class="Docs__note__heading">Environment variables set at top of a pipeline.yaml aren't applied to trigger steps</p>
-<p>The only variables that are passed through to a trigger step's build are those you set in that trigger step's `env` attribute.</p>
-</div>
+Environment variables are not automatically passed through to builds created with [trigger steps](/docs/pipelines/trigger-step). To set build-level environment variables on triggered builds, set the trigger step’s `env` attribute.
 
 ### Agent environment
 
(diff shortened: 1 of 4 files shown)


pages/pipelines/trigger_step.md.erb
```diff
@@ -1,13 +1,10 @@
 # Trigger Step
 
-A *trigger* step creates a build on another pipeline. You can use trigger steps to separate your test and deploy pipelines, or to create build dependencies between pipelines.
+A *trigger* step creates a build on another pipeline. You can use trigger steps to separate your test and deploy pipelines, or to create build dependencies between pipelines. Trigger steps can only be defined in your [pipeline.yml](/docs/pipelines/uploading-pipelines) file
```
harrietgrace: Missing a full stop at the end 😊

pages/pipelines/trigger_step.md.erb
```diff
@@ -127,7 +127,12 @@ _Optional_ `build` _attributes:_
     branch: "${BUILDKITE_BRANCH}"
 ```
 
-<div class="Docs__note">
-<p class="Docs__note__heading">Environment variable support</p>
-<p>Buildkite Agent 3.0 or above is required to use pipeline.yml environment variable substitution. See the <a href="/docs/agent/v2/cli-pipeline#environment-variable-substitution">pipeline command documentation</a> for more information.</p>
-</div>
```
harrietgrace: Do you think enough people are on the v3 agent to remove this note? 

Related thought: It would be good to have some note somewhere that tells people we assume you're on the latest stable agent, or something along those lines.

pages/pipelines/trigger_step.md.erb
```diff
@@ -127,7 +127,12 @@ _Optional_ `build` _attributes:_
...
+To set environment variables for the triggered build itself, use the `env` attribute:
```
harrietgrace: Could we reword this a little? I found there was a bit of mental overhead when reading 'for the triggered build itself'.

> To set environment variables on the build created by the trigger step, use the `env` attribute:

pages/builds/environment_variables.md.erb
```diff
@@ -205,25 +205,7 @@ Defining an environment variable at the top of your yaml file will set that vari
...
-    build:
-      env:
-        FOO: $FOO # triggered build will have `FOO: bar`
-```
-
-<div class="Docs__note">
-<p class="Docs__note__heading">Environment variables set at top of a pipeline.yaml aren't applied to trigger steps</p>
-<p>The only variables that are passed through to a trigger step's build are those you set in that trigger step's `env` attribute.</p>
-</div>
+Environment variables are not automatically passed through to builds created with [trigger steps](/docs/pipelines/trigger-step). To set build-level environment variables on triggered builds, set the trigger step’s `env` attribute.
```
harrietgrace: Removing all this makes me nervous 😅But I can't come up with a good reason not to do it, so 👍🏻

pages/pipelines/trigger_step.md.erb
toolmantim: Heaps better! Updated.

pages/pipelines/trigger_step.md.erb
toolmantim: Whoops! Fixed.

PR #255, opened 2018-05-16
Add file:// plugin URL example

toolmantim:
This adds mention of the `file://` plugin URL format, which people have asked about on support for how to refer to plugins which might have been pre-baked or installed on the agent machine.

It also updates it from "git repository name" to "git repository URL", because that's the [Git lingo for it](https://git-scm.com/docs/git-clone#URLS).


lox:
Nice one @toolmantim

PR #257, opened 2018-05-16
Add plugin hooks information

toolmantim:
This reworks our hooks page to include information about plugin hooks.

The "Available hooks" have been redone to be in a table just like environment variables. Hopefully this makes it much easier to read and scan, and explicitly states the order that the hooks are run in.

list-of-hooks

It also updates the plugin doc to link back to the hooks page in lots of places, because it's important for plugin authors to understand hooks, and how they're run in order and relation to all the other hooks.

You might need to check this one out locally @harrietgrace.

### DIFF
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -16,17 +16,17 @@ Hooks are called in the following order:
 1. `pre-checkout` (global) - runs before checkout
 1. `checkout` (global) - overrides the default `git checkout` behavior
 1. `post-checkout` (global) - runs after checkout
-1. `post-checkout` (local) - runs after checkout
+1. `post-checkout` (pipeline) - runs after checkout
 1. `pre-command` (global) - runs before the build command
-1. `pre-command` (local) - runs before the build command
+1. `pre-command` (pipeline) - runs before the build command
 1. `command` (global) - overrides the default script running behavior, useful for changing how the command is evaluated (such as running it inside a container)
-1. `post-command` (local) - runs after the command
+1. `post-command` (pipeline) - runs after the command
 1. `post-command` (global) - runs after the command
-1. `pre-artifact` (local) - runs before artifacts are uploaded, if an artifact upload pattern was defined for the job
+1. `pre-artifact` (pipeline) - runs before artifacts are uploaded, if an artifact upload pattern was defined for the job
 1. `pre-artifact` (global) - runs before artifacts are uploaded, if an artifact upload pattern was defined for the job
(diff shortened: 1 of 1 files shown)


toolmantim:
I just updated this to the latest master. No content changes though.

harrietgrace:
Mmm, I guess they are 'repository' level rather than pipeline... if you have multiple pipelines for one respository?

toolmantim:
Yeah! And they're checked into your repository. I'm thinking perhaps they should be called "Repository hooks" now…

toolmantim:
Okay, I've updated it to repository hooks:

image

and updated the table:

new-hooks

One thing I've also done though, is that I removed the "running in a container" from the command hook description but changed it from "script running behaviour" to "command running". I think most people's container usage would be via the docker/compose plugins, and that there's lots of plugins that override the command for other purposes (like how the [detect-clown plugin does](https://github.com/buildkite-plugins/detect-clowns-buildkite-plugin/blob/master/hooks/command)). It feels weird not having what it's useful for tho! Maybe we can re-work it?

pages/agent/v3/hooks.md.erb
```diff
@@ -1,40 +1,40 @@
...
-These hooks work on unix-based systems with bash in the path. If you're running Windows agents, you'll need some Windows-specific additions to get your hooks running. See <a href="hooks#hooks-on-windows">Hooks on Windows</a> for a detailed description.</p>
-</div>
+<table>
+  <thead>
+    <tr><th>Hook</th><th>Order</th></th><th>Description</th></tr>
+  </thead>
+  <tbody>
+    <tr><td style="white-space: nowrap"><code>environment</code></td><td style="white-space: nowrap">Agent, Plugin</td><td>Runs before all other commands. Useful for <a href="/docs/agent/v3/securing#using-environment-hooks-for-secrets">exposing secret keys</a> and <a href="/docs/agent/v3/securing#whitelisting">security whitelisting</a>.</td></tr>
+    <tr><td style="white-space: nowrap"><code>pre-checkout</code></td><td style="white-space: nowrap">Agent, Plugin</td><td>Runs before checkout.</td></tr>
+    <tr><td style="white-space: nowrap"><code>checkout</code></td><td style="white-space: nowrap">Agent, Plugin</td><td>Overrides the default <code>git checkout</code> behavior. <em>Note:</em> If multiple checkout hooks are found, only the first will be run.</td></tr>
```
harrietgrace: Should checkout be around the other way? Re: the chats in campfire just now

toolmantim: Doh! Yes, yes it should! Update :) Thanks for the catch.

toolmantim:
Yep! I'm 99% sure. Merge party!!

PR #258, opened 2018-05-16
Update to plugin.yml and add related tools

toolmantim:
This updates all the plugin.json information to the new plugin.yml, and links to all the tools that people might find useful when developing plugins.

### DIFF
--- a/pages/agent/v3/plugins.md.erb
+++ b/pages/agent/v3/plugins.md.erb
@@ -73,27 +73,33 @@ changing unexpectedly. This is done with a git tag of `vX.X.X`.
 
 The best place to get started is the [Buildkite Plugins] repository. You can also find third-party plugins with the [buildkite-plugin tag on GitHub][third party plugins].
 
-At Buildkite, our plugins are predominantly shell scripts, so we've created a [plugin tester] docker image that simplifies writing BATS tests and getting your plugin tested (so meta).
-
 ### Anatomy of a plugin
 
 Plugins are essentially a collection of hooks with extra configuration. The agent parses the plugin configuration and passes it on to the plugin in a slightly normalized form, both as JSON or environment variables.
 
 This means that you can write a plugin as a shell script, or as a binary.
 
-A minimal plugin has a `plugin.json` at it's root:
+A plugin has a `plugin.yml` at it's root that describes what it does, the executables required on the agent machine, and a schema of the plugin’s configuration options:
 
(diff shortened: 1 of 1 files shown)


pages/agent/v3/plugins.md.erb
```diff
@@ -73,27 +73,33 @@ changing unexpectedly. This is done with a git tag of `vX.X.X`.
...
-At Buildkite, our plugins are predominantly shell scripts, so we've created a [plugin tester] docker image that simplifies writing BATS tests and getting your plugin tested (so meta).
-
 ### Anatomy of a plugin
 
 Plugins are essentially a collection of hooks with extra configuration. The agent parses the plugin configuration and passes it on to the plugin in a slightly normalized form, both as JSON or environment variables.
 
 This means that you can write a plugin as a shell script, or as a binary.
 
-A minimal plugin has a `plugin.json` at it's root:
+A plugin has a `plugin.yml` at it's root that describes what it does, the executables required on the agent machine, and a schema of the plugin’s configuration options:
```
harrietgrace: Should we mention that this used to be a plugin.json? This could go in a troubleshooting note at the end of the section?

pages/agent/v3/plugins.md.erb
```diff
@@ -73,27 +73,33 @@ changing unexpectedly. This is done with a git tag of `vX.X.X`.
...
+    image:
+      type: string
+    workdir:
+      type: string
+  required:
+    - image
 ```
 
-And then hook scripts under `hooks/`. A good example is the [docker plugin], which runs a command within a docker container.
+The plugin’s hook scripts exist under `hooks/`.
```
harrietgrace: What do you think about:
> The plugin’s hook scripts are placed in the `hooks/` directory

Just so that it's super clear to less experienced folk?

pages/agent/v3/plugins.md.erb
toolmantim: Hrmmm… good questions. I don't think so. The plugin community is still small, and and I've sent pull requests to many of them already. But actually there's still a few we should help to update… I'll go do that now.

toolmantim:
Thanks @harrietgrace — I've reworked this a little whilst I was here, just trying to improve it a bit for plugin developers. Might need another local checkout review if you could please!

toolmantim:
Thanks @harrietgrace!

PR #260, opened 2018-05-18
Document pipeline upload --no-interpolation

lox:
We introduced a `pipeline upload --no-interpolation` flag in 3.1.1.


harrietgrace:
Sweet, thanks @lox 👌🏻

PR #262, opened 2018-05-23
New sidebar arrangement

harrietgrace:
This shuffles all the pages around into the new sidebar arrangement:
sidebar-layout

Goes along with the PR for all the routes.


harrietgrace:
Thanks for all the feedback @toolmantim! I'm not 100% on the configuration/managing pipelines names either, but I am keen for some separation there. 

The thing I was trying to highlight with those two sections was that some of the docs are about actually doing things with your pipeline and code etc, but some of them are about how to set things up in the UI. Maybe it would work better as subsections?

toolmantim:
Ah interesting, I didn't see the split there between code vs the web UI. Interesting. Yeah there's lots of different components you can "Configure" in lots of different ways. That's gonna be a tricky one! Hrmm.

I can help rubber ducky a sorting exercise if you wanted?

harrietgrace:
After much moving around of pages, I don't think that the code vs web ui split is going to be the best option. Here's my current thinking, based around what people are wanting to do:

image

What are your thoughts @toolmantim?

toolmantim:
Interesting! Yeah, it's a tricky split.

The two that stand out as being odd are Parallel Builds, and Branch Configuration.

I think Parallel Builds is more like a tutorial? And maybe it should be called "Parallelizing Builds"?
https://buildkite.com/docs/builds/parallel-builds

Maybe Docker-Based Builds should also be a tutorial?
https://buildkite.com/docs/builds/docker-containerized-builds

I wonder how that changes things?

harrietgrace:
I like the idea of moving those two into tutorials! I think the rest of the workflows list are still better separate to the top list. 
 
I've been looking at the docker one today actually, I think it needs a few updates too (unrelated to this PR).

harrietgrace:
@toolmantim I'd like to keep permissions where it is - the permissions with teams doc isn't really a tutorial, it describes the permissions levels and gives and example of doing programmatic management. All the other things we've collected into Tutorials walk through steps to do something 😊

toolmantim:
@harrietgrace okay, sounds sensible! 🚀🚀

harrietgrace:
Ur a wizard @toolmantim 🙌🏻 thanks for the updates!

PR #264, opened 2018-05-27
Fixup junit-annotate plugin example.

asford:
Fixes minor error in junit-annotate plugin example, updating to valid plugin version.


toolmantim:
Whoops, thanks for the catch @asford!

PR #270, opened 2018-06-21
Update env var to reflect code

harrietgrace:
Fixed the name of the env var, re: slack convo this arvo 😊

Should we also update the v2 version of this doc?

### DIFF
--- a/pages/agent/v3/configuration.md.erb
+++ b/pages/agent/v3/configuration.md.erb
@@ -24,7 +24,7 @@ debug=true
       <th><code>token</code></th>
       <td>
         The agent token from your organization’s Agents page
-        <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_AGENT_TOKEN</code></p>
+        <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_AGENT_ACCESS_TOKEN</code></p>
       </td>
     </tr>
 
--- a/pages/agent/v3/configuration.md.erb
+++ b/pages/agent/v3/configuration.md.erb
@@ -23,8 +23,8 @@ debug=true
     <tr>
       <th><code>token</code></th>
       <td>
-        The agent token from your organization’s Agents page
-        <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_AGENT_ACCESS_TOKEN</code></p>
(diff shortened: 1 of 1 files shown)


keithpitt:
Doh! So what you had before is actually correct.

`BUILDKITE_AGENT_TOKEN` Is the "registration" token you get from your organization agents page. This token can *only* register new agents (and can see high level stats about queue numbers)

`BUILDKITE_AGENT_ACCESS_TOKEN` is the token the agent uses to communicate to the Agent API after it's been registered. This is the token you need to pass to Docker containers and other processes that want to use the `buildkite-agent` token.

In that Slack thread I think I was talking about the names being too similar.

So maybe we can just change the description to:

"When an agent connects to Buildkite, it exchanges this token for a `BUILDKITE_AGENT_ACCESS_TOKEN`. This is the token that's used for all Agent API communication. When an agent disconnects, that token is revoked and can't be used again"

Something something?

keithpitt:
👍 The other thing to note is that `BUILDKITE_AGENT_ACCESS_TOKEN` is managed by us, and not something the user can change. It's present in the job environment, but it won't be in the agent configuration environment (if that makes sense)?

harrietgrace:
Hmm, maybe it should be mentioned on the `buildkite-agent start` page or somewhere like that? There's a section about agent targetting on that page, we could put in a note about docker and other processes targetting agents? (Is that what they're doing? 😅)

I've put the name of the variable back to what it's supposed to be, and added a little extra to its description.

keithpitt:
@harrietgrace I think what you've got there is probably enough to be honest!

What you could do is add a section in https://buildkite.com/docs/builds/docker-containerized-builds#adding-buildkite-agent-to-the-docker-group around "Mounting the `buildkite-agent` binary from within Docker"

(this bit of code here is good: https://github.com/buildkite-plugins/docker-buildkite-plugin/blob/master/hooks/command#L30)

The 3 required ENV vars that need to be sent from the base machine to the Docker instance are:

`BUILDKITE_JOB_ID`, `BUILDKITE_BUILD_ID` and `BUILDKITE_AGENT_ACCESS_TOKEN`

keithpitt:
(Just gonna remove myself as a reviewer for now until the next round of changes 👍)

harrietgrace:
@keithpitt I think I'll merge this, and add the other access token info to the docker page separately. I'd like to make a few other updates to the docker guide at the same time 😊

PR #277, opened 2018-07-02
Document the “plugins” property of command steps

### LINKED ISSUE #275: Buildkite plugin step
Would it be possible to get some more docs around **https://buildkite.com/docs/agent/v3/plugins**, please?

E.g. can standalone plugin-steps target specific agents?

joscha:
See https://github.com/buildkite/docs/issues/275#issuecomment-400604220

closes #275 

cc @lox

### DIFF
--- a/pages/pipelines/plugins.md.erb
+++ b/pages/pipelines/plugins.md.erb
@@ -54,6 +54,8 @@ steps:
         run: app
 ```
 
+A plugin step is essentially a [command step], as such you can use all properties that are also availble for command steps (including agent targeting, concurrency, etc.).
+
 ## Plugin sources
 
 If you refer to a plugin just by name (say `detect-clowns`), it defaults to
--- a/pages/pipelines/plugins.md.erb
+++ b/pages/pipelines/plugins.md.erb
@@ -5,14 +5,13 @@ Plugins provide a way to extend the functionality that Buildkite natively offers
 
 Plugins can be hosted on GitHub, BitBucket or a private git repository, and are automatically fetched as required. See [Plugin sources] for how you can host your own.
 
-Buildkite offers a [collection of useful plugins][Buildkite Plugins]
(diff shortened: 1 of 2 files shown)


toolmantim:
Thanks so much for the contribution @joscha! We should also mention the `plugin` property in the list of [Command Step Attributes](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) as well. I'll get this fixed up and merged in for you.

pages/pipelines/plugins.md.erb
```diff
@@ -3,73 +3,74 @@
...
+  - command: yarn install && yarn run test
+    plugins:
+      docker#v1.0.0:
+        image: node
+        workdir: /app
+```
 
-A real-life example is the [Docker Compose plugin]. It provides an integration
-with docker-compose that assists with running your builds within a docker environment:
+More advanced plugins, such as [Docker Compose plugin], can be run multiple times within a build, and use [buildkite-agent meta-data] to share data between the steps:
```
harrietgrace: What does 'can be run multiple times in a build' mean? I think we need to explain a bit earlier (before the docker plugin example too), what it means to 'run' a plugin. In the first line we say that the hooks are 'downloaded and run as part of your builds', maybe that's the point where we could be clearer? 

I don't think this comment is very clear either 😅I'm mostly confused by the bits that say that the plugins are hooks, but they're 'run' like a script, but also that there's configuration too and I'm not sure where that comes in/how it's executed/what it's for?

toolmantim: Thanks, they're good questions! I've tried to clarify how that works in the latest commit…

toolmantim:
@harrietgrace thanks! Yeah, usually how it works is that every fresh look at the doc I'll be "Oh darn, so much to improve" — I reckon the whole config part is really key, and totally missing.

PR #280, opened 2018-07-10
Expand on "Migrating existing steps to a pipeline.yml"

ticky:
This adds some extra details to the migration part of the "defining steps" doc.

Previously, we did not include a "wait" step to ensure other steps couldn't run in concert and thus be preserved.

This also adds a screenshot showing the described configuration.

image


ticky:
I’m all for changing names, I think an additional explanation would be awkward. Are you happy with “existing step 1”?

harrietgrace:
Yeah agreed! Yep, that would be great 😄 thanks!

PR #281, opened 2018-07-10
Adds an example of whitelisting plugins

zsims:
This doesn't filter versions, but helps implement the security advice on plugins (e.g. whitelist plugins, and possibly version SHAs)

### DIFF
--- a/pages/agent/v3/securing.md.erb
+++ b/pages/agent/v3/securing.md.erb
@@ -100,6 +100,51 @@ if [[ "${BUILDKITE_PLUGINS:-}" != "" ]]; then
   exit 1
 fi
 ```
+  
+### Whitelisting Plugins
+
+Per above, whitelisting can be done with the [agent’s environment hook](hooks). It may be desirable to whitelist a set of plugins that agents are allowed to execute.
+
+```bash
+#!/bin/bash
+
+set -euo pipefail
+
+pluginWhitelist=(
+  'github.com/buildkite-plugins/docker-buildkite-plugin'
+  'github.com/buildkite-plugins/docker-compose-buildkite-plugin'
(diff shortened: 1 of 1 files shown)


pages/agent/v3/securing.md.erb
```diff
@@ -100,6 +100,51 @@ if [[ "${BUILDKITE_PLUGINS:-}" != "" ]]; then
   exit 1
 fi
 ```
+  
+### Whitelisting Plugins
+
+Per above, whitelisting can be done with the [agent’s environment hook](hooks). It may be desirable to whitelist a set of plugins that agents are allowed to execute.
```
toolmantim: Thanks for providing some updated code and putting the time to put together a PR, I helped write the [initial version](https://gist.github.com/toolmantim/e52f86da6525829cd49557d813830386) of this ☺️🙌🏼

The reason I didn't add it to the docs right away, was that it was untested, and a little too complicated to use as a copy+paste example. And it would at least need some tests.

I've just written tests and published the initial version here (with some fixes):
https://github.com/buildkite/buildkite-plugin-whitelister

Did you want to collaborate on that repo, to integrate your changes? And then we can link to that from the doc, rather than inlining the code?

zsims: Yeah awesome, much better idea. Can also start to do complex things like version pinning that way 😁

zsims: Happy if I update the docs to reference ☝️? Or would you prefer I close this PR? (I'm happy either way)

toolmantim: Yeah, let's update the docs to reference that repo I reckon!

zsims:
Updated with a link to the whitelister instead of the example

PR #282, opened 2018-07-12
Add doc for --tags-from-host from v3.3.0

petemounce:
This was done via #791 and released in v3.3.0. The PR prose refers to a command-line flag that is wrong :(


lox:
Many thanks @petemounce, what did I get wrong in the PR?

petemounce:
I think it was `--add-host-tags` in #791's description/comments.

PR #283, opened 2018-07-12
Alpha-sort agent configuration setting table + add 'required?' column

petemounce:
I find docs easier to read when sorted. Unfortunately:
* this splits apart the semantic-groupings for things like paths
* this makes token last, but it's required, so I added a column to denote that fact

Suggestion - JS-based table-sorting/filtering thing?

### DIFF
--- a/pages/agent/v3/configuration.md.erb
+++ b/pages/agent/v3/configuration.md.erb
@@ -18,14 +18,23 @@ debug=true
 
 ## Configuration Settings
 
+_Required attributes:_
+
 <table>
-  <thead>
-    <tr>
-      <th>setting</th>
-      <th>purpose</th>
-      <th>required?</th>
+  <tbody>
+   <tr>
+      <th><code>token</code></th>
+      <td>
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks @petemounce! We'll check this out.

harrietgrace:
Yay sorting! I find these tables way better when they're alphabetised too 😄
To bring this doc in line with our other tables, rather than add an extra column, we could split it out into required/optional like we do in our [Step Type docs](https://buildkite.com/docs/pipelines/command-step)?

petemounce:
@harrietgrace I thought about that and elected to not, because the effect of the sorting is to make `token` (a required attribute) nowhere near the first thing a reader sees, and so the possibility of their not noticing is much higher I thought.

Making a separate column allows that information to be more obviously distinct to a reader, and also makes it easier to add some client-side table plugin at some point that allows filtering/sorting.

harrietgrace:
I'll do the table update - I'll add a commit to your branch before we merge, if that's alright @petemounce?

petemounce:
Please feel free!

PR #284, opened 2018-07-12
Add docs for tags-from-host

petemounce:
Done in #791, released inside `v3.3.0`, but PR mentions flag in prose that doesn't match flag in code.


petemounce:
Closing, #282 did this. Brainfart.

PR #295, opened 2018-08-15
Mention BUILDKITE_GCS_ACCESS_HOST

petemounce:
Didn't see this in docs, added it. Please obviously feel free to hijack and extend :)


harrietgrace:
Thanks @petemounce! 🙏🏻✨ I've added a few extra bits about other permissions options.  
While I was poking around in the agent code @sj26 and I found there were a bunch of other env vars we have available, so I added them to the page as well 🙌🏻

PR #297, opened 2018-08-16
Fix agent tag command step link

toolmantim:
The [command step attributes](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) incorrectly reference the build meta-data docs, instead of the agent tags docs (probably from when agent meta-data was renamed to agent tags). This updates the link and words to point to the new [setting tags](https://buildkite.com/docs/agent/v3/cli-start#setting-tags) doc, and keeps the link to the targeting section as well.


toolmantim:
Those sneaky links.

PR #300, opened 2018-08-17
doc: fix various typos

StevePorter92:


harrietgrace:
This is so awesome 🤩 Thanks a bunch for all these fixes @StevePorter92! I've added in some more of the same ones from other parts of the docs 👍🏻

We do our best to use American English spelling, so unfortunately I had to roll back the 'u's in behaviour and flavoured 😔This also lead me to find a bunch more of the British English version that had slipped through in past additions, so thanks for that! 🙌🏻

StevePorter92:
No worries at all! Glad to help 😄

PR #302, opened 2018-08-20
Add self-serve SSO guide using GraphQL

harrietgrace:
Adds a guide about how to set up SSO with GraphQL 🎉


harrietgrace:
@keithpitt it would be great to get your feedback on the layout of the doc, and the accuracy of the explanations/examples 😄If it's all looking reasonable, I'll make all the dot points into full sentences 👌🏻

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,289 @@
...
+### Get your internal Organization ID
+
+- to set up your organisation for SSO, you need your internal Buildkite id
+- use the get method with your org slug to retrieve your org id
+- this id will be required in all other sso methods
+
+For example, the following code will retrieve any SSO providers that have been created for the organisation `Acme-Inc`:
+
+```graphql
+query GetProviders {
```
lox: @keith suggested that we add an Organization ID field to the Organization settings page, so we can drop this query.

lox:
Given the complexity of supporting the variety of different providers, what would you think of splitting into specific guides for setting up SSO with SAML, and SSO with (Github|Google)?

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,289 @@
...
+      }
+    }
+  }
+}
+```
+
+- test the login flow, ensure that a user can successfully complete the new authentication and gain access to your buildkite organisation before continuing
+- the enable method will error if a successful test hasn't occurred
+
+### Enable the provider for your organisation
```
lox: Would be good to mention that Enabling SSO marks all current users as requiring SSO. The user will likely want to mark external staff and robot users as not needing SSO manually after this.

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,289 @@
...
+  identityProvider:{
+    metadata:{
+      xml: "..."
+    }
+  }
+})
+```
+
+xml url options:
+- identity provider (metadata > url)
```
lox: I reckon this one would be good to roll into the main "creating a SSO provider" example, as it should be the main use case.

pages/tutorials/graphql_sso.md.erb
harrietgrace: Oh cool, how does one go about marking people as not needing SSO? I thought that it was an all-or-nothing kind of thing.

pages/tutorials/graphql_sso.md.erb
harrietgrace: I thought that people would still be creating a blank provider to start with? Don't people need to get the service provider information first, before they can complete their identity provider setup?

pages/tutorials/graphql_sso.md.erb
harrietgrace: Sweet, I've moved the code block down the end to the 'other operations' section so people can view the providers that they've created

toolmantim:
Alright @harrietgrace — I've just done a retake of the guide.

image

Can you review it and let me know what you think / find? 🙏🏼

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,250 @@
...
+
+Buildkite’s SSO login can be set up by emailing support, or you can set it up manually using our [GraphQL APIs](/docs/apis/graphql). This tutorial covers how to set up SSO manually using GraphQL, for most of the common SSO providers.
+
+For details on every available option available in the GraphQL APIs, please use the documentation sidebar built into the [GraphQL Explorer](/docs/graphql-api#getting-started).
+
+<%= toc %>
+
+## Finding your organization’s GraphQL ID
+
+For every type of SSO provider, you’ll need your Buildkite organization’s GraphQL ID. You can find the organization `id` using the following GraphQL query:
```
harrietgrace: In all the other examples, we've got the name of the query, what do you think about adding that here rather than 'the following'?

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,250 @@
...
+query OrgId {
+  organization(slug: "myorg") {
+    id
+  }
+}
+```
+
+## Setting up G Suite
+
+The first step in setting up a G Suite SSO provider is to use the `ssoProviderCreate` mutation.
```
harrietgrace: Could we bring back the line here that explains what the example code will do? I guess the audience of this doc is going to be familiar with graphql... but I'd still think that explaining examples reduces the mental overhead for everyone.

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,250 @@
...
+      state
+      url
+    }
+  }
+}
+```
+
+The second step is to perform a test login by visiting the `url` returned from the create mutation.
+
+Once a test login has been completed, you can do the final step: enabling the provider using the `ssoProviderEnable` mutation. Running this mutation will require all your users to login via your SAML provider before they can access your organization on Buildkite.
```
harrietgrace: Google is OAuth rather than SAML 😊

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,250 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+The next step is to log into your SSO provider’s system and setup Buildkite using the `issuer`, `ssoURL` and `metadata` returned above. Once setup, your SSO provider should provide a metadata URL that you can give to Buildkite.
+
+Using the metadata URL returned from your SSO provider, call the `ssoProviderUpdate` mutation to have Buildkite automatically retrieve the details and set it up, ready for a test login. This will not yet affect any of your Buildkite users.
```
harrietgrace: Do you `call` mutations or are they `run`?

pages/tutorials/graphql_sso.md.erb
```diff
@@ -0,0 +1,250 @@
...
+}
+```
+
+You should now see that the provider’s state is enabled.
+
+See the `SSOProviderUpdatePayload` documentation for other properties that can be configured on your SSO provider, such as `sessionDurationInHours` and `note`.
+
+## Finding an SSO provider’s details
+
+If you need to find the id of a particular SSO provider, you can query the `ssoProviders` field of your organization:
```
harrietgrace: What's the use case for this? I like the section, I just don't understand what people would be doing that makes them come to this page and click this heading.

toolmantim:
Thanks so much for the review @harrietgrace! I'll do another round of edits on the document first thing next week.

pages/tutorials/graphql_sso.md.erb
toolmantim: Ah, the other examples show the mutation/field names. This one is a bit different, because `OrganizationId` isn't anything in our API at all… it's just the label we've given the query:

```graphql
query OrganizationId {
  organization(slug: "myorg") {
    id
  }
}
```

It's actually optional, and I'm tempted to remove it:

```graphql
query {
  organization(slug: "myorg") {
    id
  }
}
```

but then, if you copy + paste it into the GraphQL Console/Explorer, you can't choose between the multiple queries/mutations if they don't have a label. Argh!

pages/tutorials/graphql_sso.md.erb
toolmantim: Cool yep, I'll fix that now!

pages/tutorials/graphql_sso.md.erb
toolmantim: Oh, good catch!

pages/tutorials/graphql_sso.md.erb
toolmantim: Oh yeah, run is way better!

pages/tutorials/graphql_sso.md.erb
toolmantim: Ah. So mainly it's if you go off-piste, and you run another GraphQL query during the setup process and lose the response from a previous mutation. Because they're create/updates, you can't just re-run them… so you have to go and find the details (like the ID) again yourself mainly.

But instead of just the `id`, I just added a whole bunch of properties that are queried, for other situations as well, like seeing whether the provider is activated (which affects all your users), and other tasks like that.

If we had a proper SSO Settings page where you could manage it all yourself, we probably wouldn't need this section.

pages/tutorials/graphql_sso.md.erb
toolmantim: I tweaked the words to try to make it clearer…

PR #319, opened 2018-10-01
Adds a section on plugin ordering

zsims:
Per the current plugin documentation, it wasn't clear that there's an array syntax for plugins -- this is especially important where plugin ordering is important.

I'm happy to update this when https://github.com/buildkite/agent/pull/824 lands

### DIFF
--- a/pages/pipelines/plugins.md.erb
+++ b/pages/pipelines/plugins.md.erb
@@ -59,19 +59,57 @@ You can see the [full list of available Buildkite Plugins](https://github.com/se
 
 ## Plugin ordering
 
-When using the array syntax, the hooks within each plugin will be executed in the order they're referenced. For example:
+Plugins will be run in the same order they were defined in the `pipeline.yml`. For example, the following pipeline step runs the `docker-login` plugin before the `docker-compose` plugin:
 
-```
+ ```yaml
 steps:
-  - label: "\:clown_face\:"
+  - name: "\:docker\: Push"
     plugins:
-      - detect-clowns#v1.0.0: ~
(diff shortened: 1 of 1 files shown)


lox:
Awesome, merged buildkite/agent#824 too.

lox:
Might be worth noting that in `3.5.0` we released a fix that ensured the map syntax had the correct order, but we still recommend using the array syntax to be sure?

zsims:
Done! Let me know if I can tweak wording

toolmantim:
Just letting you know that we haven't forgotten this PR @zsims!

I've just made some updates/changes after some discussion with @harrietgrace. We're also looking into whether the standard/default syntax for plugins should be lists/arrays (and whether we should do that across all our docs and plugin readmes), but this change doesn't assume that just yet.

We shouldn't merge this until v3.5.0 is released anyhow, so we'll keep working on this until that's ready.

Thanks @zsims ❤

zsims:
v3.5.0 is out now -- can this be merged @toolmantim ?

toolmantim:
Rather than call with out @zsims we actually went all-in on the array syntax. #323 migrated all the docs to use it, the buildkite-plugins/buildkite-plugin-linter#155 added it to the linter, and we're updating all the plugin readmes to the new array syntax.

We've still got a bit of work to do, including publishing a changelog giving people a heads up, but I think we can close this PR. Thanks for taking the time to go through it on both the agent and docs PRs! Probably wouldn't have happened without your input 🙏🏼

zsims:
Awesome! Great outcome 😁

PR #321, opened 2018-10-09
Update the location of buildkite-agent.cfg for GKE clusters

blakestoddard:
`buildkite-agent.cfg` is actually located at `/etc/buildkite-agent/buildkite-agent.cfg`, not `/buildkite/buildkite-agent.cfg`.

This took me a pretty fun half-hour of scratching my head to figure out why it wasn't working.

### DIFF
--- a/pages/agent/v3/gcloud.md.erb
+++ b/pages/agent/v3/gcloud.md.erb
@@ -211,7 +211,7 @@ volumes:
 
 ### Further configuration
 
-To [configure](/docs/agent/v3/configuration) the agent further you can create a [config map](https://kubernetes.io/docs/user-guide/configmap/) and volume mount it over the default agent configuration file in `/buildkite/buildkite-agent.cfg`.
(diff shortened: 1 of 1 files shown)


toolmantim:
Oh, that is strange! I thought myself that we set the `BUILDKITE_AGENT_CONFIG` path to be `/buildkite/buildkite-agent.cfg`, but it doesn't appear so:

* https://github.com/buildkite/agent/blob/master/packaging/docker/alpine-linux/Dockerfile#L22
* https://github.com/buildkite/agent/blob/master/packaging/docker/ubuntu-linux/Dockerfile#L23

For GKE, do you think it would it be better to set the `BUILDKITE_AGENT_CONFIG` to a path you'd prefer? Or to tell people to use one of the default search paths? (like this PR does)

blakestoddard:
I’d rather have it in /buildkite to keep all of the configuration together, if possible. Seems cleaner that way.

PR #325, opened 2018-10-30
Explain how our GraphQL IDs work

toolmantim:
This adds info about how our GraphQL IDs work, so we can refer to them in guides such as https://buildkite.com/docs/tutorials/sso-setup-with-graphql


pages/apis/graphql_api.md.erb
```diff
@@ -55,6 +55,58 @@ curl https://graphql.buildkite.com/v1 \
...
+{
+  "data": {
+    "organization": {
+      "id": "T3JnYW5pemF0aW9uLS0tYTk4OTYxYjctYWRjMS00MWFhLTg3MjYtY2ZiMmM0NmU0MmUw"
+    }
+  }
+}
+```
+
+If you have a GraphQL ID, you can use it with the global `node` query. You’ll need to specify the type using an [Inline Fragment](https://graphql.org/learn/queries/#inline-fragments) to access type-specific properties.
```
harrietgrace: What does the global node query do? I was planning to suggest swapping the order of the sentence around, but realised I don't know what the node query is doing 😄Starting with 'if you have a graphql id' sounds a bit like there's some circumstances when there isn't an id available to you, as opposed to you just not having found it yet. 

Something like: "To find <result of performing this query>, use your graphql id with the global 'node' query".

toolmantim: Oh yeah! What about:

> You can look up any GraphQL node using its `id` and the global `node` query. To return specific properties of the object, you’ll need to specify the object’s type using an [Inline Fragment](https://graphql.org/learn/queries/#inline-fragments).
>
> For example, …

(for context, you'll probably need to use the fragment syntax 99.999% of the time)

pages/apis/graphql_api.md.erb
```diff
@@ -55,6 +55,58 @@ curl https://graphql.buildkite.com/v1 \
 
 For documentation on the full list of fields and types, see the interactive documentation in the [GraphQL Explorer](#getting-started).
 
+## GraphQL IDs
+
+All node types have an `id` property, which is a global identifier encoding the object’s type and UUID. Buildkite’s GraphQL IDs are based on the [Relay Global Object ID](https://facebook.github.io/relay/graphql/objectidentification.htm).
```
matthewd: I've read through https://facebook.github.io/relay/graphql/objectidentification.htm twice now, and find it pretty tough going... largely because it's written as a specification for a server to comply with, rather than a documentation of the protocol a client can use (which are technically the same thing, but the former warrants a far more formal approach).

I'm also slightly unsure about how this implicitly subordinates our GraphQL API definition to Relay, which is a specific client library. :confused:

toolmantim: Yeah, it's a really good point.

There's definitely people who will want to use a Relay client to talk to Buildkite. But that probably wasn't the audience for this section of the documentation… this was so we can link the "GraphQL ID" words on the [SSO Setup Guide](https://buildkite.com/docs/tutorials/sso-setup-with-graphql) to this new section.

Given that our ids are base64 encoded `{type}-{uuid}` Relay compatible ids, I wonder if we should still mention it though? Maybe not, and we just omit it entirely. We mention Relay in the section below anyhow.

@keithpitt @plasticine do GraphQL clients rely on being able to base64 decode the IDs, or are they meant to treat them as being completely opaque?

keithpitt: > @keithpitt @plasticine do GraphQL clients rely on being able to base64 decode the IDs, or are they meant to treat them as being completely opaque?

They're supposed to just be rando strings, but they're unique to each object, and be used with the `node` query. Base64 encoding them is just the way we do it (and others/easiest). We could encrypt data into them and decrypt it server side and still be 100% compatible with the spec.

toolmantim: Cool thanks! I'll go rework this a bit more then 👌🏼

pages/apis/graphql_api.md.erb
harrietgrace: Yeah that sounds great!

PR #326, opened 2018-11-02
Small word updates to the meta-data doc

toolmantim:
I found this small fixup in my git stash, from I-can't-remember-when, and it still seemed like an okay change? It just rewords the "practical examples" because there are still some practical examples on these docs pages.

I'm happy for you to decide if the change is right or wrong @harrietgrace! But thought I'd push it up here before I `rm` it.


toolmantim:
🙌🏼 Thanks @harrietgrace

PR #334, opened 2018-11-16
block steps: Added note about different format for different numbers of option fields.

bwplotka:
Hello :wave:  

This would be nice to have on docs (: Unless there is more to it than I just found.

However, I need from you this magic `X` number. 

Let me know if that makes sense.

### DIFF
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -117,6 +117,11 @@ steps:
 
 ## Select Input Attributes
 
+<div class="Docs__note">
+<p class="Docs__note__heading">Format</p>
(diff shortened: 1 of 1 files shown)


harrietgrace:
Hey @bwplotka 👋🏻Thanks for taking the time to PR! We love having new contributors 🎉

The different formats are listed a little further down the page under the 'options' attribute in the 'Required Attributes' section of 'Select Input Attributes' 😄

image

We should totally make this more obvious though! I've added another commit that puts that info front and center in the introduction section at the top of the page 👍🏻

bwplotka:
Wow, I missed that 0.o 

Thanks for sorting this out. (:

PR #336, opened 2018-11-22
HTML available in annotations

benschwarz:
Hey! 

The docs page says that markdown is supported, but most of the examples in the page show HTML. I suspect that the reason HTML is supported is because you can put basic tags into markdown, but maybe it would be good to highlight this to readers? 

Should the supported tags be listed?


harrietgrace:
Hey @benschwarz 👋🏻 Thanks for taking the time to PR! 

I love the idea of giving people more info about what's supported in annotations! I've added a bit more to your PR to mention CommonMark and link to their spec. Because their HTML parsing is a bit different to usual, I've also added a note about how it treats newlines.

benschwarz:
Yay! This is good.

PR #340, opened 2018-11-26
Add annotation limit info

### LINKED ISSUE #337: Documentation for annotation limit
Hey! 

I wrote to support regarding annotations and learnt that while many annotations can be added to a build, only the first five are displayed. 

I spent a bunch of time reviewing my pipeline thinking that I'd done something wrong there — What do you think about a change to the docs to explain that only the first five will be displayed?

harrietgrace:
Adding a note about how many annotations will display in the build for each step. Closes #337.

@eleanorakh I got this info from your support ticket, just wanted you to have a squiz before I merge 😄

### DIFF
--- a/pages/agent/v3/cli_annotate.md.erb
+++ b/pages/agent/v3/cli_annotate.md.erb
@@ -70,14 +70,14 @@ Annotations are intended to group messages of the same type. You can add more da
 For example, a 'test-failures' annotation can be created to track failed tests:
 
 ```
-buildkite-agent annotate "Failed tests:" --context test-failures
+buildkite-agent annotate --context test-failures --style "error" "# Failed tests:\n\n"
 ```
 
-Using the `--append` option, the 'test-failures' annotation is updated with each failure as they occur: 
+Using the `--append` option, the 'test-failures' annotation is updated with a failure when it occurs: 
 
 ```
 if test-fails
-  buildkite-agent annotate "Test 6 failed with exit code 1: <error>" --context test-failures --append
+  buildkite-agent annotate --context test-failures --append "* Test 1 failed :sad_face:\n"
 ```
 
 ## Annotation styles
--- a/pages/agent/v3/cli_annotate.md.erb
+++ b/pages/agent/v3/cli_annotate.md.erb
@@ -1,6 +1,6 @@
 # buildkite-agent annotate
 
-The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown. 
+The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown. Up to five annotations can be displayed on each step in your pipeline.
 
 <%= image "overview.png", alt: "Screenshot of annotations with test reports" %>
 
--- a/pages/agent/v3/cli_annotate.md.erb
+++ b/pages/agent/v3/cli_annotate.md.erb
@@ -1,12 +1,18 @@
 # buildkite-agent annotate
 
-The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown. Up to five annotations can be displayed on each step in your pipeline.
+The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown.
 
 <%= image "overview.png", alt: "Screenshot of annotations with test reports" %>
 
 <%= toc %>
 
 ## Creating an annotation
+
+The `buildkite-agent` cli's `annotate` command creates an annotation associated with the current build.
+
+The first five annotations created will be displayed on the build page, however there is no limit to the amount of annotations you can create. All annotations can be retreived using the [GraphQL API](/docs/apis/graphql-api).
+
+Options for the `annotate` command can be found in the `buildkite-agent` cli help: 
 
 ```
 $ buildkite-agent annotate --help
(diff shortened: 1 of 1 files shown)


eleanorakh:
>Only the last five annotations will be displayed on the build page

👌🏻 
Sorry I missed this yesterday!

harrietgrace:
Added a bunch more info about grouping annotations and such. @toolmantim do you think the new groups section needs a screenshot, or is it okay as it as?

toolmantim:
The grouping section brings up quite a few questions I think!

My understanding was that `--context` is more of an identifier that can be used to either replace or append an existing (single) annotation, more than a mechanism for grouping multiple annotations of the same type. Grouping brings up other ideas, such as `--group "Some Group"` which would help to group together multiple annotations… vs `--context` which I think acts as an identifier, and targets only a single annotation?

Or is the idea that we're trying to show people how to work around build annotation limits by using `context` and `append` to update a single annotation? If that's the case, I wonder if there's another way of framing it more directly. Something more like:

> If you need to add more content, but want to stay under the display limit of 5 annotations, you can use the `--context` and `--append` arguments to update a single annotation with more content. For example, ...

Perhaps the example needs to be showing each one with a separate header?

```bash
buildkite-agent annotate --context test-failures --style error --append "# Model spec failures:\n\n..."
buildkite-agent annotate --context test-failures --style error --append "# Feature spec failures:\n\n..."
buildkite-agent annotate --context test-failures --style error --append "# Integration spec failures:\n\n..."
```

harrietgrace:
Hmm, thanks for explaining your perspective, I hadn't thought about this section in either of those ways! 

What I was going for was more like your second suggestion, of getting around the display limit. I didn't want it to be a 'workaround', rather I wanted to phrase it as a 'this is how this feature was intended' if that makes sense? In the code sample I wanted to show a single annotation being created and then appended, so that people could see an example of ~why they would want to use append rather than ~how to use it. 

I'll have a think about how I could rephrase this section to better communicate this idea!

toolmantim:
Cool! Yeah, going from why rather than how is a great idea.

Another idea on approaching the “why” could be to add more examples, showing the different outcomes you might be wanting to achieve, and then how to achieve them, rather than words. Perhaps:

* A coverage report using a `wait` after some test steps.
* A list of test failures, appended/streamed from parallel jobs as they run
* A big grid of screenshots (artifacts) to review (ala @benschwarz’s email example he might be able to share, which is what raised the issue)

That way people can use them as a conceptual base to approach the problem they’re solving? “Like that, but different”.

Happy to help with the examples! We’ve got all of them somewhere…

toolmantim:
There’s also quite a few plugins that implementation annotations too! (like the two test failure ones) Might be worth linking to, so people can read their source or get ideas?

benschwarz:
> A big grid of screenshots (artifacts) to review (ala @benschwarz’s email example he might be able to share, which is what raised the issue)

Yeah, absolutely!

screen shot 2018-12-07 at 8 33 10 am

I wasn't able to (without a significant amount more effort) really do what I wanted to with the annotation view, but I think the screenshot represents 80% of where I wanted to get with it. 

One thing that comes to mind (but is out of scope for this issue), is that it might be useful to have a command to append to an existing annotation. (Right now, I build up a huge chunk of markup, then inject it as one piece later on). 

@harrietgrace, for the section where you've mentioned the GraphQL API. Do you think it'd be possible to have a code example that'd give someone the ability to view their annotations? Just as a starting point this would potentially really help people make the next step. 

I can't currently come up with any reasons for why I'd want to access annotations that can't be viewed on the page though, so maybe it's not a big deal?

toolmantim:
Thanks @benschwarz! Did you see the `--append` argument?

benschwarz:
> Thanks @benschwarz! Did you see the --append argument?

I think was when I wrote to support, that the suggestion was made to add everything in one annotation, but I didn't realise that the `--append` argument was even a thing. I just went and added it all as one annotation. Nevermind!

pages/agent/v3/cli_annotate.md.erb
```diff
@@ -1,13 +1,19 @@
...
-The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown. 
+The Buildkite Agent’s `annotate` command allows you to add additional information to Buildkite build pages using CommonMark Markdown.
 
 <%= image "overview.png", alt: "Screenshot of annotations with test reports" %>
 
 <%= toc %>
 
 ## Creating an annotation
 
+The `buildkite-agent` cli's `annotate` command creates an annotation associated with the current build.
```
toolmantim: I don't think we refer to `cli` anywhere for the agent, and might be confused with https://github.com/buildkite/cli — perhaps just this?

```suggestion
The `buildkite-agent annotate` command creates an annotation associated with the current build.
```

pages/agent/v3/cli_annotate.md.erb
```diff
@@ -57,6 +63,23 @@ Options:
    --debug-http                Enable HTTP debug mode, which dumps all request and response bodies to the log [$BUILDKITE_AGENT_DEBUG_HTTP]
 ```
 
+## Grouping annotations
```
toolmantim: Perhaps we leave this section out for now and merge the mention of the 5 annotation build page limit, or look at a "Updating and appending annotations" section instead?

harrietgrace: Sounds good!

harrietgrace:
@toolmantim I've pulled out the grouping section for the moment, I'll add it in another PR some other time. Also merged with master, so this should be ready to go 👍🏻

PR #342, opened 2018-12-04
Add the pull request params to the Builds API docs

toolmantim:
Adds the pull request params to the [Create a Build](https://buildkite.com/docs/apis/rest-api/builds#create-a-build) documentation:

image


toolmantim:
"Manually triggering a pull request build" should totally be something in our docs, especially if we add the environment variable support like we're discussing! That's an awesome idea.

Without it though, it's still a bit of an advanced use case. I think it's only for people who are accepting GitHub webhooks themselves, doing some processing, and then triggering the Builds API… because our default GitHub configuration gear isn't suitable for some reason. "I know you can run pull request builds in the normal Buildkite ↔ GitHub setup. How do we replicate that with our custom setup?"

PR #344, opened 2018-12-06
Update wait docs

### LINKED ISSUE #222: continue_on_failure wait step needs a tilde (~)?
https://buildkite.com/docs/pipelines/wait-step

```
- command: "command.sh"
- wait: ~
  continue_on_failure: true
- command: "echo This runs regardless of the success or failure"
- wait
- command: "echo The command passed"
```

is listed as an example of a wait step-including-pipeline that continues. Is the `wait: ~` mandatory? What if we attached a blank key like
```
- command: "command.sh"
- wait: 
  continue_on_failure: true
- command: "echo This runs regardless of the success or failure"
- wait
- command: "echo The command passed"
```

Instead? Should that still continue? It's not clear according to the docs what role `~` has here.

Buildkite team can check this pipeline: https://buildkite.com/gusto/buildkite-superpull/builds/95#772d22ee-e9f1-44b7-a63b-f89577f91624 for an example of a `wait:` key with `continue_on_failure` set that doesn't have`~`

harrietgrace:
Adds a quick explainer about the `~` in the continue on failure example. Mentions that it's an explicit null character, that it's not required, and that we recommend its use to prevent other things getting in between the wait and the continue lines. 

I'm a little unsure of the wording I've used, do you have any feels about it @toolmantim?

Closes #222

### DIFF
--- a/pages/pipelines/wait_step.md.erb
+++ b/pages/pipelines/wait_step.md.erb
@@ -32,3 +32,5 @@ If the previous steps failed, the build will be marked as failed only after the 
 ```
 
 In this example, if `command.sh` succeeds, both of the following command steps will be run. If `command.sh` fails, only the first will be run, and the build will then be marked as failed.
+
+The explicit null `~` syntax used in the above example isn't required, but is reccommended as a best practice. It ensures that nothing else is accidentally added to the `wait` before the `continue_on_failure` attribute.
--- a/pages/pipelines/wait_step.md.erb
+++ b/pages/pipelines/wait_step.md.erb
@@ -33,4 +33,4 @@ If the previous steps failed, the build will be marked as failed only after the 
 
 In this example, if `command.sh` succeeds, both of the following command steps will be run. If `command.sh` fails, only the first will be run, and the build will then be marked as failed.
 
-The explicit null `~` syntax used in the above example isn't required, but is reccommended as a best practice. It ensures that nothing else is accidentally added to the `wait` before the `continue_on_failure` attribute.
+The explicit null `~` character used in the above example isn't required, but is reccommended as a best practice. It ensures that nothing else is accidentally added to the `wait` before the `continue_on_failure` attribute.


toolmantim:
Do we have any YAML examples of what errors/situations it tries to protect you from?

keithpitt:
I believe this is also valid:

```yaml
- wait: true
  continue_on_failure: true
```

I initially used the `~` character because I just thought it looked nicer, and was a bit more "correct". It essentially is saying "This key exists, and I have nothing further to say"

These examples aren't valid YAML:

```yaml
- wait:
  continue_on_failure: true
```

```yaml
- wait
  continue_on_failure: true
```

A longer explanation around `~` would be something like:

Buildkite looks at the keys defined in a step to infer what sort of step it is. For example, if Buildkite sees a `command` key, we assume it's a `command` step. And we use the contents of that key as the command to run on an agent.

```
steps:
  - command: "blah.sh" # This object has a key "command", therefor we treat it as a "command" step
```

Wait steps are special because you can just use a literal string "wait" to define a wait step:

```
steps:
  - wait
```

The `wait` key itself doesn't require any configuration, so when defining a wait step with additional properties, you need to set (something) as the value since without a valid, it's invalid YAML. We chose `~` because it's short and sweet, but `true` works (as well as your favourite emoji:)

```
steps:
  - wait: true
    continue_on_failure: true
  - wait: ~
    continue_on_failure: true
  - wait: "I like pie"
    continue_on_failure: true
  - wait: ⚽️
    continue_on_failure: true
```

keithpitt:
(there's a brain dump on stuff, feel free to chop/change/ignore all the things!)

toolmantim:
Thanks for the braindump! Every validator I've tried finds this to be valid YAML, which is what had me confused:

```yaml
- wait:
  continue_on_failure: true
```

lox:
Yup, totes valid yaml! Perhaps it was an issue with the golang yaml parser initially @keithpitt? For instance, it doesn't actually understand that `~` is `null` 😅

keithpitt:
> Thanks for the braindump! Every validator I've tried finds this to be valid YAML, which is what had me confused:
> 
> - wait:
>   continue_on_failure: true

You're 100% correct, that is valid YAML. I think my computer busted itself, and I was having Golang memories (as @lox suggested).

So you're right, this is also totally valid:

```
- wait:
  continue_on_failure: true
```

keithpitt:
It's completely server side, so no agent agents. Do you mean, make the value-less the new default?

toolmantim:
Yeah, value-less as the default would be ❤️ I think!

keithpitt:
I think I prefer the semantics of:

```yaml
  - wait: ~
    continue_on_failure: true
```

Mostly because the value-less option just looks miss-configured:

```yaml
- wait: 
  continue_on_failure: true
```

Feels to me like it's missing some configuration, and doesn't read *as nice*. At least the `~` is *something*, and once peeps learn what it means, it makes sense. It also looks a lot like this:

```yaml
- wait: 
    continue_on_failure: true
```

Which is still a valid pipeline, but that syntax won't be around forever (since it's a little confusing). I'm also unsure what auto-indenters/formatters in editors do with the value-less option. Since whitespace is significant, it's probably hard for editors to know what the intention is.

I'm happy to be convinced that a value-less option is better though!

harrietgrace:
I think I'm most keen on the `- wait: ~` or having the second line indented. Having nothing after the wait, and then having the next line be on the same indent level makes it look like something is missing to me.

harrietgrace:
@keithpitt @toolmantim some more thoughts on the possible options for wait syntax - this is our only step type that doesn't indent its attributes. Maybe we should go for the indented version so that it's more intuitive?

toolmantim:
Now we have a similar nesting in our new plugin syntax, as far as left whitespace indenting goes, perhaps we should settle on the nested one? That’s the one you’re suggesting yeah @harrietgrace?

```yml
- wait: 
    continue_on_failure: true
```

harrietgrace:
Yep, that's the one I was talking about @toolmantim 👍🏻

PR #345, opened 2018-12-10
Clarified -1 handling and retries

keithpitt:
Not too sure of the wording, just wanted to open a PR up with the correct rules so it can be adjusted and merged.

### DIFF
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -203,7 +203,7 @@ _Optional Attributes_
 
 <div class="Docs__note">
 <p class="Docs__note__heading">-1 Exit Status</p>
-<p>An exit status of -1 results when a job loses connection with the agent running it. Jobs that exit with -1 will be automatically retried, unless you have overridden the retry behavior in the <code>pipeline.yml</code> file.</p>
+<p>A job will show an exit status of -1 when we loose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
 </div>
 
 ```yml
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -203,7 +203,7 @@ _Optional Attributes_
 
 <div class="Docs__note">
 <p class="Docs__note__heading">-1 Exit Status</p>
-<p>A job will show an exit status of -1 when we loose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
+<p>A job will show an exit status of -1 when we lose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
 </div>
 
 ```yml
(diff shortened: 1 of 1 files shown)


pages/pipelines/command_step.md.erb
```diff
@@ -203,7 +203,7 @@ _Optional Attributes_
 
 <div class="Docs__note">
 <p class="Docs__note__heading">-1 Exit Status</p>
-<p>An exit status of -1 results when a job loses connection with the agent running it. Jobs that exit with -1 will be automatically retried, unless you have overridden the retry behavior in the <code>pipeline.yml</code> file.</p>
+<p>A job will show an exit status of -1 when we loose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
```
lox:
```suggestion
<p>A job will show an exit status of -1 when we lose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
```

pages/pipelines/command_step.md.erb
```diff
@@ -203,7 +203,7 @@ _Optional Attributes_
 
 <div class="Docs__note">
 <p class="Docs__note__heading">-1 Exit Status</p>
-<p>An exit status of -1 results when a job loses connection with the agent running it. Jobs that exit with -1 will be automatically retried, unless you have overridden the retry behavior in the <code>pipeline.yml</code> file.</p>
+<p>A job will show an exit status of -1 when we lose the connection with the agent and Buildkite has given up trying to reconnect. When using `automatic: true`, Buildkite will automatically retry the job 2 times before showing it as failed.</p>
```
toolmantim:
```suggestion
<p>A job will show an exit status of -1 when we lose the connection with the agent. When using `automatic: true`, Buildkite will automatically retry the job 2 more times before failing.</p>
```

keithpitt:
@harrietgrace happy for you to merge if you're happy with it!

pages/pipelines/command_step.md.erb
```diff
@@ -203,7 +203,7 @@ _Optional Attributes_
 
 <div class="Docs__note">
 <p class="Docs__note__heading">-1 Exit Status</p>
-<p>An exit status of -1 results when a job loses connection with the agent running it. Jobs that exit with -1 will be automatically retried, unless you have overridden the retry behavior in the <code>pipeline.yml</code> file.</p>
+<p>A job will show an exit status of -1 when we lose the connection with the agent. When using `automatic: true`, Buildkite will automatically retry the job 2 more times before failing.</p>
```
harrietgrace: A couple of little wording suggestions:

```suggestion
<p>A job will show an exit status of -1 if the connection to the agent is lost. When using `automatic: true`, Buildkite will try to reconnect to an agent twice before marking the job as failed.</p>
```

In the first sentence I removed the 'we' because it makes what you're referring to ambiguous (Is it Buildkite that's lost the connection to the agent? Me as a user? The job itself?). I also changed 'when' to 'if' so that it sounds like this isn't a regular occurrence.

In the second sentence, I wanted to relate it more to the exit -1 code, and it felt a bit like we were just describing the behaviour of the `automatic:true` attribute. It also sounded a little like it was going to be marked as failed whatever the outcome of the retry was 😅

I've had a stab at the second half of the second sentence - if that's not actually what is happening let me know 😄

PR #346, opened 2018-12-10
Clarify empty string is the same as false when skipping

benclive:
Clarifying the use of `""` within the skip command section. Wasn't sure whether to add this as a sentence or as an example case, but I decided it wasn't clear what the example case results in.


harrietgrace:
Great addition! Thanks for taking the time to PR @benclive 😄🙌🏻

PR #350, opened 2018-12-17
SSO guides fixes

harrietgrace:
A bunch of small fixes for the SSO guides. 

@toolmantim I'm particularly after some help with the Okta guide addition - an extra line in the explanation of what we do on our end once you give us your SSO provider info. 

I want to be clear that people have to test their setup, have a successful login, and then we'll enable it before logging everyone out. I don't know if it's clear enough that a sucessful login/authentication/authorization (not sure which one is applicable here), is required and that we aren't going to just enable it without making sure that people can get into their org. 

This section is in each of the guides, so this will be copied over to those too once we settle on the wording. 

Feedbacks on any of the additions is welcome though! I've been having a bit of a time with all of them ☹️


toolmantim:
I had a go at trying to make things clearer for what the process was, and how it won't break anyone's logins until support activates things after a test login. What do you reckon?

I also had a go at trying to make the number list a little clearer too (just locally in Safari, I haven't committed it yet).

Before:

image

After:

image

```css
span.external-ui-reference {
  border: 1px solid #ccc;
  padding: 1px 4px;
  border-radius: 3px;
}
.TextContent li, .BlogArticleBody__body li, .Docs__article li {
  margin-top: .25em;
  margin-bottom: .25em;
}
```

Looking at it again now, it could even be a green outline around the labels… because they're pretty important for what they have to go searching for. What do you reckon about something like that too?

pages/tutorials/okta.md.erb
```diff
@@ -1,23 +1,14 @@
...
-Self-service SSO provisioning in Buildkite is available via our GraphQL explorer. For further information, see the [GraphQL SSO Setup Guide](/docs/tutorials/graphql-sso). 
-
-### Setup in your Okta organization
-
-The instructions below guide you through adding the Buildkite application to your Okta account. The Buildkite team will enable SSO for your organization, with the details you provide from the end of the setup process.
+Okta can be used as an SSO provider for your Buildkite organization. To complete this tutorial, you will need admin privileges for both Okta and Buildkite.
 
 <div class="Docs__note">
-<p class="Docs__note__heading">Before starting your Okta setup</p>
-<p>Email the Buildkite team at support@buildkite.com to request your unique SAML service URL, you will need this when setting up your Okta application.</p>
```
toolmantim: I forgot to ask, is this still needed? I didn’t see it mentioned in the Okta setup instructions before? If it is, maybe we can add it back as a troubleshooting note after this, and mention it in the instructions?

harrietgrace: I specifically wanted to call it out at the start of the guide, not as part of the steps. Otherwise it's a bit like "step 1, email us and wait". At least when it's outside the instructions as a prerequisite people are able to continue with their setup a bit more speedily once they start? If that makes sense?

harrietgrace: And yes it's still needed 😄

toolmantim: That’s cool. I was just a bit confused, because if you follow the instructions to email for the URL, it then doesn’t appear in the steps below. Where do you copy and paste in the URL?

harrietgrace: I thought it was part of the instructions on the Okta side, but I just checked and it’s totally not! So nevermind about adding it back 👍🏻 

I also realised that we can just tell people about the metadata URL straight off the bat and not worry about getting them to follow the okta instructions at all - okta provides a link to their metadata on their sso setup page already 👌🏻

harrietgrace:
Updated the rest of the guides to the new layout, and have updated Okta to use the metadata rather than the old Okta instructions. Also removed the 'email BK for your special url' bit from onelogin, because you don't need it there either. 

@toolmantim let me know what you think!

PR #354, opened 2018-12-30
Update defining_steps.md.erb

ayersin-tyro:
The behaviour described here for dynamically generated steps is not exact. I hope my description is clear enough on the actual behaviour of the buildkite agent. If not I can provide an example


harrietgrace:
Thanks for the addition @ayersin-tyro, and for taking the time to PR! 😄🙏🏻

Good catch, your description was great. I've added another code example in to expand on your explanation and made a few little wording changes 👍🏻

PR #358, opened 2019-01-10
Remove unmaintained NodeJS library

conorgil:
The most recent commit to the buildkite-node library was [2 years ago](https://github.com/fraserxu/buildkite-node/commits/master) and the package is [not available in NPM](https://www.npmjs.com/search?q=buildkite-node).


harrietgrace:
Good catch @conorgil! Thanks for taking the time to PR too 😊👌🏻

PR #360, opened 2019-01-14
Add script for checking agent env is documented

lox:
Initially I opened this in https://github.com/buildkite/agent/pull/896, but I think this might be a better home for it. We could run a scheduled build that checks that all the current agent env are documented?

Another option for this would be to move the agent specific docs into the agent repo, or having a YAML source document in the agent repo that has the ENVs documented.

### DIFF
--- a/.buildkite/steps/check-agent-env.sh
+++ b/.buildkite/steps/check-agent-env.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


harrietgrace:
So exciting! Thanks for putting this together 🙏🏻

I think it would be good to keep all of the env vars listed in one place, even if they're related to different things.

lox:
Superseded by #376

PR #362, opened 2019-01-20
Update the URL to Terminal

toolmantim:
This updates the URL to Terminal (see: https://github.com/buildkite/terminal-to-html/issues/52)


toolmantim:
Will merge now to fix the links, so they're broken, but if you have any edits @harrietgrace we can do them in master or a new PR?

PR #363, opened 2019-01-21
Command step attribute cleanups

toolmantim:
While helping on https://github.com/buildkite/agent/pull/899 I realised that perhaps we need to update the documentation on automatic retries. And while I was there, I gave all the examples a bit of a once over to make sure we weren't adding attributes on types of jobs that typically didn't use them.


toolmantim:
Thanks @harrietgrace!

PR #364, opened 2019-01-22
Minor updates to the graphql sso guide

harrietgrace:
@eleanorakh mentioned a few things that could be clarified in our GraphQL SSO guide:

- We don't explain that the provider id is different to the organization id, we just sneakily start using it instead
- We don't explain the constraints we have on the emailDomainVerificationAddress param, and they're pretty strict!

This PR adds:
- Some very basic info about retaining your provider id in step 1 for later use (drawing attention to the different types of ids used in each step)
- The provider ID to the output of the saml provider create mutation (otherwise people don't get this value without running an additional org providers query)
- Info about the verification address, and how it has to match the domain in the previous param, and it must be one of four options for the start of the address

### DIFF
--- a/pages/tutorials/sso_setup_with_graphql.md.erb
+++ b/pages/tutorials/sso_setup_with_graphql.md.erb
@@ -45,6 +45,8 @@ mutation CreateProvider {
 }
 ```
 
+You'll need the provider's `id` from the output of this mutation in step three. 
+
 ### Step 2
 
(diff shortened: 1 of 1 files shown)


harrietgrace:
Neither! I noticed them when I was setting stuff up for the Okta testing account just a couple of weeks ago, and got an error message about it in the graphql explorer.

PR #371, opened 2019-01-25
Refactored agent installation docs

keithpitt:
We don't insert tokens magically in docs anymore, so just cleaning up this business


keithpitt:
After I removed `token`, I realised that `render_agent_setup` isn't needed anymore at all!

I moved the actual docs back into the main documents. I've done a look through and I think everything looks ok.

I'm gonna leave this one for a proper review from @harrietgrace though!

(note: this isn't anything important, so put it at the bottom of the importance queue :P)

keithpitt:
@harrietgrace it's a good q! They *used* to be shared, but frontend now has a copy/paste of them - so no worries about breaking paths: https://github.com/buildkite/frontend/tree/master/app/components/docs

PR #373, opened 2019-01-28
Update job event descriptions

harrietgrace:
We've been using the descriptions from the main bk app's config yml in here for a while. They don't really describe what's actually going on, and we've had a few people now ask about what the difference is between started and activated. 

I couldn't find if those snippets are used elsewhere in the app, so I decided to just change them in the docs 👍🏻

### DIFF
--- a/pages/apis/webhooks/job_events.md.erb
+++ b/pages/apis/webhooks/job_events.md.erb
@@ -8,15 +8,15 @@
 <tbody>
   <tr>
     <th><code>job.started</code></th>
-    <td><%= webhook_description "job.started" %></td>
+    <td>A command step has been started</td>
   </tr>
   <tr>
     <th><code>job.finished</code></th>
-    <td><%= webhook_description "job.finished" %></td>
+    <td>A job has been finished</td>
   </tr>
   <tr>
     <th><code>job.activated</code></th>
-    <td><%= webhook_description "job.activated" %></td>
+    <td>A block step has been unblocked</td>
   </tr>
 </tbody>
 </table>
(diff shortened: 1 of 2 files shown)


pages/apis/webhooks/job_events.md.erb
```diff
@@ -8,15 +8,15 @@
 <tbody>
   <tr>
     <th><code>job.started</code></th>
-    <td><%= webhook_description "job.started" %></td>
+    <td>A command step has been started</td>
```
toolmantim: What about "A command step job has been started"? You can't start a step, per-se. But jobs have "steps" which are like templates. Trick!

toolmantim: Also, perhaps its "A command step job has started running on an agent"? I just checked the code, and that seems to be the case.

pages/apis/webhooks/job_events.md.erb
```diff
@@ -8,15 +8,15 @@
...
   </tr>
   <tr>
     <th><code>job.finished</code></th>
-    <td><%= webhook_description "job.finished" %></td>
+    <td>A job has been finished</td>
   </tr>
   <tr>
     <th><code>job.activated</code></th>
-    <td><%= webhook_description "job.activated" %></td>
+    <td>A block step has been unblocked</td>
```
toolmantim: "A block step job has been unblocked" perhaps?

pages/apis/webhooks/job_events.md.erb
harrietgrace: Hah I had 'job' in my first draft, then took it out 😄 Happy to add it back in!

toolmantim: Hah, cool! Yeah, I reckon let’s add it.

harrietgrace:
Reverted back to using the variables, and started using the variables in the main Webhooks overview document too. Actual changes to the descriptions are now in https://github.com/buildkite/buildkite/pull/3473.

PR #375, opened 2019-01-29
Fix broken link

JohnB:
The link had a space in it, which broke the actual linkage. Also, broke the line to be easier to edit.


harrietgrace:
Good catch @JohnB! Thanks for taking the time to PR 😄🙏🏻

PR #378, opened 2019-01-30
Update agent help for 3.8.2

lox:
This uses the new `scripts/update-agent-help.sh` to update all the agent help output for v3.8.2.


keithpitt:
This is cool!

lox:
I have grand plans of one day automating this more, but it's ok for now.

PR #384, opened 2019-01-31
Artifact delete REST API documentation

eleanorakh:


toolmantim:
Sorry @harrietgrace — we're going to merge this now so we can changelog it, but if you'd like to fix it up please do, or just comment any problems there. This was a bit of a tricky one.

pages/apis/rest_api/artifacts.md.erb
```diff
@@ -150,3 +150,17 @@ curl "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pipeline.
 Required scope: `read_artifacts`
 
 Success response: `302 Found`
+
+## Delete an artifact
+
+Marks the artifact as deleted, removes it from the job and builds artifact list, and removes the file from Buildkite’s AWS S3 artifacts bucket.
```
harrietgrace: Where is the artifact marked as deleted? If it's removed from the artifact lists, is the 'marked as deleted' something that the user sees, or is it just internal?

pages/apis/rest_api/artifacts.md.erb
```diff
@@ -150,3 +150,17 @@ curl "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pipeline.
...
+
+If the artifact was stored in your own [AWS S3](https://buildkite.com/docs/agent/v3/cli-artifact#using-your-own-private-aws-s3-bucket) or [Google Cloud](https://buildkite.com/docs/agent/v3/cli-artifact#using-your-own-private-google-cloud-bucket) storage, it will not be automatically deleted from your store.
```
harrietgrace: What happens in this case? It's still removed from the lists, but not actually deleted from their S3?

harrietgrace:
Nice one @eleanorakh! 🎉 Thanks so much for adding this 🙏🏻 I've added a couple of little questions I had while I was reading it 😊

pages/apis/rest_api/artifacts.md.erb
toolmantim: It’s mostly internal, for the moment. We didn’t want to say it’s “deleted” because it’s ambigious as to whether it’s the DB meta-data and the file, and we don’t delete the meta-data (filename) but mark it as deleted which makes it disappear from the UI and artifacts APIs.

We might add UI or APIs to show/filter by deleted artifacts tho, so we figured “mark as deleted” might be an okay concept to introduce. And especially with data deletion it’s best not to say it was “deleted from the database” if that’s not the case.

Hope that helps with the background! Still not 100% happy with the words, so all suggestions are mega welcome 😊

pages/apis/rest_api/artifacts.md.erb
toolmantim: Yep, that’s it.

pages/apis/rest_api/artifacts.md.erb
harrietgrace: Awesome, thanks! What do you think about saying something like 'Internally marks the artifact as deleted'? I'm not entirely happy with those words, but I feel like we need to make it clearer what 'marks as deleted' means. 

Would there be objections to further transparency, something like how you described it above: "The filename is marked as deleted in the Buildkite database, and the artifact itself is removed from the Buildkite AWS S3 bucket. It will no longer be displayed in the job or build artifact lists, and it will not be returned by the artifact APIs."

toolmantim: Sounds good! Perhaps “the artifact record” instead of “the filename” though?

toolmantim: It’s be great to review the UI copy once we’ve updated these docs too. The UI needs this PR’s thinking applied back to it.

harrietgrace: Updated in https://github.com/buildkite/docs/pull/393 😊 Happy to look at the UI copy too, sounds rad!

PR #389, opened 2019-02-04
Fix incorrect BUILDKITE_BUILD_ID docs

lox:
The docs currently suggest that `BUILDKITE_BUILD_ID` can be blank on rebuilds, surely not?


keithpitt:
Probs a typo from REBUILD or something. Yeah, BUILD_ID is always a thing!

Issue #390, opened 2019-02-05
Mis-numbered steps in SSO GSuite docs


eleanorakh:
I noticed the steps go from steps 1-3 [here](https://buildkite.com/docs/integrations/sso/google-oauth).

harrietgrace:
😅Thanks @eleanorakh! Will fix 🏃🏻‍♀️

PR #392, opened 2019-02-05
Fix google sso step numbers

harrietgrace:
Closes https://github.com/buildkite/docs/issues/390 - thanks for pointing it out @eleanorakh! 

While checking the other docs for the same numbering issue, I realised that the custom SAML doc never made it over to the new steps-based structure. This updates to use the new, standard words and formatting.


eleanorakh:
Awesome! :) 👏👏👏👏

Issue #412, opened 2019-02-19
Hyperlinks in job output undocumented


ticky:
We [support hyperlinks in job output](https://github.com/buildkite/terminal-to-html/blob/bc7274cbce9fb5c752c32fdb4ea267cb4260417f/link.sh) using our underlying `terminal-to-html` renderer, however, this code is not documented and, presumably, left as an exercise for the reader to know we’re imitating iTerm 2 and co!

It’d be good to add this, perhaps alongside the existing [images in job output](https://buildkite.com/docs/pipelines/images-in-log-output) doc.

harrietgrace:
Rad, thanks for making an issue about it @ticky! 🙌🏻I'll leave this open until it's added.

ticky:
@harrietgrace I’d be happy to draft this up, do you think it should be its own page, or should the existing “Images in job output” be renamed, maybe to “Links and images in job output” (w/ redirect)?

toolmantim:
I reckon the latter?

ticky:
I have realised that our codes aren’t the same as iTerm’s - ours are in fact completely incompatible with iTerm’s! Opened this issue around that: buildkite/terminal-to-html#58

PR #419, opened 2019-02-22
Document links in log output

ticky:
Proposed change for #412


pages/pipelines/links_and_images_in_log_output.md.erb
```diff
@@ -1,12 +1,40 @@
...
+
+  printf '\033]1339;'$LINK'\a\n'
+}
+```
+
+You can use it like so:
+
+```bash
+inline_link "https://buildkite.com/"
+```
```
sj26: Maybe also worth mentioning that you can link artifacts using "artifact://..." urls?

sj26:
This looks good!

pages/pipelines/links_and_images_in_log_output.md.erb
ticky: Done!

sj26: Nice :+1:

ticky:
Uh-oh, this broke the images in this document…

ticky:
Looks like it was actually a deploy bug!

sj26:
The deploy bug should be fixed now 👌

PR #426, opened 2019-03-05
Add some guidance about docker permissions problems

lox:
Often folks have issues where files are left as root. I often link to https://github.com/buildkite/agent/issues/407#issuecomment-370573431 for some help, but would be great to have that help here. 

It's a really tricky thing to explain, but it bites lots of folks. Thoughts @toolmantim / @harrietgrace?

### DIFF
--- a/pages/agent/v3/docker.md.erb
+++ b/pages/agent/v3/docker.md.erb
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
 COPY hooks /buildkite/hooks/
 ```
 
+## Permissions Errors when using Docker
+
+A common problem encountered when using Docker for builds along with volume mounts from the host is the mismatch of permissions that can occur.
+
+If a build in Docker writes files to the volume that is bind mounted to the host system, the user and group that is set for the files on the host system can end up as `root:root` (the user that the docker daemon is running as). If the rest of your builds are running as `buildkite-agent` then you can run into issues where subsequent checkouts fail with errors such as `Warning: checkout failed to remove ...`.
+
+To demonstrate this issue locally you can run the following:
+
+```bash
+docker run -v "$PWD:/pwd" --rm -it alpine:latest mkdir /pwd/llamas
+ls -al llamas
+```
+
+You should see that the llamas directory is owned by either root, or whatever user docker is run by.
+
+There are a few solutions to this.
+
+* [User namespace remapping](https://docs.docker.com/engine/security/userns-remap/) to the same user that your buildkite-agent is running as.
+
+* Run a script before or after builds that resets permissions. You can do this either via docker (because it runs as root) or via sudo. See https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions and https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/sudoers.conf for examples.
+
+
 ## Allowing builds to use Docker
 
 To use Docker and volume mounting from build scripts, you need to ensure that the builds directory, and the `buildkite-agent` binary path, are mounted in from the host machine, and that their paths on the host and in the agent container are the same.
--- a/pages/agent/v3/docker.md.erb
+++ b/pages/agent/v3/docker.md.erb
@@ -77,18 +77,14 @@ COPY hooks /buildkite/hooks/
 
 ## Permissions errors when using Docker
 
-A common problem encountered when using Docker for builds along with volume mounts from the host is the mismatch of permissions that can occur.
+A problem you may encounter when using Docker volume mounts (-v), is that the container may create files on the host system with root user permissions, which can result in errors like the following:
 
-If a build in Docker writes files to the volume that is bind mounted to the host system, the user and group that is set for the files on the host system can end up as `root:root` (the user that the docker daemon is running as). If the rest of your builds are running as `buildkite-agent` then you can run into issues where subsequent checkouts fail with errors such as `Warning: checkout failed to remove ...`.
-
-To demonstrate this issue locally you can run the following:
-
-```bash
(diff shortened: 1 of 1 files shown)


pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
 COPY hooks /buildkite/hooks/
 ```
 
+## Permissions Errors when using Docker
+
+A common problem encountered when using Docker for builds along with volume mounts from the host is the mismatch of permissions that can occur.
```
toolmantim: What about?

> A problem you may encounter when using Docker volume mounts (`-v`), is that the container may create files on the host system with invalid user permissions, which can result in errors like the following:
>
>```bash
>$ git clean -fxdq
>warning: failed to remove dist/
>```

I idea is to try to get that error message searchable, because our agent does a `git clean` and that's usually when this problem crops up.

pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
...
+
+If a build in Docker writes files to the volume that is bind mounted to the host system, the user and group that is set for the files on the host system can end up as `root:root` (the user that the docker daemon is running as). If the rest of your builds are running as `buildkite-agent` then you can run into issues where subsequent checkouts fail with errors such as `Warning: checkout failed to remove ...`.
```
toolmantim: We can probably remove this with the above suggested edit.

pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
...
+To demonstrate this issue locally you can run the following:
+
+```bash
+docker run -v "$PWD:/pwd" --rm -it alpine:latest mkdir /pwd/llamas
+ls -al llamas
+```
+
+You should see that the llamas directory is owned by either root, or whatever user docker is run by.
+
+There are a few solutions to this.
```
toolmantim:
```suggestion
To fix the permissions of files created, you can:

* Change the way permissions are set on the files created by your Docker container, but modifying your container’s `USER` or modifying your build commands.
```

pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
...
+
+* [User namespace remapping](https://docs.docker.com/engine/security/userns-remap/) to the same user that your buildkite-agent is running as.
```
toolmantim:
```suggestion
* Configure [user namespace remapping](https://docs.docker.com/engine/security/userns-remap/) on your Docker host, to ensure that container users are remapped to the user your buildkite-agent is running as.
```

pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
...
+
+* Run a script before or after builds that resets permissions. You can do this either via docker (because it runs as root) or via sudo. See https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions and https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/sudoers.conf for examples.
```
toolmantim:
```suggestion
* Run a script before or after builds that resets permissions. You can do this either via Docker (because it runs as root) or via sudo. See the Buildkite Elastic Stack for AWS’s [fix-buildkite-agent-builds-permissions](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions) and [sudoers.conf](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.4/packer/conf/buildkite-agent/sudoers.conf) scripts for an example of using an agent hook and sudo command.
```

pages/agent/v3/docker.md.erb
```diff
@@ -75,6 +75,28 @@ FROM buildkite/agent:3
...
+## Permissions Errors when using Docker
+
+A common problem encountered when using Docker for builds along with volume mounts from the host is the mismatch of permissions that can occur.
+
+If a build in Docker writes files to the volume that is bind mounted to the host system, the user and group that is set for the files on the host system can end up as `root:root` (the user that the docker daemon is running as). If the rest of your builds are running as `buildkite-agent` then you can run into issues where subsequent checkouts fail with errors such as `Warning: checkout failed to remove ...`.
+
+To demonstrate this issue locally you can run the following:
+
+```bash
+docker run -v "$PWD:/pwd" --rm -it alpine:latest mkdir /pwd/llamas
```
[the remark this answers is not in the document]
lox: That's because of Docker for Mac magic, but it will replicate on a linux system. We could exclude this example entirely?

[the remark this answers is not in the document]
toolmantim: I see! https://docs.docker.com/docker-for-mac/osxfs/

Given many of our developers are Mac users, maybe we just remove it? Also maybe we should mention this is only a problem for Linux and Windows Docker hosts?

I found this for Windows: https://docs.docker.com/docker-for-windows/troubleshoot/#volumes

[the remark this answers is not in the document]
lox: Yeah, think you are right on both counts

lox:
Changes made @toolmantim, lemme know what you think or if I missed any.

harrietgrace:
Great addition @lox! It's definitely a complex situation to explain, but I think this has turned out really nicely. I've made some little updates, but nothing changing the content:
- Added in a mention of Windows/Linux, looks like this came up in the comments above but didn't make it into the doc (is it still a thing? I can take it out if not)
- Removed some extra words, I had a bit of a hard time parsing some of the more complex sentences, hopefully they're a little cleaner now
- Fixed a few little grammar and formatting things

lox:
Thanks folks!

PR #428, opened 2019-03-07
Integrated Rails to make docs a standalone application

keithpitt:
So I've wrapped the docs in a Rails app - and it works (mostly). I've pulled all the rendering bits and pieces from the main app and imported them here.

**Y u do dis?**

Having docs be a thing that runs by itself gives us a few new super powers:

- It can be run independent of the main application. If there's planned maintenance or a downtime event, Google won't penalise us for having 500s on docs pages
- If there *is* an outage, at least you'll be able to get to the docs!
- Contributing to the docs is easier since you can run the whole thing locally

If we don't want to host it ourselves, maybe we can think about hosting it somewhere else (i.e. we own `buildkite.help` for example, maybe we could host the site there on Heroku or something to keep it simple?)

**What about Middleman and making it a static page?**

I spent a few hours playing with Middleman trying to make the docs static, but alas, it was a dead end. There's a few reasons this is a bad idea:

- We loose powers of dynamic pages
- We can't do canonical redirects (i.e. "../github_enterprise" -> "../github-enterprise"
- When we make big changes to docs, it's really nice having redirect super powers, we'd loose this
- If it was static, we'd need to rely on 3rd party JS for interesting things (like search)

**Cool, so is it done?**

Not yet, there's a few TODOs laying around that need to be dealt with. I've only done raw copy/pasting of things, and some bits were a bit too hard to copy at the time.

Also the CSS could do with a cleanup since I think there's some stuff not specific to docs in there.

**Can we merge it now?**

Actually, yes!

I've not changed any of the actual documentation pages, and I've symlinked them from their current location into the right spots in the Rails app. That means we can safely merge this without affecting the current docs implementation on Buildkite. Then we can bit by bit working on improving the app here, and when we're ready, do a cut over or something.

But before we can:

- [ ] Rotate the keys thing before pushing to production and figure out what to do with the session key


config/credentials.yml.enc
```diff
@@ -0,0 +1 @@
+ISDN1jKTGDsf91DLMWjozmKAZAhDzV9OOPsI+rZgPdg5Nt6aVBJDA7jLlGsfCYqticLVp+/4brgsmdgpPhBJS/UrErPTDNY9b05HNUIBhfcAQ86QpncjRvRzlZu8VGaeuwWDDQ/p4pccHcswI+1v2oJG5C2iCZclUKhPW0z+EHiqv+1TmgT+vgZpbGxWQoODtSgeSM7xbf/p5N8C8WH8z5l2A4WE4hdfEqN0vr32vH8uCzl2M4B1hgGc/QFKSIuT2qi4BW2wRWw7vbIaBoJgJxz04slN2TKDaE52hx/s37Bafpet1kW0C1aW7fZN2BZlSjcHSkF6kHMMTYkTlWnCN1v4dHdTb2t7StzP89aAXcpNLXN/E3ISxo+k4ZOCXQeEtkC2lq7LuB8ZPAlq1pMssHWQlk+92Pt7AAKs--1MiOy4keiVeyswCw--jru7cR2e0KN8bJqGXKOwag==
```
keithpitt: Need to remove this file that was generated by Rails, and do something with it.

@matthewd is it needed for the app to run? Can I just `rm` it?

keithpitt:
Things I could use a hand on/someone else do :)

- [ ] Wrap it in a Docker thing and get it running `:rspec:` on CI (specifically with `soft_fail: true`) while we get the rest of the specs fixed in master

toolmantim:
Woohoo! Such excite 👏🏼

toolmantim:
I added a Dockerfile, and a docker-compose.yml setup too, along with a pipeline.yml to run the specs.

Unforuntately the prod build of the Docker image isn't serving the correct CSS and JS assets, so things are busted. But it works for dev and specs at the moment.

config/credentials.yml.enc
keithpitt: Turns out this file isn't as bad as I thought:

https://www.engineyard.com/blog/rails-encrypted-credentials-on-rails-5.2

I rekon we need to `rm` it and have a step in the Readme that's like:

"You'll need to generate a session token" or something.

I thought about disabling sessions, but we need them for CSRF which we may use for search forms later on.

toolmantim:
@keithpitt I noticed emoji rendering seems broken. I wonder how we can get emoji support in here 🤔

image

config/credentials.yml.enc
toolmantim: We shouldn't need CSRF for search forms though? Should be able to just disable?

keithpitt: If the we have a search controller on our end we may want to do POST's instead of GET to stop it from being smashed by some bot or what ever.

Also disabling sessions/cookies in Rails seems like a lot of effort - I don' think it's something worth doing - and if we ever want to do anything session-ish later on, we'll have that super power.

keithpitt:
@toolmantim my current plan with Emojis is to merge without them, open source our Emoji renderer into the emojis repo, turn it into a Gem, and include it in both as a Rails engine.

config/credentials.yml.enc
toolmantim: 🤷🏼‍♂️👍🏼

toolmantim:
I just did a 5 minute test with 0d20e34 and got emojis working @keithpitt, because I think it's the last thing that needs doing. Everything else is, I think, mostly ready to go?

harrietgrace:
So how do docs deploys happen now? Do I just do the usual docs bump on bk/bk?

toolmantim:
Yep, still gotta do the usual bump @harrietgrace

toolmantim:
Deploys haven’t changed just yet. Once this is running fully standalone, merging to master will have it automatically deploy 🎉

Automobileengineeringservices:
#428

Issue #141, opened 2017-07-20
Document how to run multiple agents with different config files


toolmantim:
In the docs we show [how to run multiple agents](https://buildkite.com/docs/agent/ubuntu#running-multiple-agents) but we don't show any way to customise the agent configuration files for each agent… say, for giving them different tags/meta-data.

Someone just submitted the process it took for them:

```markdown
To run multiple agents, on the same host, with different configurations, and behind a proxy server.

Once you have installed and tested an agent as per the installation instructions.

## Kill the single agent.

Stop the service and disable it "sudo systemctl stop buildkite-agent" and "sudo systemctl disable buildkite-agent"

## Config Files.

In the buildkite-agent home directory, `/ect/buildkite-agent` make a copy of buildkite-agent.config for each different config you want, e.g. `buildkite-agent-1.cfg`, `buildkite-agent-2.cfg`, etc. The name is irrelevant so call it what you like.

Then edit the new config files and set whatever configuration you want - I just set different queue names in the meta-data/tags.

## System launch file.

This file is `/lib/system/system/buildkite-agent.service`

Make a copy for each agent you want to run, i.e. `buildkite-agent@1.service`, `buildkite-agent@1.service`, etc., and edit this file.

There is an ExecStart line that launches the service, add to the end of this, in each file, ` -config /etc/buildkite-agent/buildkite-agent-1.cfg` or whichever conf you want to use.

Now you can enable the service, `sudo systemctl enable buildkite-agent@1`, and it should run.

## Repeat for however many agent you have setup.

This creates a symlink called `/etc/system/system/mutli-user.target.wants/buildkite-agent@1.service` that should link to the system launch file `/lib/sytemd/system/buildkite-agent@1.service` that was created earlier.

## Start the agents

Finally start the agents: `sudo systemctl start buildkite-agent@1`
```

harrietgrace:
@toolmantim is this still the only way to have multiple config files? Does the config path env var/flag not allow you to specify different files?

toolmantim:
[...]

PR #429, opened 2019-03-08
Add new 'running agents' page

harrietgrace:
An outline of an idea for a new page under the Agent/v3 menu item, to possibly go after the Configuration page. 

It will pull together a bunch of info that's presently scattered over several docs - starting, configuring, running multiple, doing scaling etc. It won't be a deep dive on anything that we have docs on already like config, it will have an overview and then link out to those existing sections. It'll also include the info from #141.

Thoughts @toolmantim? Anything else you'd like to see included?

### DIFF
--- a/pages/agent/v3/running.md.erb
+++ b/pages/agent/v3/running.md.erb
@@ -1,0 +1,25 @@
+# Running Buildkite Agent
+
+- intro
+
+<%= toc %>
+
+## where to run
+	your machine or other physical hardware, cloud services
+	docker or other containers
+
+## running one
+	need an agent token, link to token docs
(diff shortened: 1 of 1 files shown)


toolmantim:
Oh, this is a really interested idea!

I'm not sure if I can see it as a standalone page. Perhaps it should be a series of new sections on the agent overview page (https://buildkite.com/docs/agent/v3) and we enable the TOC at the top?

harrietgrace:
Add note about --spawn command for multiple agents

harrietgrace:
Closing this for now, keeping the content elsewhere for a future PR 👍🏻

PR #431, opened 2019-03-11
Add ID attribute

harrietgrace:
From https://github.com/buildkite/buildkite/pull/3511, this adds the ID attribute to the command step table. I'm not 100% on its function, @toolmantim any suggestions for a better example?

### DIFF
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -92,7 +92,7 @@ _Optional attributes:_
   <tr>
     <td><code>id</code></td>
     <td>
-      A string to identify the step. The value is available in the `BUILDKITE_STEP_IDENTIFIER` <a href="/docs/pipelines/environment-variables">environment variable</a><br>
+      A string to identify the step. The value is available in the <code>BUILDKITE_STEP_IDENTIFIER</code> <a href="/docs/pipelines/environment-variables">environment variable</a>.<br>
       <em>Example:</em> <code>"linter"</code>
       <em>Alias:</em> <code>identifier</code>
     </td>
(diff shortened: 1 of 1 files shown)


pages/pipelines/command_step.md.erb
```diff
@@ -89,6 +89,14 @@ _Optional attributes:_
       <em>Example:</em> <code>RAILS_ENV: "test"</code>
     </td>
   </tr>
+  <tr>
+    <td><code>id</code></td>
+    <td>
+      A string to identify the step. It is exported as the `BUILDKITE_STEP_IDENTIFIER` environment variable.<br>
+      <em>Example:</em> <code>"linter"</code>
+      <em>Alias:</em> <code>Identifier</code>
```
toolmantim: I think this should be lowercase:

```suggestion
      <em>Alias:</em> <code>identifier</code>
```

pages/pipelines/command_step.md.erb
```diff
@@ -89,6 +89,14 @@ _Optional attributes:_
       <em>Example:</em> <code>RAILS_ENV: "test"</code>
     </td>
   </tr>
+  <tr>
+    <td><code>id</code></td>
+    <td>
+      A string to identify the step. It is exported as the `BUILDKITE_STEP_IDENTIFIER` environment variable.<br>
```
toolmantim: Awesome! Do we use “export” anywhere in here? It means something specific to Linux environment variables, but perhaps that’s not what’s happening here.

What about something like…

```suggestion
      A string to identify the step. The value is available in the `BUILDKITE_STEP_IDENTIFIER` <a href="/docs/pipelines/environment-variables">environment variable</a><br>
```

PR #436, opened 2019-03-20
Add include_retried_jobs API param

### LINKED ISSUE #425: Document include_retried_jobs
The REST API builds endpoints have an optional `include_retried_jobs=true` parameter which can include retried jobs which are excluded by default. We should add it to the docs.

harrietgrace:
Adds the `inlcude_retried_jobs` param to the optional query string table in the Builds REST API doc. 

image

@sj26 is the description accurate?

Closes #425


sj26:
Oh, hmm, may need a slight wording change. We don't include the original job which has been retried by default, we include the latest retry of the job.

Maybe something like:

> Returns any jobs that were retried alongside the final attempt of the job.

... or something. Words are hard :-/

toolmantim:
Or maybe:

> Include all retried jobs in each build’s jobs list.

harrietgrace:
Updated to use @toolmantim's words. Thanks all! 🙇🏻‍♀️

PR #446, opened 2019-04-01
Document `continue_on_failure` `wait` after normal `wait`

huonw:
This also includes a sequence of `continue_on_failure`s.


huonw:
Hi @harrietgrace, we'd encountered a case where we needed to know the details of having a `wait` step before a `wait: ~`/`continue_on_failure: true` step, and so we experimented to determine it. I thought it would be worth writing down the results of that experiment for future reference, even though it's a bit complicated. What do you think?

(Apologies if you're not the right person to ping. 🙂)

harrietgrace:
Thanks @huonw! I am definitely the person to ping 😄Before I merge I'm just going to double check with the team that the multiple `continue_on_fail` wait steps running one after the other is the intended behaviour. It's not a use case I've run into before - I'd be keen to hear what you're using it for in your pipeline! 🤓

huonw:
We have a build that does something like (in sequence)

1. build some temporary docker images in parallel
2. wait for those builds to all finish
3. trigger another pipeline to test them
4. publish the temporary images as "permanent" ones (if testing passed)
5. clean-up the temporary images

In graph form, we want the steps to run something like:

image


That is, we want:
- everything to be sequenced, so `wait`s between 3, 4 and 5
- 5 to always run, so those `wait`s (along with 2) need to be `continue_on_failure: true`
- to not run 3 if building fails, and not run 4 if testing fails, which we do by checking the buildkite API, and have conditional logic within the steps (basically, `if <build is not failing>; then <run the real step>; fi`, where `<run the real step>` is a dynamic upload of a pipeline containing a trigger step for 3.)

(In some ways, this particular use of this is a work-around for not having https://forum.buildkite.community/t/declare-dependencies-between-jobs/116.)

harrietgrace:
Ah awesome! Thanks for the extra context. The job dependencies is a pretty hot topic internally right now, so hopefully we'll have some options for that soon! 

A little rundown of what I've added in: I had some trouble wrapping my head around the big example code block, so I've split it up into two smaller ones. There's one that covers the 'regular wait steps block everything' case, and one that covers the 'multiple continue-on-failure steps all run' case. I found seeing it in a pipeline helped me sort it out in my head, so I've thrown in a couple of pipeline screenshots too. I got a bit carried away and ended up doing some edits to the rest of the 'Continuing on failure' section too 😄

Thanks so much for this PR, it's a great addition! 👏🏻🙏🏻

huonw:
> The job dependencies is a pretty hot topic internally right now, so hopefully we'll have some options for that soon!

Great! I'm excited for any improvements here, as we've got several places that do the hit-the-API work-around I described above.

> A little rundown of what I've added in: I had some trouble wrapping my head around the big example code block, so I've split it up into two smaller ones

Much better! 😄 

> Thanks so much for this PR, it's a great addition! 👏🏻🙏🏻

Thanks for merging it so quickly. 🎉

PR #453, opened 2019-04-10
Update agent man pages for v3.10.4

lox:
I realized that we didn't have docs for the ` --disconnect-after-idle-timeout` feature we discuss in the new Elastic Stack release.


lox:
Not sure why some of the ordering has changed!

PR #462, opened 2019-05-06
Add SIGQUIT behavior to docs

marktai:
We found out about the `SIGQUIT` behavior from source code, which we are using to immediately fail and retry jobs when we spin down AWS instances. 

Feel free to make any changes to the wording.

### DIFF
--- a/pages/agent/v3.md.erb
+++ b/pages/agent/v3.md.erb
@@ -54,3 +54,5 @@ The agent’s behavior can be customized using hooks, which are just shell scrip
 When a build job is cancelled the agent will send the build job process a `SIGTERM` signal to allow it to gracefully exit. If the process does not exit within 10 seconds it will be forcefully terminated with a `SIGKILL` signal.
 
 The agent will also accept `SIGTERM` directly, and will gracefully exit. After it has processed the current job, it will disconnect and stop accepting jobs. 
+
+Lastly, the agent accepts `SIGQUIT` directly, and will immediately terminate. The agent will immediately report its loss to Buildkite and will fail jobs. 


harrietgrace:
Looks great! Thanks for taking the time to PR @marktai, much appreciated 😄🙏🏻

pages/agent/v3.md.erb
```diff
@@ -54,3 +54,5 @@ The agent’s behavior can be customized using hooks, which are just shell scrip
 When a build job is cancelled the agent will send the build job process a `SIGTERM` signal to allow it to gracefully exit. If the process does not exit within 10 seconds it will be forcefully terminated with a `SIGKILL` signal.
 
 The agent will also accept `SIGTERM` directly, and will gracefully exit. After it has processed the current job, it will disconnect and stop accepting jobs. 
+
+Lastly, the agent accepts `SIGQUIT` directly, and will immediately terminate. The agent will immediately report its loss to Buildkite and will fail jobs.
```
toolmantim: Perhaps we want to go with something different from `report its loss` @harrietgrace? Also it can't fail multiple jobs, just a single job.

What if we went with a list of the signals the agent accepts? Something like this, but not sure of the formatting:

```suggestion
The agent also accepts the following two signals directly:

* `SIGTERM` - Instructs the agent to gracefully disconnect after completing any job that it may be running.
* `SIGQUIT` - Instructs the agent to cancel any job that it may be running and then immediately disconnect.
```

harrietgrace: Ah right, I assumed it would cancel all subsequent jobs that had run. But I like the new formatting and re-worked words! I'll PR those separately.

PR #463, opened 2019-05-08
Fix MY_ENV typo

jvke:
In the sentence edited, var name changes to `MY_ENV`, rather than `MY_ENV1` used above.


harrietgrace:
Good spot @jvke 😊 Thanks for taking the time to PR! 🙏🏻

PR #467, opened 2019-05-09
Update formatting and words for signal handling section

harrietgrace:
An update to the addition in #462. Added in @toolmantim's formatting and words suggestion.


toolmantim:
Thanks! 🙌🏼

PR #470, opened 2019-05-13
Add new configuration option to agent GCS docs 

DoomGerbil:
This adds a doc to cover the change in https://github.com/buildkite/agent/pull/1000.  It probably should only be merged _after_ that one.

FYI @lox @petemounce


harrietgrace:
Great addition @DoomGerbil! Thanks for the PR 🙏🏻 I restructured it a little to split up the code block. Given that the agent PR is merged, I'll merge this one too 🎉

PR #486, opened 2019-06-07
Correct the order of `<pre />` and `<code />` tags in annotation docs

ticky:
Turns out our docs had been misleading people somewhat the whole time!

When writing an HTML block in markdown, the wrapping element needs to be a recognised starting html element. `<pre />` qualifies, but `<code />` does not. This means that if someone used our example, the code would come through as processed markdown, causing strange rendering errors.

Fixes buildkite/terminal-to-html#64


ticky:
Thanks for the review! 😃

PR #487, opened 2019-06-10
Add the `multiple` attribute to the block step docs

ticky:
Not sure what else to put in this, but this seems like a start, and I don’t want to forget! 😅

Docs for buildkite/buildkite#3864


ticky:
Have updated this @harrietgrace! The behaviour changed a bit.

PR #488, opened 2019-06-11
Token API docs

keithpitt:
I added a `/token` API to GET and DELETE the current token being used. I put together a page to document it.

I called it "Token Management", but perhaps "Management" is too much considering you can only do 1 thing? I thought "maybe in the future there'd be a "rotate" command or something, and that'd fit in here.

image

Changes welcome!


toolmantim:
Woohoo!

I think just "Token API" for the page title, "Token" for the sidebar nav, and `/docs/apis/rest-api/token` for the URL? We need to update the app layout in the docs repo these days too (can't wait to remove this duplication!).

We should probably be clear that they're "API Access Tokens" in the words too (perhaps linking to https://buildkite.com/docs/apis/rest-api#authentication)… so as not to confuse them with other tokens (like agent tokens)?

keithpitt:
@toolmantim cool cool - changes made!

keithpitt:
Here's the URL: https://buildkite-docs-review-pr-488.herokuapp.com/docs/apis/rest-api/token

toolmantim:
@harrietgrace I'm curious if you think "Token API" is obvious enough, or if not, perhaps "Access Token API" might be better? (i.e. "Access Token" in the nav, and "Access Token API" in the header? And perhaps move it to the top of the nav?)

🤷🏼‍♂️

harrietgrace:
I think Token is alright by itself (rather than Access Token), but I think it needs to be 'Tokens' given that all the others are pluralized 😄 Otherwise looks great! We'll need to update this in bk/bk too given that the sidebar isn't actually coming from this repo.

keithpitt:
@harrietgrace I left it as "Token" since there's only 1 token you can ever get information for (the current token). There's no "token list" - so never any plural.

Do you think it's OK with that in mind? Or is the pluralisation because the page is talking about "tokens" generally?

harrietgrace:
I was just chatting about this with @toolmantim actually 😄I think it should be plural because as a user you can have multiple tokens. Also, because I would expect that a page called 'Token' or 'Access Token' would tell me about generating a token for the REST API and how to use it, I wouldn't expect an API page. 'Access Tokens' I would think was about managing my tokens/doing things with tokens, and even though the API only lets you deal with one at a time, I think it still makes sense to have it plural in the sidebar.

pages/apis/rest_api/token.md.erb
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+Required scope: none
+
+Success response: `200 OK`
+
+## Revoke current API Access Token
+
+Immediately revokes the current [API Access Token](/docs/apis/rest-api#authentication) and sends an email to the owner. Once revoked, the token cannot be used anymore.
```
toolmantim:
```suggestion
Revokes the API Access Token that was used to make the API request. Once revoked, the API Access Token can no longer be used.
```

pages/apis/rest_api/token.md.erb
```diff
@@ -0,0 +1,34 @@
+# Token API
```
toolmantim:
```suggestion
# Access Token API
```

pages/apis/rest_api/token.md.erb
```diff
@@ -0,0 +1,34 @@
...
+
+<%= toc %>
```
toolmantim:
```suggestion
The Access Token API lets you perform actions on an individual [API Access Token](https://buildkite.com/docs/apis/rest-api#authentication), using the authentication details of REST API request.

<%= toc %>
```

pages/apis/rest_api/token.md.erb
```diff
@@ -0,0 +1,34 @@
...
+
+## Get current API Access Token
+
+Returns limited information about the current [API Access Token](/docs/apis/rest-api#authentication) being used to access the REST API.
```
toolmantim:
```suggestion
Returns details about the API Access Token that was used to make the API request.
```

keithpitt:
@toolmantim happy for you to merge what ever changes you want in!

toolmantim:
I'll make those docs changes!

Did you wanna make the REST endpoint `/v2/access-token` @keithpitt?

toolmantim:
Based on Keith's demo words, we should probably add an intro for some of the common use cases of this API, such as finding an API Access Token and not knowing what user it's attached to, and wanting to make sure it's revoked.

PR #469, opened 2019-05-13
Add deploy section

harrietgrace:
This adds a new Deployments section to the documentation, with an overview of a bunch of common ways you can deploy with Buildkite, and a guide for deploying to Kubernetes from Buildkite pipelines.

PR #491, opened 2019-06-14
Add deploy overview doc

harrietgrace:
@toolmantim here's the deploy overview doc draft 🙇🏻‍♀️


toolmantim:
Thanks @harrietgrace! Will give this a go next week.

toolmantim:
I've moved this into #469 and will finish it off there instead, and we can deploy in one hit!

PR #503, opened 2019-07-16
Add timed_out build state

keith:
I've seen this when the `timeout_in_minutes` value is exceeded, but it
wasn't documented.


keith:
@toolmantim mind taking a look?

toolmantim:
Hi @Keith! Thanks for taking the time to send through the PR.

I looked into this for you yesterday, and turns out there's no build state called `timed_out` — that's actually a state of the individual jobs inside the build. I double checked the code, and the docs seem to line up perfectly, so I might close this issue for now.

Happy to help to figure out what you're trying to do though, or what led to this one 🤔

keith:
Ok that makes sense, but as far as I can tell the only mention of `timed_out` is this one: https://github.com/buildkite/docs/blob/master/pages/apis/rest_api/jobs.md.erb#L7 so it would be great to see a list of all the job states, just like how this is a list of all the build states

PR #504, opened 2019-07-17
Fix the SSO heading

toolmantim:
Fixes the following bug on https://buildkite.com/docs/integrations/sso:
image


plasticine:
Blerg, thanks for sorting this out, @toolmantim!

PR #515, opened 2019-08-10
Add Annotations REST API to docs and sidebar

ticky:
🥫!


toolmantim:
This sidebar isn’t used in the Rails app yet btw, so you’ll need to add it there too in the docs bump.

ticky:
> This sidebar isn’t used in the Rails app yet btw, so you’ll need to add it there too in the docs bump.

Oh is THAT why I couldn't get it to work locally 😅

PR #518, opened 2019-08-16
Document BUILDKITE_STEP_ID

toolmantim:
🥫


toolmantim:
Actually this is incorrect…

```yml
steps:
  - command: echo specs
    id: specs
```

results in:

```
BUILDKITE_JOB_ID="80818485-83fb-4cc0-afaf-85719f5a934b"
BUILDKITE_STEP_ID="5726eaa1-b695-4698-95f4-3375971a54b0"
BUILDKITE_STEP_KEY="specs"
```

We should probably document these separately. Will open a new PR!

PR #519, opened 2019-08-19
Add STEP_* env vars

harrietgrace:
Env vars page update for https://github.com/buildkite/buildkite/pull/4063:

- deprecates the `BUILDKITE_STEP_IDENTIFIER` var
- adds the `BUILDKITE_STEP_ID` and `BUILDKITE_STEP_KEY` vars to the main list

Also snuck in a little alphabetise of the vars list - the S3 vars were at the end of the S's when they should have been at the start. 

@toolmantim do those descriptions of the new STEP_* vars look legit?

### DIFF
--- a/pages/pipelines/environment_variables.md.erb
+++ b/pages/pipelines/environment_variables.md.erb
@@ -344,24 +344,6 @@ How many times this job has been retried. The value cannot be modified.
 
 Example: `"0"`
 
-<h3 class="h3-caps">BUILDKITE_SHELL</h3>
-
-The value of the `shell` [agent configuration option](/docs/agent/v3/configuration). The value cannot be modified as an environment variable.
-
-Example: `"/bin/bash -e -c"`
-
-<h3 class="h3-caps">BUILDKITE_SOURCE</h3>
-
-The source of the event that created the build. The value cannot be modified.
-
-Values: `"webhook"`, `"api"`, `"ui"`, `"trigger_job"`, or `"schedule"`
-
-<h3 class="h3-caps">BUILDKITE_SSH_KEYSCAN</h3>
-
-The opposite of the value of the `no-ssh-keyscan` [agent configuration option](/docs/agent/v3/configuration). The value cannot be modified as an environment variable.
-
-Values: `"true"` and `"false"`
-
 <h3 class="h3-caps">BUILDKITE_S3_ACCESS_KEY_ID</h3>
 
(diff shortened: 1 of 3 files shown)


pages/pipelines/environment_variables.md.erb
```diff
@@ -392,6 +374,36 @@ The secret access key for your S3 IAM user, for use with [private S3 buckets](/d
...
+
+<h3 class="h3-caps">BUILDKITE_STEP_ID</h3>
+
+A unique string that identifies a step. This value cannot be modified.
+
+Example: "080b7d73-986d-4a39-a510-b34f9faf4710"
+
+<h3 class="h3-caps">BUILDKITE_STEP_KEY</h3>
+
+A unique string set by the user to identify a step. This value cannot be modified as an environment variable.
```
toolmantim: It might be a good idea to mention that this is from the `key` property set in the command step? (i.e. where is this set by the user)? We could link to the command step doc maybe? Also it might be nicer to say "you" instead of "the user"?

harrietgrace: I've borrowed the same format that we used in the vars that come from the agent config options, but still kept the bit about it being a unique string that you set. What do you think?

pages/pipelines/environment_variables.md.erb
```diff
@@ -392,6 +374,36 @@ The secret access key for your S3 IAM user, for use with [private S3 buckets](/d
...
+
+<h3 class="h3-caps">BUILDKITE_STEP_ID</h3>
+
+A unique string that identifies a step. This value cannot be modified.
+
+Example: "080b7d73-986d-4a39-a510-b34f9faf4710"
+
+<h3 class="h3-caps">BUILDKITE_STEP_KEY</h3>
+
+The value of the `key` [command step option](/docs/pipelines/command-step#command-step-attributes), a unique string set by you to identify a step. This value cannot be modified as an environment variable.
```
toolmantim: `s/option/attribute/`?

toolmantim: Ah, I see we use option above too! 🤔 Whatever really!

harrietgrace: I like attribute better for this one, given that the link to the command step docs has 'attribute' in it 👍🏻

harrietgrace:
@toolmantim do you think this needs a changelog? We're deprecating a variable... but I don't know if a single var is worth such an announcement?

toolmantim:
Yeah probably! But we never announced it in the first place, so it’s probably a “We’ve added step keys, and if you used the old `id` we’ve deprecated that” type thing?

PR #521, opened 2019-08-22
Add list of supported CSS classes in annotations

### LINKED ISSUE #387: Specifically list all supported CSS classes
We've got some great docs here on styling annotations:

https://buildkite.com/docs/agent/v3/cli-annotate#supported-css-classes

@toolmantim extracted them out into a gist:

https://gist.github.com/toolmantim/7d78e02e4b726a28b0601eef43fad0eb

Perhaps we can just list what we support? It's neat to be able to see all the classes at a glance!

harrietgrace:
@keithpitt was this the kind of thing you'd imagined with #387? 

This adds a big list of all the classes we support from Basscss. I've done some further grouping of the classes to separate them out a bit, does it look like they're all in appropriate places @ticky?

Closes #387

### DIFF
--- a/pages/agent/v3/cli_annotate.md.erb
+++ b/pages/agent/v3/cli_annotate.md.erb
@@ -71,6 +71,43 @@ We accept the following parts of version 8.0 of Basscss within annotations:
   * `list-reset`
 * [Type Scale](http://basscss.com/#basscss-type-scale)
 
+An exhaustive list of classes that annotations support can be found below:
+
+```
+bold regular italic caps underline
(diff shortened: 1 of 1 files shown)


ticky:
I was hesitant to do this primarily because the class names only tell some of the story, and I didn't think duplicating the basscss docs made sense. This list looks reasonable otherwise!

keithpitt:
Perfect, thanks @harrietgrace!

harrietgrace:
Thanks @ticky and @keithpitt! 🙏🏻

PR #524, opened 2019-08-27
Outline new section on timestamps

harrietgrace:
We've had 9 tickets come through support over the last year about the differences between our job timestamps. Usually it's confusion around `scheduled_at` being earlier than `created_at`, or it's about what `runnable_at` even is.

This PR adds a new section to the build REST API doc that lays out what each of the timestamps is for both builds and jobs. As these values are mostly asked about in relation to the build, I think it sits well in this doc rather than split across the builds and jobs API pages. I'll be adding a link to it from the jobs page.


harrietgrace:
@plasticine could you please have a check through the descriptions I've given for each of these timestamps? 😊🙏🏻

pages/apis/rest_api/builds.md.erb
```diff
@@ -615,3 +615,70 @@ curl -X PUT "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pi
...
+
+The timestamps are available via both the GraphQL and REST APIs. They differ slightly between the build and job objects.
+
+Each <em>build</em> is provided with the following timestamps:
+
+<table>
+<tbody>
+  <tr>
+    <th><code>scheduled_at</code></th>
+    <td>The time the build was created</td>
```
plasticine: For a `Build`  the value of `scheduled_at` is the time the Build is created, however for a `Job` the value of `scheduled_at` is actually the time that the scheduler process processes the job. It’s usually pretty much immediately following the job being created however, so often there is only a couple hundred ms in it.

harrietgrace:
It would be rad to have as part of the main docs! What do you think about shipping this where it is at the moment with plans to move it later? I've booked in a whole big review of our getting started/intro docs for the new year, and I reckon this would be an awesome thing to have as part of those docs. 

> explicitly list all the values in full for each object type

What do you mean there? Like examples? Or more in-depth descriptions?

I've updated the scheduled_at for the job a little, but left everything else for the minute.

plasticine:
> It would be rad to have as part of the main docs! What do you think about shipping this where it is at the moment with plans to move it later? 

Oh, hell yeah — for sure!

> > explicitly list all the values in full for each object type
>
> What do you mean there? Like examples? Or more in-depth descriptions?

Oh, I just meant that at the moment we list the values as though they’re shared, and then call out the differences — I wonder if given the slight differences we might better just listing all of the timestamps for builds and all of them for jobs.

harrietgrace:
@plasticine That first list is just for builds, further down the page there's all the job ones in their own table. So we've kind of already got that? Yeah/nah? 🤔

plasticine:
@harrietgrace Yep, they do but I wonder if the `Jobs have an extra value:` bit might be worth tweaking as when read from start to finish it kinda groups it a bit with the build fields? Tbh though I’m prolly 4/10 on it, and this doc is awesome so we should probably ship it regardless. Sorry for lack of clarity/timeliness on this 😕

harrietgrace:
Thanks @plasticine! Ah yeah, I can see that. I agree it would sit better without that rando extra callout! I've removed it and put in a line about how the values are different for builds even though the timestamps are the same 👍🏻

PR #527, opened 2019-09-03
Add JIT provisioning to FAQ

### LINKED ISSUE #451: Mention JIT provisioning support in SSO overview
JIT (Just In Time) provisioning is something our SSO system supports, and something people ask about, but I can't find the words "JIT" or "provisioning" in the SSO overview. Let's add it to the FAQ? I don't think there's anything specific we need to say at the moment.

harrietgrace:
As @toolmantim pointed out in #451, we don't mention JIT provisioning anywhere in our SSO docs, even though it's a thing we have always supported.

This PR adds a (very) brief FAQ containing those keywords:
image

Closes #451


toolmantim:
✨

PR #528, opened 2019-09-03
Add provider settings properties to pipelines API

### LINKED ISSUE #481: Document provider_settings for the pipelines API
Currently the section for `provider_settings` is undocumented, and its hard to find what parameters are accepted or how to enter them. I did some checking on https://github.com/buildkite/go-buildkite/blob/master/buildkite/providers.go#L48 and it seems not all settings are "documented" or codified there.

eg. `separate_pull_request_statuses`

We are currently building a quick tool to bootstrap pipelines form a yaml/json definition till we/someone works on https://github.com/buildkite/cli/issues/48

**Note**
This probably should be another issue, but I also notice the API accepts YAML as well: https://github.com/buildkite/python-docker-example/blob/master/.buildkite/template.yml maybe it would be good to add that.

harrietgrace:
As pointed out in #481, we don't have a full list of the properties available in the `provider_settings` object. 

This PR adds a few tables that contain the different properties available, split up by provider. I duplicated the two for Bitbucket server because there were so few for it. Bitbucket Cloud, GitHub, and GitHub enterprise all share a bunch of properties, so I grouped those together and then added an overflow table for the extra ones that only GitHub and GitHub enterprise support. 

Depends on https://github.com/buildkite/docs/pull/530 being merged first. 

Closes #481


harrietgrace:
@toolmantim what do you think about the layout of this new bit? I had a hard time figuring out how to present all the options for each provider without too much duplication, and this is where I ended up.

pages/apis/rest_api/pipelines.md.erb
```diff
@@ -332,6 +332,130 @@ Optional [request body properties](/docs/api#request-body-properties):
 </tbody>
 </table>
 
+Provider settings properties
+
+Bitbucket Cloud, Bitbucket Server, GitHub, and GitHub Enterprise all have optional `provider_settings`. Other source providers such as Gitlab have no optional settings; however they will return the read-only `repository` field in the `provider_settings` object of a pipeline REST API response.
```
toolmantim: The readonly `repository` field section could be a bit confusing, because the rest is talking about the params you pass in, and this is talking about the response? I was wondering what the response looks like too? It is `{"provider_settings": {"repository": "git@myserver.com"}}` type thing?

harrietgrace: Ah right. I didn't realise that was part of the response, that makes much more sense! Do you think we need to include a separate example response for this section? It is listed up in the create provider response:

image

harrietgrace: Oh, clearly I did realise at the time I wrote it given that it says that it's part of the response 🤦🏻‍♀️I'll just remove that bit I think.

harrietgrace:
@toolmantim Thanks for the suggestions! I've moved the provider settings into its own section at the end and put links to it from the create and update bits. I think it feels way better with it like this 👌🏻

PR #531, opened 2019-09-11
Add a Slack integration docs page

toolmantim:
Similar to https://buildkite.com/docs/integrations/amazon-eventbridge this adds a new page to docs for our Slack integration


ticky:
Exciting!

harrietgrace:
Added some more words/edits, and expanded on the notes at the end 😊 Let me know what y'all think!

ticky:
We should probably be unifying our wording between this and buildkite/buildkite#4228.

Also, I’m not sure where October 1st came from; the date will be three months from when we pass Slack’s App Directory review, which they’ve asked us to submit to by the 15th of October, so I think we need to be non-specific in the first pass and then add the correct date later.

ticky:
Thanks for the edits @harrietgrace, and thanks for taking the steps and expanding on them! 😄

I’ve fixed one typo I found, and removed the specific date. We’ll need to go through Slack’s App Directory review process before we know the specific date. Once we know, we can merge buildkite/buildkite#4228 and ship an update to this!

PR #538, opened 2019-10-01
Add API access tokens page

harrietgrace:
Adding docs for the new API access token audit page ✨

@eleanorakh here's an outline with some notes in it 👋🏻 Is there anything missing, or anything incorrect?


eleanorakh:
All great words @harrietgrace! You nailed the outline! The only thing is I'm not entirely sure is having it under pipelines. It doesn't feel like the best place, as it isn't pipeline specific or constrained to pipelines -- although having it under a workflow headings does feel right. I'll have a think!

harrietgrace:
This needs a new image of the Token page, but other than that it should be ready!

eleanorakh:
Screen Shot 2019-10-07 at 10 10 31 am
how's  this screenshot?

toolmantim:
Perfect! The index pages did change a little bit too, if you had it handy @eleanorakh

eleanorakh:
@toolmantim I've updated all the  screenshots 👍🏼

PR #539, opened 2019-10-02
Update the cancel grace period in Signal Handling

arturopie:
The cancel grace period can be customized


harrietgrace:
Thanks @arturopie! Sorry it's taken me so long to get this merged; cancel-grace-period has been a hot topic around here and I wanted to make sure there was nothing else to add before I put this live. 

I've added back in the default grace period, so that people know that it's there as a backup and they don't have to set the config option. Thanks for taking the time to PR! Much appreciated 🙇🏻‍♀️

PR #545, opened 2019-10-16
Update conditionals docs to escape $ char

### LINKED ISSUE #542: Conditionals with dollar signs need to be escaped in examples
See https://github.com/buildkite/agent/issues/1102

harrietgrace:
Updates the note at the end of the 'supported syntax' section to be a troubleshooting note and adds info about escaping $. Also updates the example code at the end of the doc so that both the instances of $ are escaped with \. 

Closes #542

### DIFF
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -83,7 +83,7 @@ The below expressions are supported by the `if` attribute:
 
 <div class="Docs__troubleshooting-note">
 <h1>Formatting regular expressions</h1>
-<p>When using regex, the expression must be on the right-hand side of the operator.<br>Any <code>$</code> signs must be escaped so they aren't interpolated as variables, e.g. <code>$$</code> or <code>\$</code>.</p>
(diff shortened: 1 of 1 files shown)


pages/pipelines/conditionals.md.erb
```diff
@@ -81,8 +81,9 @@ The below expressions are supported by the `if` attribute:
 
 
 
-<div class="Docs__note">
-<p>When using regex, the expression must be on the right-hand side of the operator.</p>
+<div class="Docs__troubleshooting-note">
+<h1>Using regex</h1>
```
toolmantim:
```suggestion
<h1>Formatting regular expressions</h1>
```

pages/pipelines/conditionals.md.erb
```diff
@@ -81,8 +81,9 @@ The below expressions are supported by the `if` attribute:
...
+<p>When using regex, the expression must be on the right-hand side of the operator.<br>Any <code>$</code> signs must be escaped so they aren't interpolated as variables, e.g. <code>$$</code> or <code>\$</code>.</p>
```
toolmantim:
```suggestion
<p>When using regular expressions in conditionals, the regular expression must be on the right hand side, and the use of the <code>$</code> anchor symbol must be escaped to avoid <a href="/docs/agent/v3/cli-pipeline#environment-variable-substitution">environment variable substitution</a>. For example, to match branches ending in <code>"/feature"</code> the conditional statement would be <code>build.branch =~ /\/feature$$/</code>.
```

PR #548, opened 2019-10-17
Add Python Buildkite api wrapper

pyasi:
Adding a link to a python client for the Buildkite REST API. Available via `pip` 📦


harrietgrace:
Thanks for the PR @pyasi! 🙏🏻 And thanks for making a Python client too 😄

PR #549, opened 2019-10-17
Add dependencies docs

harrietgrace:
This PR adds all the docs needed for the DAG launch. 

- [x] Standalone page about dependency management
- [x] Add `depends_on` to each step type doc 
- [x] Add page for Input Step
- [x] Expanding all the words into proper sentences
- [x] Merge in conditionals update
- [x] Editing

@matthewd what do you think about the outline of the Managing Step Dependencies doc? Anything else it needs? Anything incorrect? I've also started on the Input step doc, if you wanted to squiz that too (it's mostly copied from the block step doc).


pages/pipelines/dependencies.md.erb
```diff
@@ -0,0 +1,65 @@
...
+  - command: "thing.sh"
+  - wait
+  - command: "blah.sh"
+    depends_on: ~
+```
+
+## order of operations
+
+- `if` then `depends_on` then `concurrency`
+- if the step you're dependent on doesn't exist for any reason, the current step will run as soon as the rest of the jobs complete
```
matthewd: The only time a step will run with a dependency unsatisfied is if the depended-upon step is present but disabled via `if` (or `branches`) -- if the build settles and the named dependency hasn't appeared, the build will fail, still without running the depending step

pages/pipelines/dependencies.md.erb
```diff
@@ -0,0 +1,65 @@
...
+  - command: "a.sh"
+  - command: "b.sh"
+  - wait
+  - command: "c.sh"
+  - command: "d.sh"
+```
+
+- block steps perform the same function, as well as requiring the unblocking either manually or via an API call
+
+- if you are collecting information with your block steps using the `prompt` or `fields` attributes, but don't require the actual blocking behaviour, you can use an input step
```
matthewd: Arguably the "blocking" part of the behaviour remains -- it will stop its dependents from running until the input has been supplied -- it's just the vacuuming up of implicit dependencies that's removed. (I suspect I'm just picking on an ambiguously-worded notation here, but figure I might as well flag it just in case.)

pages/pipelines/input_step.md.erb
```diff
@@ -0,0 +1,223 @@
...
+    <td><code>branches</code></td>
+    <td>
+      The <a href="/docs/pipelines/branch-configuration#branch-pattern-examples">branch pattern</a> defining which branches will include this input step in their builds.<br>
+      <em>Example:</em> <code>"master stable/*"</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>if</code></td>
+    <td>
+      A boolean expression to restrict the running of the step. See <a href="/docs/pipelines/conditionals">Using conditionals</a> for supported expressions.<br>
```
matthewd: Maybe a note for the existing block step page too, but "the running of the step" feels a bit off here? 🤷🏻‍♂️

pages/pipelines/input_step.md.erb
```diff
@@ -0,0 +1,223 @@
+# Input Step
+
+An *input* step is used to collect information from a user.
```
matthewd: Given they're both going to exist and be documented next to each other, I think it's probably worth some heavy cross-promotion of the fact their options are identical -- so people can feel confident swapping between them as needed, without feeling obliged to re-check that all the options are still valid.

pages/pipelines/input_step.md.erb
harrietgrace: Yeah, it's totally awkward. I've come up with some alternates, would love to get your feels on them!

A boolean expression:
to restrict when a step runs
to restrict when this step is added to a build
to restrict when the step is run in the pipeline
to define the circumstances when this step will be added to a build from this pipeline
to limit the step to only run under specific circumstances
to prevent the step from being added to the build under certain conditions

harrietgrace:
Thanks @matthewd! I've done a bunch of editing and (I think) added in your feedback. Could you please have another look over it and see how it's sitting? 😊

harrietgrace:
Oh, I've also had a go at a description for the `depends_on` and `allow_dependency_failure` attributes. Are these accurate?

image

keithpitt:
Those descriptions for `depends_on` and `allow_dependecy_failure` are ace. We support a few different formats to `depends_on`

```yaml
depends_on: "test-suite"

depends_on:
  - "test-suite"
  - "another-thing"

depends_on:
  - step: "test-suite"
  - step: "another-thing"

depends_on:
  - step: "test-suite"
    allow_failure: true
  - step: "another-thing"
    allow_failure: false
```

`allow_dependency_failure` sets `allow_failure: true` for each dependency. By default it's `false` (which means the default for each `allow_failure` is false). You can control on a per-dependency basis it's `allow_failure` option.

harrietgrace:
@keithpitt thanks for the examples! Can you do && like you can in a command step? Like this:

```yml
depends_on: "step-one && step-two"
```

matthewd:
> Can you do && like you can in a command step?

No; that's shell command syntax. The only way to specify multiple dependencies is via the YAML array syntax (either `-`-prefixed lines, or `['a', 'b']`). I'd incidentally suggest using that (@keithpitt's second example) as the default style in the documentation: if people aren't sure about syntax details, a single-element array will still work just as well, so it degrades better than having someone try to figure out the multi-value version from a bare `depends_on: foo`. (The latter two are more advanced, and much more verbose, so I think it's fine to restrict them to where they're actually necessary.)

pages/pipelines/input_step.md.erb
matthewd: How about something like "that [omits/disables/ignores] the step when false"?

harrietgrace:
Okie dokie, I've added some more code samples and a bit more detail to the dependencies page about the different syntax options for `depends_on` and the two different options for the allow failure bit. 

I also updated the description for `depends_on` in the step docs so that it links to the Managing Step Dependencies page. I thought about adding in more bits to those pages to list the 'step' and 'allow_failure' stuff, but I thought we might be able to get away with just a link to the dependencies page. What do y'all think?

pages/pipelines/input_step.md.erb
harrietgrace: Yeah cool. I'll do the update for that in the conditionals update PR I've got on the go 👌🏻

matthewd:
> I thought about adding in more bits to those pages to list the 'step' and 'allow_failure' stuff, but I thought we might be able to get away with just a link to the dependencies page.

Sorry, I missed this question. Yeah, I think it seems fine to leave the details of the different possible shapes for `depends_on` to its own doc -- I did [wonder](https://github.com/buildkite/docs/pull/549#discussion_r342614969) whether we want some small "this sets the default for the dependency-specific [`allow_failure`](#) value" mention on `allow_dependency_failure`, just in case someone's staring at that while wanting narrower control, but I think we could safely do that, if we so chose, still without expanding on the rest.

pages/pipelines/dependencies.md.erb
```diff
@@ -0,0 +1,116 @@
...
+
+All steps in pipelines have implicit dependencies, often managed with wait and block steps. To manually change the dependency structure of your steps, you can define explicit dependencies with the `depends_on` attribute.
+
+<%= toc %>
+
+## Implicit dependencies with wait and block
+
+[Wait](/docs/pipelines/wait-step) and [block](/docs/pipelines/block-step) steps provide an implicit dependency structure to your pipeline.
+
+By adding these steps to your pipeline, the Buildkite Agent will automatically know which steps need to be run in serial and which can be run in parallel.
```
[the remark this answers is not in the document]
harrietgrace: I think lower case is good 👍🏻

pages/pipelines/block_step.md.erb
```diff
@@ -52,6 +56,20 @@ _Optional attributes:_
...
+    <td><code>depends_on</code></td>
+    <td>
+      A list of step keys that this step depends on. This step will only run after the defined step has completed. See <a href="/docs/pipelines/dependencies">managing step dependencies</a> for more information.<br>
+      <em>Example:</em> <code>"test-suite"</code>
+    </td>
+   </tr>
+   <tr>
+    <td><code>allow_dependency_failure</code></td>
+    <td>
+      Whether to continue to run this step if any of the steps defined in the <code>depends_on</code> attribute fail.<br>
```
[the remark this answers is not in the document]
harrietgrace: I think leave it as is for now, we can always add it in later if people are asking questions about it. I think it would get too busy if we tried to smush the extra info in there.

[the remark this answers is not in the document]
harrietgrace: I've gone through and updated all the `depends_on` and `allow_dependency_failure` attributes using a combo of your other suggestions 😄

pages/pipelines/block_step.md.erb
```diff
@@ -2,6 +2,8 @@
 
 A *block* step is used to pause the execution of a build and wait on a team member to unblock it via the web or the [API](/docs/apis/rest-api/jobs#unblock-a-job).
 
+An block step is functionally identical to an [input step](/docs/pipelines/block-step), however a block step creates implicit dependencies to the steps before and after it.
```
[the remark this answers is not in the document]
harrietgrace: Yes definitely!

pages/pipelines/dependencies.md.erb
```diff
@@ -0,0 +1,116 @@
...
+steps:
+  - command: "tests.sh"
+    key: "tests"
+  - command: "build.sh"
+    key: "build"
+    depends_on: "tests"
+    allow_dependency_failure: true
+```
+
+However, even if you continue to run the next step, if there are any failures at all, the build will still end up failing.
```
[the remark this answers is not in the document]
harrietgrace: I stuck it in with the first couple of sentences - I think it's a pretty important point to note, so I'd rather put it right up front than at the end.

pages/pipelines/dependencies.md.erb
```diff
@@ -0,0 +1,116 @@
...
+
+To explictly allow or deny failures on an individual dependency basis, use the `allow_failure` attribute on a defined step dependency.
```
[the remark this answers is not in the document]
harrietgrace: I like the start, but I'm not sure about the 'expanded' bit at the end. What do you think of:

> For finer control, you can explictly allow or deny failures on an individual dependency basis using the `allow_failure` attribute with a step dependency.

harrietgrace:
Epic, thanks so much @matthewd! 🙇🏻‍♀️ I've integrated all your words/thoughts. How do you feel about shipping this tomorrow? 😱We can leave it sans sidebar link until the feature goes live? 😄

harrietgrace:
Chatted with @matthewd about merging this/deploying - aiming for end of the week 👍🏻

harrietgrace:
I just swapped around the first two sections so that 'Defining Explicit Dependencies' is first followed by 'Implicit dependencies with wait and block'. Reading through it after the changelog, it felt weird having implicit first when the document is about explicit deps.

matthewd:
Hmm... I disagree: the changelog is about explicit deps, but the doc is about deps in all their forms (though implicit gets less words, because it's simpler, because it doesn't offer any options).

Putting implicit dependencies first ensures people can't miss that it's a thing (if you start thinking you're only defining explicit dependencies and extras magically show up, that could be quite surprising), and encourages people to consider them as the default, semantically simpler solution if they're sufficient for their needs. (It also better mirrors the structure of the opening paragraph, which I think would change substantially if it was reversed.)

harrietgrace:
You speak sense @matthewd! 🙇🏻‍♀️ Swapped back. I'll do some edits to the changelog so that it works better linking to this page.

PR #550, opened 2019-10-21
Update environment_variables.md.erb

GaryPWhite:
https://github.com/buildkite/agent/pull/1041

noted that there was a duplicate of `BUILDKITE_CANCEL_GRACE_PERIOD` -- putting a proper definition for `BUILDKITE_CANCEL_SIGNAL`

### DIFF
--- a/pages/pipelines/environment_variables.md.erb
+++ b/pages/pipelines/environment_variables.md.erb
@@ -136,7 +136,9 @@ Default: `10`
 
 <h3 class="h3-caps">BUILDKITE_CANCEL_SIGNAL</h3>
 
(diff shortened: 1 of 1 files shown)


harrietgrace:
Hey @GaryPWhite, ah nice catch, thanks! 🙏🏻 I've updated the cancel signal definition to match the other vars that come from the agent config options. 

I liked that you mentioned the different defaults for the agent versions! However, we won't be adding the 4.x agent information to the official documentation until it's the stable agent - we like to keep the pre-release info to just the GitHub repo 😊

Thanks for taking the time to PR 👌🏻

PR #552, opened 2019-10-22
Fix type in deploying to kubernetes doc

jorourke:


harrietgrace:
Good catch @jorourke! Thanks for taking the time to PR 🙏🏻 This file has seen a few typos today, so I've merged in master.

PR #554, opened 2019-10-22
Custom SSO and ADFS guides update

harrietgrace:
Update: Doing a proper update for the Custom SAML and ADFS guides to go along with 
https://github.com/buildkite/buildkite/pull/4364#issuecomment-546754779

------
As @lox pointed out the other day, for a couple of SSO providers we still tell people to email us for their URL, but we could also be pointing them towards the graphql guide. They can generate their own URL using the SAML setup instructions. 

This PR adds a link to the graphql guide just below where we tell them to email us for their URL.

image

@lox did you want people to have more info than that?


keithpitt:
@harrietgrace I made some changes to how you can get the magical URL - I've surfaced it in the UI. Here's a screenshot from development:

image

I can take you through it when you're online - it might be good to include some extra words around where you can find the URL.

lox:
Will remove myself from review until after that chat!

harrietgrace:
@toolmantim could you please have a look at the [Custom SAML page](https://buildkite-docs-review-pr-554.herokuapp.com/docs/integrations/sso/custom-saml) and let me know what you think of the new structure? I couldn't figure out a better way to show that there're two options for each of the first two steps of the process 🤔

pages/integrations/sso/custom_saml.md.erb
```diff
@@ -4,22 +4,35 @@ Any identify provider that supports SAML 2.0 can be used to authorize access to
...
+After setting up your provider, you'll need to set up your SSO provider in your Buildkite organization's Single Sign On Settings.
 
 <div class="Docs__note">
-<p class="Docs__note__heading">Before starting your provider setup</p>
-<p>Email the Buildkite team at support@buildkite.com to request your unique SSO service URL, you will need this during set up.</p>
+  <p>You can also set up SSO providers manually with GraphQL. See the <a href="/docs/integrations/sso/sso-setup-with-graphql">SSO Setup with GraphQL Guide</a> for detailed instructions and code samples.</p>
 </div>
 
-### Step 1. Option A - Setting up with meta-data
+## Option A - Setting up your IdP using a meta-data URL or ACS URL
```
toolmantim: Yeah, like you mentioned, this nav/structure is getting pretty wild!

image

What if we made the "Option A/B"'s third-level headings?

```suggestion
### Option A - Setting up your IdP using a meta-data URL or ACS URL
```

harrietgrace: I wanted the different options to be visible up front/as soon as you get to the page, so I was hoping to keep it in the TOC somehow... but I don't see how that's possible. Do you think we could figure out some way of showing people that there's two ways to do things that's not the TOC? Or just leave them to scroll and figure it out? Could it be solved by another line of text in that top intro paragraph section before the TOC?

toolmantim: @harrietgrace want to have a quick chat about it? Maybe if we talk it through we'll have some other ideas… at the moment I'm stumped too.

harrietgrace:
I've redone the structure to be two distinct sections. Not sure about the words/titles, I just threw it down to get an idea of what it would be like. Thoughts @toolmantim?

PR #555, opened 2019-10-22
Add new conditionals vars

harrietgrace:
Adds `build.env()` to the supported variables list on the Conditionals page. I've also alphabetised the table so it's easier to find things. 

@keithpitt could I please get your help with an example? Also, is the description correct? 😄

image

Edit: Adds a bunch more new supported vars too! (From https://github.com/buildkite/buildkite/pull/4348)

- [x] Add new vars
- [x] Update description for `if` in the step docs (from #549)
- [x] Fix build.env description
- [ ] Add new examples

### DIFF
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -48,7 +48,7 @@ _Optional attributes:_
   <tr>
     <td><code>if</code></td>
     <td>
-      A boolean expression to restrict the running of the step. See <a href="/docs/pipelines/conditionals">Using conditionals</a> for supported expressions.<br>
+      A boolean expression that omits the step when false. See <a href="/docs/pipelines/conditionals">Using conditionals</a> for supported expressions.<br>
       <em>Example:</em> <code>build.message != "skip me"</code>
     </td>
   </tr>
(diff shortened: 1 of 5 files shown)


keithpitt:
`build.env()` is a function. An example would be `build.env("CUSTOM_ENVIRONMENT_VARIABLE")`

Maybe the description is like "This function returns the value of the environment passed into it's first argument. If environment variable has not been set, it'll return an empty string"

Hrm...I wonder if it'd be better if `build.env` returned null if the environment variable doesn't exist? That way you'd be able to tell the difference between an ENV var that's empty "" and an env var that's not been set `null`

keithpitt:
@harrietgrace I decided that `build.env` should return `null` if there's no env with it set. That way folks can tell the difference between an empty ENV, and one that hasn't been set.

harrietgrace:
@keithpitt awesome, thanks! I've updated the description and edited it a bit. Do you have a real-life-ish example env var and value we could put in? I've just put in filler text into the example for now. 

The new vars are also in there now with draft descriptions 👍🏻

harrietgrace:
Added an extra example with @matthewd's help! Merging this so I can get it in with the Step Deps PR 👍🏻

PR #557, opened 2019-10-28
Update git clean flags default

harrietgrace:
As mentioned in ye olde PR https://github.com/buildkite/agent/pull/875, the `git-clean-flags` has defaulted to `-ffxdq` since v3.8.0. 

This PR adds some more version info to the agent config option. The env var for this option already has the correct default value.


harrietgrace:
@toolmantim what do you think of this formatting for Agent version info?

image

PR #561, opened 2019-10-29
Add TLS client cert info to GitHub Enterprise guide

harrietgrace:
Adds a bit of info and a screenshot describing the TLS client section of the GitHub Enterprise setup, from https://github.com/buildkite/buildkite/pull/3421. 

Does that cover it @sj26? Is there anything else we need here?


sj26:
Looks right to me!

Issue #511, opened 2019-08-01
RESTAPI: create pipelines


elruwen:
Moin!

The field default_branch is not documented, but it works. So please add it ;)

Cheers
Ruwen

harrietgrace:
Thanks for pointing that out Ruwen! 🙏🏻 I'll leave this issue open until it gets added 😊

PR #564, opened 2019-10-30
Add default_branch api field

harrietgrace:
As brought up in #511, the `default_branch` option is available on create and update in the pipelines endpoint, but wasn't listed in the docs. 

This PR adds `default_branch` as well as alphabetising the attribute tables.


harrietgrace:
The UI already mentions the pipeline page filtering! 👍🏻 I've added to this as well. Thanks @toolmantim!

PR #567, opened 2019-10-31
Add health-check-addr agent config option and env var

harrietgrace:
There was a new Agent config option added in https://github.com/buildkite/agent/pull/1066 - `health-check-addr`. 

This PR adds the config option and the associated env var. @lox what would be an appropriate example to use for the 'address:port'?


lox:
I'd suggest `localhost:8080`.

PR #574, opened 2019-11-04
access token api returns uuid instead of id

intuxicated:
Access Token API `https://api.buildkite.com/v2/access-token` is returning `uuid`.


harrietgrace:
Brilliant, thanks @intuxicated! And thanks for taking the time to PR 😊🙏🏻

PR #575, opened 2019-11-06
add missing env and steps to pipeline docs

intuxicated:
[List pipelines](https://buildkite.com/docs/apis/rest-api/pipelines#list-pipelines) will return `env` and `steps` like other pipeline API responses.


harrietgrace:
Sweet, thanks again @intuxicated! 👌🏻

PR #578, opened 2019-11-13
Add section about exit codes

### LINKED ISSUE #448: Add more detail about agent exit codes
From the [Buildkite community forum](https://forum.buildkite.community/t/doc-example-mismatch/393/2):

> A 255 happens when an agent is canceled or interrupted, but still terminates cleanly. The -1 exit code happens when Buildkite loses contact with the agent or it stops reporting to us. I’ll add a note to the docs about them so that the difference between the two codes is clearer 👍🏻

harrietgrace:
From a [question on the Forum](https://github.com/buildkite/docs/issues/448) back in March - we only have the tiniest bit of info about our Agent exit codes in the docs. 

This PR adds a new section to the Agent overview page with a table describing the exit codes. 

Closes #448


pages/agent/v3.md.erb
```diff
@@ -59,3 +61,30 @@ The agent also accepts the following two signals directly:
 
 * `SIGTERM` - Instructs the agent to gracefully disconnect, after completing any job that it may be running.
 * `SIGQUIT` - Instructs the agent to forcefully disconnect, cancelling any job that it may be running. 
+
+## Exit Codes
+
+The Agent reports its activity to Buildkite using exit codes. The most common exit codes and their descriptions can be found in the table below.
```
pda: I'm wondering if this is clear/true.

I'd say the agent reports the exit status of its child processes to Buildkite via the agent API. That covers the `0` and `1` case documented below, and presumably most/all other arbitrary exit status codes from the job.

I gather `-1` means Buildkite never received an exit status code from the agent, and so it's using `-1` as a placeholder to represent that condition. So that case doesn't fit under "the Agent reports its activity to Buildkite ...".

I also have some concerns about exit status `-1` and `255` having distinct meanings ,because I think in some contexts the two values are equivalent; when interpreted as a signed 8-bit integer (`int8`), which apparently some shells/unixy things do in some cases. But... that's probably not an issue for these docs :)

So, I wonder if these docs can say something along the lines of:

> The Agent reports the exit status code of its child job processes to the Buildkite agent API, e.g. `0` for success, and `1` being a common error exit code. If the agent is asked to shut down during a build, it reports a special exit code of `147` or `255`. If Buildkite loses contact with the agent, an exit status code of `-1` is assumed.

(I remain confused about the difference between `147` and `255`... I'd have thought "Killed by signal SIGTERM" was just a specific case of "canceled or interrupted but still terminated cleanly"... unless it really means "Killed by SIGKILL" rather than SIGTERM).

Sorry if this isn't helpful, let me know if it's useful to chat more.

harrietgrace: Thanks @pda, this is super helpful! 🙇🏻‍♀️🙌🏻Thanks for looking into it so much 👌🏻

As far as I'm aware, the 255 case is actually a sort-of bug? But people rely on it now, so it's not going to get changed/fixed anytime soon. Here's the forum post that started me down this track: https://forum.buildkite.community/t/doc-example-mismatch/393/2. There's also this post from Lox about signal handling in the Agent: https://3.basecamp.com/3453178/buckets/11763568/messages/1804435204 which is where I got most of my info from. 

I really like:
> The Agent reports the exit status code of its child job processes to the Buildkite agent API

But I'd also like to keep the table rather than explaining in a paragraph. I'd be happy to remove/change the bit about 'common exit codes in the table below' though if that would sit better?

pda:
I finally loaded this back into my brain and documented the signal-related exit codes.

It seemed difficult to include them in the main status code table, so I decided pulling them into their own separate table made sense. It's worth noting that this 128+N stuff isn't defined by us, it's done by shells like bash/zsh etc. And it's not worth trying to exhaustively document all the signals, as many are irrelevant and some may vary between flavours of POSIX/Unix systems.

Iterations on this welcome:

image

harrietgrace:
Rad, thanks so much @pda! 🎉

I moved the info about the signal numbers into a note and rephrased it so that it was a bit clearer (I hope). Does it still work, do you think?

image

pda:
Nice :ok_hand: 

Although I don't think this sentence will be useful to the audience, because they won't have shell access to the job that Buildkite is reporting on:

> You can get this value from the special variable $?.

Perhaps we could remove that sentence and then :shipit: ?

harrietgrace:
image
😂

PR #581, opened 2019-11-14
Add config for idle timeout

harrietgrace:
As noted here https://github.com/buildkite/agent/pull/1009#issuecomment-553642001, we're missing the non-deprecated config option for job timeout! 

This PR adds the agent config option `disconnect-after-idle-timeout`. As neither of the disconnect config options were listed in the env vars table, I've added them to that too.


harrietgrace:
@lox does it wait for the number of agents specified with `spawn`, or just all agents? Say you had `spawn = 5`, then were also running 3 other agents from your local machine, would the idle timeout wait for just the 5 spawned together, or all 8?

keithpitt:
@harrietgrace it waits for all agents started together with `spawn`. Any other individual `buildkite-agent start` calls just do their own thing.

So with `buildkite-agent start --spawn 5`, once a job finishes across any of them, it waits 5 seconds. If there's no new work in that period, then it disconnects.

harrietgrace:
Thanks @keithpitt! 🙏🏻 Does this addition make sense about the spawn stuff?

image

keithpitt:
@harrietgrace 💯

PR #582, opened 2019-11-14
Update GCP docs to use Kubernetes engine

### LINKED ISSUE #534: Kubernetes + buildkite-agent - mounting docker binary as volume doesn't seem to work
The documentation at https://buildkite.com/docs/agent/v3/gcloud says to add the docker binary as a volume, but I've found that this doesn't work (it causes a segfault when trying to run docker in the pod). I could only get my deployment to work when I *removed* the `docker-binary` volume.

I also found that [buildkite/charts](https://github.com/buildkite/charts/blob/36f0a1fbfdc68347ce71e3ead5f54eb99d1bfb19/stable/agent/templates/deployment.yaml#L72) also doesn't mount the binary; only the socket.

[One of the top results on google](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/) also recommends against bind-mounting the binary, since it might be unstable.

Are the docs out of date?

harrietgrace:
It looks like our GCP docs are out of date: https://github.com/buildkite/docs/issues/534.

I went and tested out our instructions in the Google Kubernetes Engine and it was mostly still the same. I've updated the screenshots and a bit of the k8s deploy code. As mentioned in the above issue, mounting the docker binary as a volume causes issues, so I've removed that. Some of the syntax changed around the secret create/get, and some of the URLs they provide updated. 

@sj26 could I get you to run your eyes over this to check it looks alright? 🙏🏻

Closes #534


pages/agent/v3/gcloud.md.erb
```diff
@@ -44,18 +44,18 @@ Double check it’s running by taking a look at the cluster info:
...
 $ kubectl cluster-info
 Kubernetes master is running at https://123.45.67.89
-GLBCDefaultBackend is running at https://123.45.67.89/api/v1/proxy/namespaces/kube-system/services/default-http-backend
-Heapster is running at https://123.45.67.89/api/v1/proxy/namespaces/kube-system/services/heapster
-KubeDNS is running at https://123.45.67.89/api/v1/proxy/namespaces/kube-system/services/kube-dns
-kubernetes-dashboard is running at https://123.45.67.89/api/v1/proxy/namespaces/kube-system/services/kubernetes-dashboard
+GLBCDefaultBackend is running at https://123.45.67.89/api/v1/namespaces/kube-system/services/default-http-backend:http/proxy
+Heapster is running at https://123.45.67.89/api/v1/namespaces/kube-system/services/heapster/proxy
+KubeDNS is running at https://123.45.67.89/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy
+Metrics-server is running at https://123.45.67.89/api/v1/namespaces/kube-system/services/https:metrics-server:/proxy
```
plasticine: These will vary quite a bit just based on how the user configured their cluster when they created it — dunno if that’s a problem, but it’s pretty likely they could end up with some different stuff here...

harrietgrace: Ah right! I don't think that's a problem. Hopefully people will let us know if they're finding it confusing.

sj26:
Sorry for the delay @harrietgrace, but this does look good!

PR #586, opened 2019-11-29
fix: key value

joscha:
Got this in a pipeline upload:
```
422 One of the steps you provided was invalid: Step keys can't have the same pattern as a UUID (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```


harrietgrace:
Nice addition, thanks @joscha! 🙇🏻‍♀️

PR #587, opened 2019-11-29
document the new build.pull_request.draft variable

yob:
This variable was recently added to the conditional system. It's only available on pull requests from github.


harrietgrace:
Eep, so sorry for the delay @yob! 😬 This looks great 🙌🏻 If you could add in an example of it to the Examples section at the end, that would be super. Then I think we're ready to ship!

yob:
All good, this wasn't urgent!

I've added an example

PR #590, opened 2019-12-09
bump ruby to 2.6.5

yob:
This is the latest 2.6 release, and the version we're running in prod


harrietgrace:
Cool beans! Thanks @yob! I don't really know much about the code side of the docs app, so I'm not the best reviewer for this sorry 😅 My only question is: should the build be passing? 🤓

yob:
It totally should

yob:
fixed! Ok to merge @keithpitt? @sj26?

PR #596, opened 2019-12-26
Change on-premise to on-premises

call:
Hope this helps. 👍 

https://en.wikipedia.org/wiki/On-premises_software


harrietgrace:
Fantastic, great catch @call! Thanks for taking the time to fix/PR 😊🙏🏻

PR #597, opened 2019-12-30
Add info about passing block step values to other steps

### LINKED ISSUE #595: Trigger build with values coming from metadata
Hello,

I have a build with a block step, which has a multi select prompt, that unlocks a [trigger step](https://buildkite.com/docs/pipelines/trigger-step).

I want to know how can I send in the trigger the value selected in the multi-select prompt?

Thank you,

harrietgrace:
As mentioned in #595, we don't provide any details about how to pass the values from a block step to anywhere else. It's come up a few times in support, so it should be added to the docs. Even though passing meta-data to a trigger step doesn't have a tidy solution at the moment, people still need to know how to do the workaround.

This PR adds two sections: one about retrieving meta-data in a script, and one about doing the dynamic trigger step upload dance. 

Closes #595


keithpitt:
I just did a sneaky review of the words, looks great. I reckon we can totes ship this!

harrietgrace:
Sweet, thanks @keithpitt! 🙏🏻

PR #598, opened 2020-01-03
Add warning for conditionals with custom build.env values

jayco:
Adding a warning around custom env vars as I have had two users caught out by this.

A bit of background:
Currently there are two flows when the build is created and they both operate differently in how they handle the processing of env var precedence. The new beta YAML pipeline steps supports this but the original pipeline upload does not (possibly breaking existing pipelines if we changed the precedence).

- Not sure if we want to link to the change log here as I did not see any other documentation around this.
- I did try and initially add this as **Note** to the build.env in the table column but it was pretty large

### DIFF
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -235,8 +235,8 @@ The below variables are supported by the `if` attribute:
 </table>
 
 <div class="Docs__troubleshooting-note">
-<h1>Using build.env with custom environment variables</h1>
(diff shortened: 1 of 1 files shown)


harrietgrace:
Epic! Thanks @jayco! 🙌🏻🙇🏻‍♀️ I swapped some words around, but it still says pretty much the same thing. I also added some code formatting for `build.env()`.

It sucks that we have to link to a beta feature changelog for that ☹️ At least we'll be able to change that once it's out of beta 😄

jayco:
Awesome @harrietgrace 🙌, looks much better now!

PR #603, opened 2020-01-13
Fix typo in notification docs

jayco:
Couple of quick fixes here:
- Fixing a typo here from our example that uses an unknown value status rather than build state. 
- This also fixes an error about the passed value not being a variable.


eleanorakh:
Oopsie!! Thanks for picking up @jayco !!!

PR #609, opened 2020-01-16
Fix wrong link

andor-pierdelacabeza:
Link named "input step" was linking "block step" (that is, this same page)


harrietgrace:
Fantastic, thanks @andor-pierdelacabeza! Thanks for taking the time to PR as well 😊🙏🏻

PR #611, opened 2020-01-17
agent diagram: remove gap between "AGEN" and "T" in AGENT.

pda:
Before; note the gap in `AGEN T` (Firefox 72, Linux): 

image

After:

image

Here's a screenshot of the diff in `--color-words` mode, since it's quite unreadable otherwise:

image


harrietgrace:
Thanks @pda 🙇🏻‍♀️

toolmantim:
Good catch!

PR #613, opened 2020-01-17
Add EventBridge filtering and logging information

toolmantim:
This replaces the EventBridge logging example with a new "Logging" section that describes how to route events directly to AWS Cloudwatch Logs, rather than using a Lambda, and use Insights for querying the logs.

It also adds a "Filtering" section showing how to use the Event Pattern filter rules to limit which events are processed. We show you how to filter by pipeline slug, which is particularly handy because unlike webhooks we don't let you specify which pipelines trigger EventBridge events via your Buildkite settings.

- [x] Stub out new sections
- [x] Test new stuff actually works
- [x] Add screenshots


toolmantim:
@jayco up for pairing with me on giving these rules a test, and getting some screenshots for them?

jayco:
@toolmantim Absolutely, Let's do it!

toolmantim:
Can we merge this one @harrietgrace?? So keen!

toolmantim:
👏🏼 Thanks @harrietgrace!

PR #616, opened 2020-01-20
Fix concurrency parameter in deployments.md.erb

andor-pierdelacabeza:
Concurrency parameter is called "concurrency_limit" instead of just "concurrency" as in the rest of the docs.


harrietgrace:
Great catch! Thanks a bunch 😊

PR #617, opened 2020-01-21
update to add in conditional service setup

jayco:
I added some additional documentation around setting up slack for conditional notifications.

The slack integrations page:
Screen Shot 2020-01-21 at 8 51 39 pm

The pipelines notifications page:
Screen Shot 2020-01-21 at 8 53 25 pm

I was not entirely sure about where to put this - let me know if you want it moved around or updated.


jayco:
@harrietgrace yeah I think I should reword it something like:

> To ensure the default behaviour of the notification services is overridden when notifications are to be triggered by a conditional evaluation from a pipeline [notify](/docs/pipelines/notifications#conditional-notifications) attribute, the Slack service needs to be configured with 'Only Some Pipelines' selected and the pipeline using the 'notify' attribute un-checked

If the pipeline remains checked, then the channel will receive notifications regardless of the conditional value. 

Does that sound ok?

jayco:
I have updated that text to reflect the comment above @harrietgrace

harrietgrace:
Sorry it took me so long to get back to you about this @jayco! 😬It totally slipped my mind 😔

I split it up into a little list and added a code sample. There was so much good stuff in that one sentence that I thought it deserved more airtime 😄

PR #626, opened 2020-01-27
Meta data limits

sj26:
Document the newly added meta-data value limits and intended deprecation, and add some basic guidelines about what to use meta-data for.

### DIFF
--- a/pages/agent/v3/cli_meta_data.md.erb
+++ b/pages/agent/v3/cli_meta_data.md.erb
@@ -14,6 +14,8 @@ Use this command in your build scripts to save string data in the Buildkite meta
 <%= render 'agent/v3/help/meta_data_set.txt' %>
 ```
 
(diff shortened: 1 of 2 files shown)


pages/agent/v3/cli_meta_data.md.erb
```diff
@@ -14,6 +14,8 @@ Use this command in your build scripts to save string data in the Buildkite meta
 <%= render 'agent/v3/help/meta_data_set.txt' %>
 ```
 
+Meta-data values are restricted to a maximum of 100 kilobytes. Keys and values larger than 1 kilobyte are deprecated. Please use [artifacts](https://buildkite.com/docs/agent/v3/cli-artifact) for large data which needs to uploaded and downloaded.
```
harrietgrace: Are we planning to make the deprecation the standard at some point? If so, do we have an idea when?

sj26: We haven't decided on an official deprecation yet, so I've changed the wording to "discouraged".

PR #630, opened 2020-02-02
Update osx.md.erb

bazscott:
Comment is inconsistent with the command that is being run, which confused me when installing as I copied the comment path to my terminal to check permissions and got very confused :) So better if they match up I think.


harrietgrace:
Great idea! Thanks for taking the time to PR @bazscott 😊🙏🏻

PR #632, opened 2020-02-05
Use a consistent filename for the k8s manifest template

yob:
Earlier in the tutorial, we created the template at `k8s/deployment.yml` rather than `k8s/app.yml`.

Thanks to Madhuri Yechuri for noticing this.


myechuri:
Thanks for the quick fix!

PR #633, opened 2020-02-10
Use GIT_SSH_COMMAND for debugging ssh

toolmantim:
Running a build with an environment variable set is a much easier way of debugging SSH than having to edit a `~/.ssh/config` file on the agent machine.


yob:
nice one :+1:

PR #642, opened 2020-02-21
Add member permissions page

harrietgrace:
Finally adding some docs for the new member permissions gear 🎉


harrietgrace:
@toolmantim I've mushed the new content into the old permissions page, let me know what you think! I'll need to check the new stuff is correct before this goes out too 🤓

harrietgrace:
@toolmantim you had some questions and ideas about ways we could make this page better in our call the other day - could you throw those in a comment on here? 😊🙏🏻

pages/pipelines/permissions.md.erb
```diff
@@ -1,38 +1,39 @@
+# Controlling user permissions
 
-# Controlling permissions with Teams
-
-Enabling Teams for your organizations gives you fine grained control over the permissions for each of your pipelines.
+Pipeline permissions can be centrally managed by enabling Teams. Enterprise customers have the additional option to set fine-grained user permissions for their organization with the member Permissions page.
```
plaindocs, line 3: Would it make sense to add a line here about how it looks if you have neither teams or member permissions?

harrietgrace: Good idea!

harrietgrace:
@plaindocs I just added in a specific bit about finding the create permission without teams enabled, as per your suggestion above from ages ago 😄

Could you please have a squiz of the permissions page and let me know what you think? Happy to have feedbacks on anything from copyedits to where this info should live.

pages/pipelines/permissions.md.erb
```diff
@@ -50,16 +51,32 @@ if [[ ":$BUILDKITE_BUILD_CREATOR_TEAMS:" != *":ops:"* ]]; then
...
 
-### Can I delete the “Everyone” team?
+#### Can I delete the “Everyone” team?
 Yes you can delete or edit the “Everyone” team. To ensure uninterrupted access to pipelines we recommend creating new teams before deleting the “Everyone” team.
 
-### Once enabled, can I disable Teams?
+#### Once enabled, can I disable Teams?
 You can disable teams by deleting all your teams, and then selecting “Disable Teams”.
+
+## Member Permissions
```
plaindocs, line 68:
```suggestion
## Member Permissions without teams
```

harrietgrace: I'm not sure about this one, because in the first paragraph I've said "These permissions can be used both with or without Teams enabled." 😅

PR #643, opened 2020-02-25
Correct usage string for agent artifact upload

jamesmharvey:
`buildkite-agent artifact upload` doesn't actually accept a `<destination>` argument. This is already indicated in the docs:

> Buildkite will store the file at log/test.log. If you want it to be stored as test.log without the full path, then you'll need to change into the file's directory before running your upload command:

https://buildkite.com/docs/agent/v3/cli-artifact#uploading-artifacts

This PR removes the `<destination>` argument from the `buildkite-agent artifact upload` usage string.

### DIFF
--- a/pages/agent/v3/help/_artifact_upload.txt
+++ b/pages/agent/v3/help/_artifact_upload.txt
@@ -1,6 +1,6 @@
 Usage:
 
-   buildkite-agent artifact upload <pattern> <destination> [arguments...]
+   buildkite-agent artifact upload <pattern> [arguments...]
 
 Description:
 


pda:
I believe `<destination>` is actually a thing, but it's not for specifying the destination filename/path. Instead, it's for specifying a custom artifact bucket/repository to use instead of having Buildkite host the artifacts for you.

The implementation is here:
https://github.com/buildkite/agent/blob/b957620a/clicommand/artifact_upload.go

It defaults to `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` from environment (which is normally not set) and expects values like:

* `s3://name-of-your-s3-bucket/$BUILDKITE_JOB_ID`
* `gs://name-of-your-gs-bucket/$BUILDKITE_JOB_ID`
* `rt://name-of-your-artifactory-repo/$BUILDKITE_JOB_ID`

The Examples in the documentation demonstrate that:

```
   You can also upload directly to Amazon S3 if you'd like to host your own artifacts:

   $ export BUILDKITE_S3_ACCESS_KEY_ID=xxx
   $ export BUILDKITE_S3_SECRET_ACCESS_KEY=yyy
   $ export BUILDKITE_S3_DEFAULT_REGION=eu-central-1 # default is us-east-1
   $ export BUILDKITE_S3_ACL=private # default is public-read
   $ buildkite-agent artifact upload "log/**/*.log" s3://name-of-your-s3-bucket/$BUILDKITE_JOB_ID

   Or upload directly to Google Cloud Storage:

   $ export BUILDKITE_GS_ACL=private
   $ buildkite-agent artifact upload "log/**/*.log" gs://name-of-your-gs-bucket/$BUILDKITE_JOB_ID

   Or upload directly to Artifactory:

   $ export BUILDKITE_ARTIFACTORY_URL=http://my-artifactory-instance.com/artifactory
   $ export BUILDKITE_ARTIFACTORY_USER=carol-danvers
   $ export BUILDKITE_ARTIFACTORY_PASSWORD=xxx
   $ buildkite-agent artifact upload "log/**/*.log" rt://name-of-your-artifactory-repo/$BUILDKITE_JOB_ID
```

Thanks for the pull request, but I believe it is as it should be.

pda:
(perhaps we could clarify the purpose of `destination` in the description of the command, rather than only in the examples)

harrietgrace:
@pda yep, definitely worth updating the --help description. I think that lives in the agent?

PR #644, opened 2020-02-26
nokogiri 1.10.8; fix CVE-2020-7595

pda:
> xmlStringLenDecodeEntities in parser.c in libxml2 2.9.10 has an infinite
> loop in a certain end-of-file situation.  The Nokogiri RubyGem has
> patched it's vendored copy of libxml2 in order to prevent this issue
> from affecting nokogiri.

-- https://github.com/advisories/GHSA-7553-jr98-vx47

I haven't tested this locally, but I'd be surprised if a point release of nokogiri broke anything.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -63,7 +63,7 @@ GEM
     mini_mime (1.0.1)
     mini_portile2 (2.4.0)
     minitest (5.11.3)
-    nokogiri (1.10.5)
+    nokogiri (1.10.8)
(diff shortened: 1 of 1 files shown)


harrietgrace:
@pda It all looks alright in the Heroku preview bizzo! Is there anything specific you think we need to check? Otherwise happy to merge 🌈

yob:
If it's good on heroku, it should be fine to merge!

PR #646, opened 2020-02-28
🔍 Consistent "Agent, Plugin" ordering

drnic:
I feel the need to apologise for this anally retentive fix.


keithpitt:
Good find :)

toolmantim:
I have a strong suspicion that the original docs might been accurate:
https://github.com/buildkite/agent/blob/master/bootstrap/bootstrap.go#L889

harrietgrace:
Sorry @drnic, I'm going to have to revert this one! That column has specific ordering, so it was the original way around on purpose. Thanks for the PR though, we always love contributions 😄

keithpitt:
Oh, doh! My bad!

drnic:
Thanks. I learned a thing.

PR #658, opened 2020-03-12
Use official memcache image

MetaJuanitoFatas:
Spotted from https://twitter.com/borja_burgos/status/1237800221347807233 🔎


toolmantim:
Hah! Good catch. Thanks for the PR :)

PR #661, opened 2020-03-13
Link deprecated tags to recommend tags

plaindocs:


plaindocs:
Does the linkchecker not like anchor links?

toolmantim:
Fixed the spec for ya!

PR #664, opened 2020-03-16
Fix TOC rendering for nested headings

toolmantim:
This switches the way we do the table of contents to use Nokogiri after the markdown render happens, rather than an ERB method.

This removes a bunch of weird edge cases, such as ones that cropped up in https://github.com/buildkite/docs/pull/442, where nested headings inside nested markdown blocks were getting picked up.

It also moves us closer to being able to use CommonMark/HTML::Pipeline in the future too.

`{:toc}` and `[TOC]` seem to the be common ways of doing placeholders for post-processing, so I just picked one. Happy to use something different.


toolmantim:
Thanks @sj26! Not much of any of this has specs, but no reason not to add some! Thanks for the poke. I'll do that, and then it's good to go?

sj26:
@toolmantim yeah, if we can see the spec'd markup produced I've be a hearty :+1: — looks great!

PR #667, opened 2020-03-24
Fix typo in "Deploying to Kubernetes"

agis:


harrietgrace:
Good catch @agis! Thanks for taking the time to PR 😄🙏🏻

PR #672, opened 2020-03-26
Configure the app to serve pages at /docs/* and /docs-fargate/*

yob:
We're preparing to serve https://buildkite.com/docs with this app deployed to fargate.

During the preparation we'd like to serve https://buildkite.com/docs from the current location, and https://buildkite.com/docs-fargate from this app on fargate.

Once we're happy that the rails app on fargate is working correctly, we can change the ALB to route https://buildkite.com/docs to it.

This PR attempts to make the app capable of working under both route prefixes, so there's no code changes required to start serving production traffic (and we can then remove support for `/docs-fargate`) at our leisure.

Note that assets will be loaded from `/docs-fargate/assets/*`. We'll need to revert the change in `config/initializers/assets.rb` before we remove `/docs-fargate` from the ALB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,8 @@ COPY . /app
 # Compile sprockets
 RUN if [ "$RAILS_ENV" = "production" ]; then \
       echo "--- :sprockets: Precompiling assets" \
-      && RAILS_ENV=production RAILS_GROUPS=assets bundle exec rake assets:precompile; \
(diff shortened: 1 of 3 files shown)


yob:
You can see the impact of this change in auto-deployed version (if you URL hack). Both these paths work:

1. https://docs-review-multi-route-9ff9jd.herokuapp.com/docs/tutorials/getting-started
2. https://docs-review-multi-route-9ff9jd.herokuapp.com/docs-fargate/tutorials/getting-started

PR #674, opened 2020-03-26
Some more URL-related tweaks to make /docs-fargate work nicely

yob:
While we're in the process of migrating docs to be hosted on fargate, we're routing https://buildkite.com/docs-fargate/* to the fargate containers.

This set of tweaks makes it possible to click around the site at https://buildkite.com/docs-fargate/* and stay on the  `/docs-fargate` prefix. Hopefully this makes it easier to test, with much less URL hacking.

@matthewd has pointed out that it might've been easier to leave the URLs as `/docs/*` and use an ALB rule that routes the *host* `pre.buildkite.com` to the new fargate service. he's right, that would've been easier. We've sunk the time into this approach now and I think it's done. Oops.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,13 +4,13 @@ module ApplicationHelper
   end
 
   def sidebar_link_to(name, path, options = {})
-    url = "/docs/#{path}"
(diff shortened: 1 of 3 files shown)


yob:
I've clicked around the test deployment (on heroku?) and it seems to work well. Not in prod yet, so I'll go ahead and merge.

PR #677, opened 2020-03-26
Adds support for code block filenames

### LINKED ISSUE #666: Filename for codeblock syntax
By default Markdown doesn't have a way a way to specify the filename for a codeblock.

Kramdown allows for specifying arbitrary attributes using `{: attribute="value"}`.

````md

```
READ ME
```
{: codeblock-file="README.md"}
````

And there is some discussion about adding something similar to CommonMark if we're intending to move towards that https://talk.commonmark.org/t/consistent-attribute-syntax/272

Seems good enough unless there is a good reason to use something else. 

Then it is a case of using that info to show the info in a useful manner.  `Prism.js` has some chat here, https://github.com/PrismJS/prism/issues/910 and here https://prismjs.com/plugins/show-language/ .  but I've not really found anything similar for Pygments.

```
filename
    A string used to generate a filename when rendering <pre> blocks, for example if displaying source code.
```

Seems relevant. https://pygments.org/docs/formatters/

And here is an example of how it can look (with a darker code block) 
image

@toolmantim does this look doable?
@harrietgrace any thoughts?

### PR DESCRIPTION
**Pull request by toolmantim**
Adds support for code blocks with filenames.

For example:

~~~markdown
Here's a example of a `docker-compose.yml` file for a Ruby on Rails application that depends on Postgres, Redis and Memcache:

```yml
version: '2'
services:
  db:
    image: postgres
  redis:
    image: redis
  memcache:
    image: memcached
  app:
    build: .
    working_dir: /app
    volumes:
```
{: codeblock-file="docker-compose.yml"}
~~~

image

Closes #666

### DIFF
--- a/app/assets/stylesheets/public/utils/_text_content.scss
+++ b/app/assets/stylesheets/public/utils/_text_content.scss
@@ -52,4 +52,21 @@
   .dark-gray { color: #8E8E8E; }
   .border { border: 1px solid; }
   .border-gray { border-color: #ddd; }
+
+  .highlight-figure {
+    figcaption {
+      border: 1px solid #ddd;
+      padding: .8em 1em;
+      border-top-left-radius: 5px;
+      border-top-right-radius: 5px;
+      border-bottom: none;
+      font-size: 14px;
+      color: #777;
+    }
+    pre {
+      margin-top: 0;
+      border-top-left-radius: 0;
+      border-top-right-radius: 0;
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


toolmantim:
Would be nice with a little file icon!

harrietgrace:
So cute!

plaindocs:
SWEET! Thanks a ton @toolmantim 

I reckon let's merge this in, I'll do a quick pass and add some filenames here and there. And if we want to revisit and add an icon, even better!

toolmantim:
Okay cool! I'll port it over to the backend version too before merging, so we don't block master from getting updated.

toolmantim:
You can go wild adding these everywhere now @plaindocs

PR #678, opened 2020-03-26
Add a deploy step if the build is green

yob:
A step towards continuously deploying the docs site to fargate.

I assume this might take a few cycles to get right, but hopefully this is a good start. We're still feeling out way with fargate, so I've based this on the way we deploy site.

There seems to be rough consensus that we want to store the basic task definition inside the code repo (or at least, the key parameters that will vary over time. Like CPU, memory, and image).

It's not yet clear what our ideal tooling around that will look like. I assume we'll eventually want something more capable than a shell script. However, until we settle on what that future looks like I've emulated the way we deploy site to ECS. At least when we do decide on more capable tooling, we can migrate docs and site in the same way.


toolmantim:
This is awesome!

I was hoping we'd use https://github.com/buildkite-plugins/ecs-deploy-buildkite-plugin and then backport site to use what docs does 😅 That plugin was actually extracted out of the site shell code, but then we never switched to actually using it. But I don't mind either way TBH!

yob:
oh nice, I didn't realise that plugin existed.

Using plugins for deploy steps makes me a bit nervous because it masks the steps slightly harder to run manually during an incident.

Still, I'm still in a total exploration mode so I'll give it a go.

yob:
There's no way to know if this will work until we try it (and there's no impact on production), so I'm going hit the green button and watch the result.

PR #684, opened 2020-03-31
redirect /docs-fargate to somewhere that renders a page

yob:
At the moment we're testing this app in production at https://buildkite.com/docs-fargate, and it's helpful to have that URL work (just like https://buildkite.com/docs).

Once this app is moved to service requests for https://buildkite.com/docs, we can remove this.


yob:
Hopefully uncontroversial

PR #685, opened 2020-04-01
Close bash command quotation mark

tommeier:
* Remove the empty spaces.
  * Close the bash command with missing quote mark (can see in syntax highlighting on the doc page it isn't closed too)

image


harrietgrace:
Great catch @tommeier! Thanks for taking the time to PR 😊🙏🏻

PR #686, opened 2020-04-01
Update schedules readme

HzrdIRL:
The documentation around `%` and `+` was causing confusion among some of our developers as it was not clear from what point the frequency is calculated (monthly, yearly, etc). After having a discussion on slack with Matthew Draper, and discovering some typos, it felt best to update the docs.

### DIFF
--- a/pages/pipelines/scheduled_builds.md.erb
+++ b/pages/pipelines/scheduled_builds.md.erb
@@ -62,11 +62,11 @@ Using `L` or `last` in the _day of month_ field represents the last day. For exa
 
 ##### Modulo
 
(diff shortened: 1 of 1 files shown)


HzrdIRL:
> Or maybe @harrietgrace has some more ideas?

Happy to continue the discussion or provide examples.

harrietgrace:
Thanks for taking the time to PR @HzrdIRL! 🙇🏻‍♀️ 

I added in an extra example and did a few little formatting/copy tweaks. I ended up taking out the paragraph about how the modulo is calculated for a better balance of clarity and completeness for the majority of users. However, If you feel strongly about having an example/explanation in there, we can definitely revisit it. 

Thanks again for contributing, we really appreciate it 💚

PR #691, opened 2020-04-01
Set matomo ENV vars in production

yob:
These are used to render the matomo fragment on every page

### DIFF
--- a/deploy/task-definition-template.json
+++ b/deploy/task-definition-template.json
@@ -19,6 +19,18 @@
         {
           "name": "RAILS_SERVE_STATIC_FILES",
           "value": "true"
+        },
+        {
+          "name": "MATOMO_URL",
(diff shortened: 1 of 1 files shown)


deploy/task-definition-template.json
```diff
@@ -19,6 +19,22 @@
...
           "name": "RAILS_SERVE_STATIC_FILES",
           "value": "true"
+        },
+        {
+          "name": "MATOMO_URL",
+          "value": "thttps://analytics.buildkite.com"
+        },
+        {
+          "name": "MATOMO_SITE_ID",
+          "value": "1"
```
yob, line 29: @toolmantim I cargo-culted these values from elsewhere. Is this site ID appropriate for docs?

yob:
Had a good chat with @toolmantim out of band and we realised the `MATOMO_SITE_ID` env var wasn't needed

PR #692, opened 2020-04-02
assume assets are served under /docs

yob:
While testing this app on fargate we were serving assets from `/docs-fargate`. However, now that this app is serving user traffic under `/docs`, we can remove the temporary asset path

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ COPY . /app
 RUN if [ "$RAILS_ENV" = "production" ]; then \
       echo "--- :sprockets: Precompiling assets" \
       && RAILS_ENV=production RAILS_GROUPS=assets bundle exec rake assets:precompile \
-      && cp -r /app/public/docs-fargate/assets /app/public/assets; \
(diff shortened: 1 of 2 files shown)


yob:
To test this, I built a copy of the prod docker image and ran it:

```
$ docker build . -t foo
....
some stuff
...
Successfully built b92b98ca5009                                                                                                                                                                                                                                                             Successfully tagged foo:latest

$ docker run -e SECRET_KEY_BASE=123 -it -p 3000:3000 foo
Puma starting in single mode...
* Version 4.3.3 (ruby 2.6.6-p146), codename: Mysterious Traveller
* Min threads: 20, max threads: 20
* Environment: production
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
```

And then loaded http://127.0.0.1:3000/docs in my browser and confirmed the styles loaded.

I can see in source that the assets are loading from http://127.0.0.1:3000/docs/assets instead of http://127.0.0.1:3000/docs-fargate/assets

PR #694, opened 2020-04-02
Update include_retried_jobs description

### LINKED ISSUE #645: clarify include_retried_jobs semantics
The page https://buildkite.com/docs/apis/rest-api/builds mentions include_retried_jobs several times:

```
include_retried_jobs	Include all retried jobs in each build’s jobs list.
                        Example: ?include_retried_jobs=true
```
Since the name of the parameter is `include_retried_jobs` it makes it sound as if, by default, you will receive only the results for the first execution of the step, since you have to supply the parameter to get the retries.  But in fact, it seems that by default you receive the information for the most recent execution of the step, and you need to supply the parameter to receive the information for the previous, superseded executions of the step, including the first execution, which wasn't a retry.  A better name for this parameter might be `include_superseded_jobs`.  But it's part of your API, so it's probably not a good idea to change it.  But you could document the behavior more explicitly.  For instance, the description could be:

Include all job executions in each build's jobs list. Without this parameter, you'll see only the most recent jobs for each step.

Or something to that effect.

harrietgrace:
As pointed out in https://github.com/buildkite/docs/issues/645, our description for the builds API param `include_retried_jobs` isn't super clear. 

The PR updates the description to @dws-uber's words suggestion 🙌🏻 Thanks @dws-uber! 🙇🏻‍♀️

Closes #645

### DIFF
--- a/pages/apis/rest_api/_builds_list_query_strings.md.erb
+++ b/pages/apis/rest_api/_builds_list_query_strings.md.erb
@@ -42,7 +42,7 @@
   </tr>
   <tr>
     <th><code>include_retried_jobs</code></th>
-    <td>Include all retried jobs in each build’s jobs list.<p class="Docs__api-param-eg">
+    <td>Include all retried job executions in each build's jobs list. Without this parameter, you'll see only the most recently run job for each step.<p class="Docs__api-param-eg">
       <em>Example:</em> <code>?include_retried_jobs=true</code></p>
     </td>
   </tr>
--- a/pages/apis/rest_api/_builds_list_query_strings.md.erb
+++ b/pages/apis/rest_api/_builds_list_query_strings.md.erb
@@ -42,7 +42,7 @@
   </tr>
   <tr>
     <th><code>include_retried_jobs</code></th>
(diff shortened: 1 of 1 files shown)


harrietgrace:
@toolmantim how much do you care about proper apostrophes? I'm stomping one here, but can put it back if you have feels!

pages/apis/rest_api/_builds_list_query_strings.md.erb
```diff
@@ -42,7 +42,7 @@
   </tr>
   <tr>
     <th><code>include_retried_jobs</code></th>
-    <td>Include all retried jobs in each build’s jobs list.<p class="Docs__api-param-eg">
+    <td>Include all retried job executions in each build's jobs list. Without this parameter, you'll see only the most recently run job for each step.<p class="Docs__api-param-eg">
```
toolmantim, line 45:
```suggestion
    <td>Include all retried job executions in each build’s jobs list. Without this parameter, you'll see only the most recently run job for each step.<p class="Docs__api-param-eg">
```

PR #698, opened 2020-04-02
Add some filnames!

plaindocs:
To make use of #666

### DIFF
--- a/pages/pipelines/wait_step.md.erb
+++ b/pages/pipelines/wait_step.md.erb
@@ -8,13 +8,14 @@ A *wait* step waits for all previous steps to have successfully completed before
 
 <%= image "wait-step.png", width: 1070/2, height: 508/2, alt: "Screenshot of a basic wait step" %>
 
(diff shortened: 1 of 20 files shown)


pages/pipelines/wait_step.md.erb
```diff
@@ -51,19 +52,20 @@ _Optional attributes:_
...
 In the example below, if `command.sh` succeeds, both of the following command steps will be run. If `command.sh` fails, only the first will be run, and the build will then be marked as failed.
 
 ```yml
-- command: "command.sh"
-- wait: ~
-  continue_on_failure: true
-- command: "echo This runs regardless of the success or failure"
-- wait
-- command: "echo The command passed"
+steps:
```
plaindocs, line 62: This is the non-whitespace + non-filename change.

toolmantim:
Seems legit! Is there anywhere in the paragraph before the code blocks where we no longer have to mention the filename? I remember often trying to craft the words in such a way as to spell out what the file was? For example, we can remove all the `touch <file>` from the plugin tutorial (though the `chmod +x` is still important for hooks)

plaindocs:
Good call on the `touch` commands. The few "in paragraph" entries I thought would be fine to just leave.

PR #701, opened 2020-04-05
Fix link to buildkite CLI

supaspoida:
The link to the Buildkite CLI on the plugin tools page was pointing to the shellcheck plugin.


harrietgrace:
Great catch @supaspoida, much appreciated! Thanks for taking the time to PR 😄🙏🏻

Issue #705, opened 2020-04-15
Buildkite artifact upload/download documentation and glob patterns are confusing


mboudreau:
This page in particular: https://buildkite.com/docs/pipelines/artifacts

Let's start with upload: it uses a glob pattern, which is great as it makes it easier to upload a bunch of stuff, however the documentation is all over the place.  In the text it says you can upload everything under a directory using `some-directory/**/*` which is correct and it does work, however it doesn't specify that it also stores the whole directory path associated with it. ie. the directory prefix `some-directory`.  There's also the problem of the supporting image showing 2 glob patterns that include each other: https://buildkite.com/docs/assets/pipelines/artifacts/artifact_pattern-34f72b531b2f55bd1c9018b9998a991fdba52ff35c1720a9b375f218ae5af366.png

This image confused me as it was hinting that the glob pattern didn't work as expected, where the user would need to specify a glob for the first level of the directory and another for all other levels. ie. `some-directory/*;some-directory/**/*` as shown in the image.  

This document also shows a lot of usage of the CLI but not the YML.  Does the `;` separator work in yml? This is not mentioned.

It also doesn't explain the mechanism as to how it stores it, where it stores it and what's the idea behind the download, which leads to the next problem; the download completely departs from the example given in the upload!

The download changes to downloading a zipped file, not a directory.  At first this made me thought that this was because the artifact uploaded was converted to a zip, but that didn't seem to make sense.  Then it was kinda alluding that the glob pattern was the 'id' of the thing you wanted to download, so I added a command of `buildkite-agent artifact download 'some-directory/**/*' .` to download my artifacts but just to learn after digging through code and documentation that the glob pattern isn't an id, it's used as a glob for the remote store, and then that the glob is not the same between upload and download!

Also, why isn't the artifact download a part of the yml spec?  Seems weird.

In the end, I lost at least a day of work trying to debug this issue because of confusing documentation and lack of coherency in the examples.  Please explain this feature is a clearer way,  hopefully with some consistent examples and some images to show what happens to the files and how they're stored/queried.  And please fix the globbing between the two, there's no reason not to have both working the same way.

plaindocs:
Hey @mboudreau - thanks for the helpful report.

I'll dig a little deeper into this and get some better examples up. I'll see if I can someone to address the different globbing styles also - but no promises on that front.

PR #707, opened 2020-04-17
Make the artifact docs more clear

plaindocs:
To address https://github.com/buildkite/docs/issues/705

- [x] link to H3 examples (waiting on implementation of #708 )
- [x] downloads
- [x] double check `/docs/agent/v3/cli-artifact` 

The `v3/cli` docs seem ok as is, and this is an improvement on what we have in the pipeline doc.


pages/pipelines/artifacts.md.erb
```diff
@@ -6,7 +6,25 @@ In this guide we’ll walk through using the Buildkite Agent’s [artifact suppo
 
 ## Uploading Artifacts
 
-To automatically upload artifacts created by a build step, set the `artifact_paths` attribute on your command step. In this example, any artifacts that are created in the `logs` or `coverage` directories, as well as any of their subdirectories, will be uploaded automatically:
+How you define your build pipeline determines how you automatically upload artifacts created by build steps:
```
harrietgrace, line 9:
```suggestion
How you define your build pipeline determines how you configure the automatic upload for artifacts created by build steps:
```

pages/pipelines/artifacts.md.erb
```diff
@@ -6,7 +6,25 @@ In this guide we’ll walk through using the Buildkite Agent’s [artifact suppo
...
+
+* pipelines defined in the Buildkite UI use the ["Automatic Artifact Uploading"](#uploading-artifacts-automatic-artifact-uploading) option.
```
harrietgrace, line 11: Could we pull out 'Pipelines defined in' to the top of the list, then have each dot point start from there? Also, we tend to not do full stops at the end of dot points unless there're multiple sentences 😊

plaindocs: Thanks for patiently walking me through the style guide. 

I tried pulling out the "Defined in", but i don't really like it. Both sentences end up super fragmented. Let's chat about the broader layout tomorrow.

plaindocs:
@harrietgrace I think this addresses what we chatted about.

PR #710, opened 2020-04-21
Make h3's linkable, and provide the ability for `id` and `class` overrides

### LINKED ISSUE #708: Add IDs to H3
Hey @toolmantim could you tweak `app/models/page/renderer.rb` and add ID's to all H3s? I think we're better off leaving them out of auto tocs for now - but it would be good to be able to link to them manually. I suspect we just need to dupe L54-59

toolmantim:
Makes all the `h3` elements linkable by adding an automatic id, and let's you customize the `id` and `class`.

For example, the following markdown:

```markdown
## A long complicated section description
{: id="short-id"}

### Subsection 1

## Section 2

### Subsection 2
```

generates:

```html
<h2 id="short-id">A long complicated section description</h2>

<h3 id="short-id-subsection-1">Subsection 1</h3>

<h2 id="section-2">Section 2</h2>

<h3 id="section-2-subsection-2">Subsection 2</h3>
```

This means we can now link to individual [environment variables](https://buildkite.com/docs/pipelines/environment-variables#buildkite-environment-variables), and instead of `#buildkite-environment-variables-buildkite-agent-pid` we can have `#bk-env-vars-buildkite-agent-pid` for example.

Fixes #708

### DIFF
--- a/app/models/page/renderer.rb
+++ b/app/models/page/renderer.rb
@@ -16,6 +16,7 @@ class Page::Renderer
     doc = add_custom_ids(doc)
     doc = add_custom_classes(doc)
     doc = add_automatic_ids_to_headings(doc)
+    doc = add_heading_anchor_links(doc)
     doc = add_table_of_contents(doc)
     doc = fix_curl_highlighting(doc)
     doc = add_code_filenames(doc)
@@ -68,9 +69,8 @@ class Page::Renderer
     doc
   end
 
-  def add_table_of_contents(doc)
-    # First, we find all the top-level h2s
-    headings = doc.search('./h2')
+  def add_heading_anchor_links(doc)
+    headings = doc.search('./h2', './h3')
 
     # Second, we make them all linkable and give them the right classes.
     headings.each do |node|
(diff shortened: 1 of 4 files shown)


toolmantim:
Everything here is up for grabs, so if you’ve got any preferred syntax changes or anything, let me know!

toolmantim:
Ace! Yeah, I was hoping we could use the class/id for a bunch of things.

> I'm happy with this as is, unless you fancy testing the ID clashes just converting H3 without prepending H2 to them? But it is probably not worth the time.

Yeah, I couldn’t think of a good algorithm. Would it be just adding `-2`, `-3` etc to the end of the ids that clash? So as you progress down the page?

Also I didn’t add the hover link doohicky that h2’s have. Should we?

plaindocs:
ooh, I didn't notice the lack of doohicky!  I reckon add it, saves from having to "view source" :-p. Please and thank you.

On the clash front, if there are a few dozen or so site wide, I'd be happy to add manual overrides to all of them, and just put up with an ID clash breaking the build for new content. What do you think?

toolmantim:
> ooh, I didn't notice the lack of doohicky! I reckon add it, saves me from having to "view source" :-p. Please and thank you.

All done! 

doohicky

> On the clash front, if there are a few dozen or so site wide, I'd be happy to add manual overrides to all of them, and just put up with an ID clash breaking the build for new content. What do you think?

I like the idea, but I think I'm keen to stick with the current method for now — 90% because the code is mostly ready, and 10% because of getting my head around dealing with the clashes. Let's stick with the nesting method we have now, and revisit if we're not feeling it's working well.

plaindocs:
:100:

Issue #714, opened 2020-04-26
Mixing depends_on and block/wait doesn't work as expected.


mboudreau:
Let's say I have the following steps:

```
steps:
  - command: yarn build
    key: build
    artifact_paths: dist/**/*
  
  - command: yarn lint
    key: lint
    
  - command: yarn test
    key: test
    
  - block
    
  - depends_on: build
    command:
      - buildkite-agent artifact download 'dist/*' .
      - yarn integration-tests
```

Now, when I read this I assume that the last step that depends on the build but is after the block would mean that this step doesn't run unless 2 dependencies are met: that the step above is unblocked and that the build step finished. However this is not the case in practicality.

From my observation, `block`/`wait` does not mix with `depends_on`, which is frustrating to say the least.  What I've seen is that once the build command is done, any commands that has a `depends_on: build` gets run right after, completely ignoring a `block` or `wait` command before it.  I find this unacceptable.  

Regardless of my opinion, I was confused as to why this is never mentioned in the documentation.  This scenario caused a production deployment when it shouldn't have happened as it should have been blocked;  we were lucky that the deployment didn't affect our current users.

plaindocs:
Hi @mboudreau 

Thanks for pointing this out. I've run a few tests which confirm that `depends_on` does bypass the `block`. I'll find out if this is intended behavior or a bug, and get the docs updated either way.

In the meantime, I think making the `block` step have the dependency does what you intend?

```yaml
steps:
  - command: yarn build
    key: build
    artifact_paths: dist/**/*
  
  - command: yarn lint
    key: lint
    
  - command: yarn test
    key: test
    
  - block
    depends_on: build
  
  -  command:
      - buildkite-agent artifact download 'dist/*' .
      - yarn integration-tests
```

PR #717, opened 2020-04-27
Clarify implicit and explicit dependencies

plaindocs:
Improve docs in response to #714

### DIFF
--- a/pages/pipelines/dependencies.md.erb
+++ b/pages/pipelines/dependencies.md.erb
@@ -100,7 +100,7 @@ If the step you're dependent on doesn't exist, the build will fail without runni
 
 However, if the step you're dependent on is excluded from the build due to an `if` condition, the dependency will be ignored and the step that depends on it will run once any other dependencies are satisfied.
 
-Steps that are in a `concurrency_group` run in the order they are created in, and so if a step depends on another step in a concurrency `concurrency_group` that will affect when the first step is able to run.
+Steps that are in a `concurrency_group` run in the order they are created in and can be delayed in running by the `concurrency` attribute. If your step has a dependency on a step that is in a `concurrency_group`, there is an implicit dependency on the rest of the steps in the group. For more information about concurrency groups, see the [Controlling Concurrency guide](https://buildkite.com/docs/pipelines/controlling-concurrency#concurrency-groups).
 
 ## Allowing dependency failures
 
--- a/pages/pipelines/dependencies.md.erb
+++ b/pages/pipelines/dependencies.md.erb
@@ -50,6 +50,11 @@ steps:
 
 In the above example, the second command step (build) will not run until the first command step (tests) has completed. Without the `depends_on` attribute, and given enough agents, these steps would run in parallel.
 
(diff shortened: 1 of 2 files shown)


harrietgrace:
Great addition 🙌🏻Do you think we should add notes about this elsewhere too? Like on the dependencies page?

plaindocs:
Hey @harrietgrace 

You mean more than the note that I added already? :-p

I did wonder about adding a section about the interactions, and looping in `depends_on: ~` at @matthewd 's sugggestion.

I was also confused by 

> There are three step attributes that can each affect when a step is able to run: if/branches, depends_on, and concurrency_group. They are applied to the step in the order they are listed.

So will tweak that

pages/pipelines/dependencies.md.erb
```diff
@@ -90,12 +90,18 @@ Even though the second command step in the above example is after a wait step, t
...
 ## Order of operations
 
-There are three step attributes that can each affect when a step is able to run: `if`/`branches`, `depends_on`, and `concurrency_group`. They are applied to the step in the order they are listed.
+There are three step attributes that can each affect when a step is able to run:
+
+* `if`/`branches`
+* `depends_on`
+* `concurrency_group`
 
 If the step you're dependent on doesn't exist, the build will fail without running the step that is waiting for the dependency.
```
plaindocs, line 99: @harrietgrace here do you mean if it excluded by a branches condition?

pages/pipelines/dependencies.md.erb
```diff
@@ -90,12 +90,18 @@ Even though the second command step in the above example is after a wait step, t
...
 
 However, if the step you're dependent on is excluded from the build due to an `if` condition, the dependency will be ignored and the step that depends on it will run once any other dependencies are satisfied.
 
+Steps that are in a `concurrency_group` run in the order they are created in, and so if a step depends on another step in a concurrency `concurrency_group` that will affect when the first step is able to run.
```
plaindocs, line 103: Is this what you had in mind for the third bullet point?

pages/pipelines/dependencies.md.erb
harrietgrace: Yep, or if you mis-typed the name, or a pipeline upload didn't work... anything that might cause a step to not be there.

pages/pipelines/dependencies.md.erb
harrietgrace: Mmm, not quite? Concurrency groups are the worst for explaining 😅

How about something that goes a little deeper, along the lines of:

```suggestion
Steps that are in a `concurrency_group` run in the order they are created in and can be delayed in running by the `concurrency` attribute. If your step has a dependency on a step that is in a `concurrency_group`, there is an implicit dependency on the rest of the steps in the group. For more information about concurrency groups, see the [Controlling Concurrency guide](https://buildkite.com/docs/pipelines/controlling-concurrency#concurrency-groups).
```

plaindocs: Perfect - thanks

Issue #709, opened 2020-04-21
Should list possible values for BUILDKITE_S3_ACL on https://buildkite.com/docs/agent/v3/cli-artifact


jnewbigin:
We recently had a need to use
`BUILDKITE_S3_ACL=bucket-owner-full-control`
but I had to refer to the source at
https://github.com/buildkite/agent/blob/30f8cc6526b88833d0411bcbba1527862c3a5207/agent/s3_uploader.go#L133
so see if that was supported (which it looks like it is).

We have not completed this work yet (so I don't know if it works), but this bug is about the missing list of available values in the docs.

Regards,
John.

PR #718, opened 2020-04-27
Add possible values for BUILDKITE_S3_ACL

plaindocs:
Fix for #709


harrietgrace:
Nice! It would be good to add something about there being many options for that value in here too: https://buildkite.com/docs/agent/v3/cli-artifact#using-your-private-aws-s3-bucket

plaindocs:
This is waiting on #708 for the H3 link.

harrietgrace:
I think the underscores in the link to the env var docs need to be escaped, we're getting some funky styling 😅

image

There's a bit of prior art for lists of values, do you think we could do this for the ACL one too? And we could have a `Default:` line for `public-read` too.

image

plaindocs:
Two great catches, thanks @harrietgrace 

I'm not :100: sure what is going on with the underscores, but escaping fixes it, so :man_shrugging:

PR #719, opened 2020-04-29
Revert "Make h3's linkable, and provide the ability for `id` and `class` overrides"

pda:
Reverts buildkite/docs#710

More context in https://3.basecamp.com/3453178/buckets/1735271/messages/2629087723

TL;DR: nokogiri is taking 10-30 seconds to do its thing on https://buildkite.com/docs/pipelines/environment-variables in production and development.

Here's a flamegraph of the interesting part:

image

### DIFF
--- a/app/models/page/renderer.rb
+++ b/app/models/page/renderer.rb
@@ -13,10 +13,6 @@ class Page::Renderer
     # It's like our own little HTML::Pipeline. These methods are easily
     # switchable to HTML::Pipeline steps in the future, if we so wish.
     doc = Nokogiri::HTML.fragment(html)
-    doc = add_custom_ids(doc)
-    doc = add_custom_classes(doc)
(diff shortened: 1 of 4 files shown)


toolmantim:
Thanks so much @pda! Could you share how you made the flamegraph, so I can easily recreate it when doing my spelunking?

pda:
Initially I added the https://github.com/MiniProfiler/rack-mini-profiler gem and the other gems it suggests, and loaded the slow page with `?pp=flamegraph`. But the JS flamegraph renderer kept crashing because there was so much data. It might be possible to turn down the sample rate or something.

So instead I used https://rbspy.github.io/ ... I ran `rbspy record -- bin/rails s` and immediately loaded the page, then ctrl-c stopped the rbspy around when the page load finished, and then looked at the SVG it produced. For some reason I couldn't get rbspy to attach to an already running puma, which would have been nicer.

PR #720, opened 2020-04-29
Fix typo for MATOMO_URL

JuanitoFatas:
🔎 spotted while browsing around 😅


yob:
whoops, this was me :grimacing:

Issue #706, opened 2020-04-15
Cannot view svgs


jl-applied:
URL: https://buildkite.com/docs/integrations/build-status-badges

Chrome console output:
```
VM14 docsearch-c86e5021b8a9b8a33c7f3fca68f1ab6b7c1291c1e7020a3aad0a12b92d18b827.js:2 [Report Only] Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' thttps://analytics.buildkite.com 'nonce-O6fcsw/4UOu1pOF1D6tCQQ=='".

w.makeTemplate @ VM14 docsearch-c86e5021b8a9b8a33c7f3fca68f1ab6b7c1291c1e7020a3aad0a12b92d18b827.js:2
build-status-badges:1 [Report Only] Refused to load the image 'https://analytics.buildkite.com/tracker?action_name=Docs%20%2F%20&idsite=1&rec=1&r=308548&h=0&m=13&s=59&url=https%3A%2F%2Fbuildkite.com%2Fdocs%2Fintegrations%2Fbuild-status-badges&urlref=https%3A%2F%2Fwww.google.com%2F&_id=e8b5ddf8d5df2e0f&_idts=1577140279&_idvc=111&_idn=0&_refts=1586920239&_viewts=1586917714&_ref=https%3A%2F%2Fgithub.com%2F&send_image=1&pdf=1&qt=0&realp=0&wma=0&dir=0&fla=0&java=0&gears=0&ag=0&cookie=1&res=1920x1080&cvar=%7B%221%22%3A%5B%22Application%22%2C%22docs%22%5D%7D&_cvar=%7B%221%22%3A%5B%22Logged%20In%22%2C%22true%22%5D%7D&gt_ms=181&pv_id=jvryyV' because it violates the following Content Security Policy directive: "img-src 'self' https://buildkiteassets.com/ https://buildkite.com/ thttps://analytics.buildkite.com".

sample.svg:1 Failed to load resource: the server responded with a status of 404 ()
```

Will check in later if it is still an issue. Let me know if there is anything else I can provide to help.

EDIT:
![Screenshot at 2020-04-15 00-14-46](https://user-images.githubusercontent.com/49570001/79309314-f222d600-7eae-11ea-9065-2d98d25e7e9d.png)

Occurs with firefox 75.0 and Chrome Version 80.0.3987.132 (Official Build) (64-bit)
```
$ uname -a
[...]

PR #721, opened 2020-04-29
Add BADGE_DOMAIN env

JuanitoFatas:
That points to our https://badge.buildkite.com.

This will fix the svg not showing on our [status badges page](https://buildkite.com/docs/integrations/build-status-badges).

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -19,7 +19,7 @@
 Rails.application.config.content_security_policy do |policy|
   policy.default_src :self
   policy.font_src    :self, 'https://www2.buildkiteassets.com/'
-  policy.img_src     :self, 'https://buildkiteassets.com/', 'https://buildkite.com/', Matomo::URL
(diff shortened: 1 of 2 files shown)


yob:
This seems like a worthwhile change, but does the CSP policy definitely explain why those SVGs aren't loading?

Our CSP is set to "report only", so my understanding is that it shouldn't be blocking images from loading.

#706 is a related issue.

JuanitoFatas:
The CSP error mentioned in #706 is fixed in #720.

Current production has no CSP error ✌️ ✨  

スクリーンショット 2020-04-29 23 09 53

Somehow on production right now we don‘t have this `BADGE_DOMAIN` env:

```erb
<%= image_tag "#{ENV["BADGE_DOMAIN"]}/sample.svg?status=passing" %>
```

スクリーンショット 2020-04-29 23 08 46

`ENV["BADGE_DOMAIN"]` is nil so they tried to load from `/sample.svg`.

The CSP policy is needed because loading images from a subdomain `budget.buildkite.com` is not included in current `img_src`:

```ruby
policy.img_src     :self, 'https://buildkiteassets.com/', 'https://buildkite.com/', Matomo::URL
```

We could also change the existing one to `*.buildkite.com`. I‘m fine with both, but slightly preferred to be specific 😄

toolmantim:
Thank you so much @JuanitoFatas!

PR #722, opened 2020-05-01
PHP Client library

bbaga:
Proposing the [buildkite-php](https://github.com/bbaga/buildkite-php) library to be added to the list of available clients


harrietgrace:
Thanks for taking the time to PR @bbaga! 🙏🏻 And thanks for adding the library in the first place too 😄

PR #733, opened 2020-05-22
Unrendered link in Automatic Retry Attributes blockquote

jmks:
A link in a blockquote for [Automatic Retry Attributes](https://buildkite.com/docs/pipelines/command-step#automatic-retry-attributes) had a link that was still in markdown:

image

The markdown links only seem to work if they are not nested under HTML, so I updated it to an HTML link like the majority of the links on the page:

image


harrietgrace:
Brilliant, thanks @jmks! And thanks for taking the time to PR 😊 🙏🏻

PR #734, opened 2020-05-27
Use friendlier terminology in secure agent guide v2

JuanitoFatas:
I was browsing the guide and thought we could use more friendly terms 😊 

Trailing spaces are also gone upon save.

[Preview link](https://docs-review-update-term-u4nyis.herokuapp.com/docs/agent/v2/securing#allowing)


plaindocs:
Oh damn, we should definitely be fixing this! Thanks @JuanitoFatas <3!

What you think about "safelisting" as an alternative to "allowing"?

PR #736, opened 2020-05-28
Interactive expansion of sidebar section contents

scopegate:
This allows the sections in the sidebar to be expanded in place, so users can browse the contents of a section without having to click into it. 

To-do:
- [x] Add some animation
- [x] Check and improve handling of the agent version logic (ie /v2 vs /v3)
- [ ] Add icons against sections?


app/views/layouts/_sidebar.html.erb
```diff
@@ -0,0 +1,272 @@
+<nav class="Docs__nav">
+  <div class="Docs__nav__section-container" data-docs-path="tutorials">
+    <p class="Docs__nav__section-heading">Tutorials <%= image_tag('nav-exp-icon.svg') %></p>
+    <div class="Docs__nav__section-content">
+      <ul class="Docs__nav__sub-nav">
+        <li class="Docs__nav__sub-nav__item"><%= sidebar_link_to "Getting Started", 'tutorials/getting-started' %></li>
```
blaknite, line 6: I reckon we should turn the rest of the sidebar structure into helpers. There's a lot of repetition going on here. Maybe `sidebar_section` and `sidebar_item`? Personally I'd merge the `ul` into the `sidebar_section` helper and put the `li` into the existing `sidebar_link_to` and rename it.

The `sidebar_section` helper is going to need to take a block. How's your grasp of blocks? Let me know if you want to pair.

blaknite:
Regarding the JS, it does what it says on the tin. Maybe it could be tidier but I'm plenty happy to ship that as is. @ticky may have more to add though.

app/views/layouts/_sidebar.html.erb
scopegate: Great suggestion! It definitely felt a little off to be manually shunting around this much markup. Given we don't kick off YAML proper until Tuesday, maybe we could pair on this sometime Monday?

scopegate: I've wrapped the `li` into the `sidebar_link_to`, but looking at the other sections I'm not sure there's much boilerplate to remove in creating a section helper, as the sections are actually quite different - some have one list within them, some have multiple lists with subheadings, and the Agent section actually has two completely separate sets of content. 🤔

plaindocs:
@scopegate what state is this in? Can we do a minimal amount of work and push it out?

Issue #730, opened 2020-05-19
start multiple agents with different config on macOS reboot ( startup )


patrick-tresp:
i stumbled over this documentation to start multiple agents on linux system.
- https://buildkite.com/docs/agent/v3/ubuntu#running-multiple-agents
i am running a macOS machine to start 4 agents with different tags manually now.

The documentation of macOS does not include this info, the config file seems to be made for 1 agent and not for multiple.

is there any approach to also start multiple agents on startup on macOS?

i only see one config file solution … 

Thanks for any best practices on this,

Patrick

harrietgrace:
Hey @patrick-tresp, sorry for the delay getting back to you. Did you end up sorting out your macOS agents, or are you still looking for some examples? 

There are two different ways you can get going with multiple agents, both using the `launchd` service. The quickest way is to edit the agent's plist to include a `--spawn` flag with the number of agents you need. However, if you need different config for each agent, you can duplicate the plist and change the config for each one. 

I've just started a PR to add a section about running multiple agents to the macOS installation guide 👍🏻I'll leave this issue open until that's merged 😊

patrick-tresp:
oh thanks for the support and enhancing the docs! i love it!

PR #738, opened 2020-06-01
Add multiple agents info to the macos agent install doc

harrietgrace:
So far this is just some notes, but eventually it will be full sentences and have some example code too. We already supply people with a plist, so I'm adding some explainers on how to add the --spawn option or how to dupe the whole plist. 

This PR adds a section on Running multiple agents to the macos agent installer doc, mirroring the one we've got in the Linux agent installer doc. 

Closes https://github.com/buildkite/docs/issues/730


harrietgrace:
@keithduncan could you please have a squiz at this and see if it all looks right? 😊

PR #742, opened 2020-06-04
Upgrade most of gems

JuanitoFatas:
This PR:

* Use Rails 6 and other gems
* Lock rouge because syntax highlight css classes changed (we should upgrade this separately)
* Simplify `Page#render`


yob:
Upgrade all of the things!

My personal preference is to upgrade gems in smaller batches over multiple PRs, so testing each PR is easier and any surprises are easier to track down.

This is a pretty simple app though, so maybe we can get  away with all-at-once?

JuanitoFatas:
> My personal preference is to upgrade gems in smaller batches over multiple PRs

We can turn on dependantbot and configure it in a less noisy way. 

> This is a pretty simple app though, so maybe we can get away with all-at-once?

The tests passed and clicking through the [reviews app](https://docs-review-upgrade-gem-jacc3l.herokuapp.com/docs/tutorials/getting-started) everything looks okay.

plaindocs:
I'm :+1: on dependabot ongoing.

I've clicked around the review app and looks fine.

app/models/page.rb
```diff
@@ -60,10 +60,9 @@ def get_binding
     end
 
     def render(*args)
-      av = ActionView::Base.new
-      av.view_paths = ActionController::Base.view_paths
-      av.view_paths << Rails.root.join("app/views/pages").to_s
-      av.render(*args)
+      ActionController::Base.append_view_path "app/views/pages"
+      renderer = ActionController::Base.renderer.new
+      renderer.render(partial: args.first)
```
JuanitoFatas, line 65: @matthewd What do you think about the change here? 🙇‍♂️

matthewd: The first line is mutating a global value, but this method gets called multiple times.

Sounds to me like the `append_view_path` should be in `PagesController`, then this would be:

```ruby
def render(partial)
  PagesController.render(partial: partial)
end
```

JuanitoFatas: Thanks! Fixed in 4488f93.

PR #745, opened 2020-06-10
Add sidebar docs

blaknite:
This adds a JSON document and markdown files to support the sidebar for the YAML pipepline configuration editor.


scopegate:
Looks good! The only question I have is whether, given we have a sidebar that lives within the docs app, and that this content will likely be used for more things down the track: do we want to call this something else (ie `pipeline_reference` or something) to avoid potential confusion?

blaknite:
That can work. I'm happy with a name that's based on it's content, not where it's used. This almost comes across to me as a quick reference or a cheat sheet. Maybe I can set up a path that is `/docs/quick-reference/pipelines`?

scopegate:
@blaknite sounds good to me! 👍🏻

PR #746, opened 2020-06-14
Fix minor typos

vineetgopal:
This changes a few incorrect usages of "it's" to "its".

### DIFF
--- a/pages/agent/v2/cli_artifact.md.erb
+++ b/pages/agent/v2/cli_artifact.md.erb
@@ -146,7 +146,7 @@ Example:
 
 Options:
 
-   --step value                Scope the search to a paticular step by using either it's name or job ID
+   --step value                Scope the search to a paticular step by using either its name or job ID
    --build value               The build that the artifacts were uploaded to [$BUILDKITE_BUILD_ID]
    --agent-access-token value  The access token used to identify the agent [$BUILDKITE_AGENT_ACCESS_TOKEN]
    --endpoint value            The Agent API endpoint (default: "https://agent.buildkite.com/v3") [$BUILDKITE_AGENT_ENDPOINT]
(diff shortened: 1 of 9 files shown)


pages/agent/v2/cli_artifact.md.erb
```diff
@@ -226,7 +226,7 @@ Example:
    $ buildkite-agent artifact shasum "pkg/release.tar.gz" --build xxx
 
    This will search for all the files in the build with the path "pkg/release.tar.gz" and will
-   print to STDOUT it's SHA-1 checksum.
+   print to STDOUT its SHA-1 checksum.
```
plaindocs, line 229: It might be worth rewording this one entirely. Maybe something like 

> This will search for all files in the build with the path "pkg/release.tar.gz" and will print the SHA-1 checksum of each one to STDOUT.

pages/agent/v3/help/_artifact_shasum.txt
```diff
@@ -16,7 +16,7 @@ Example:
    $ buildkite-agent artifact shasum "pkg/release.tar.gz" --build xxx
 
    This will search for all the files in the build with the path "pkg/release.tar.gz" and will
-   print to STDOUT it's SHA-1 checksum.
+   print to STDOUT its SHA-1 checksum.
```
plaindocs, line 19: As above.

Also, these files in the `help` directory are generated from https://github.com/buildkite/agent/, please open a PR there to fix the source if you'd like to.

vineetgopal: Thanks for the pointer, @plaindocs. I took a look at buildkite/agent and it looks like there was a "it's" -> "its" change made 5 months ago in https://github.com/buildkite/agent/commit/a52af04aa69f8d844996842224f5aaa8c358ff93

Do you know if there's a process to regenerate the docs in buildkite/docs from buildkite/agents? It seems like the source was changed but it didn't actually propagate to this repo. Otherwise, happy to make the changes in both places.

plaindocs: Hey @vineetgopal that would be great. I'm looking into a process for making sure changes propagate.

vineetgopal: Posted a change here: https://github.com/buildkite/agent/pull/1229

PR #753, opened 2020-06-23
Refactor installation and upgrade docs

plaindocs:
I started looking at getting rid of the "Upgrade from 2 to 3" docs, and then started thinking about promoting the installer sections out of Agent (note the formatting is off due to the JS path detection) 

image

I think we could go further and remove the v2/v3 selector from prime page real estate, and just leave a tiny 2 page with links in the menu, untill we can deprecate v2 altogether. 

image

Slight caveat around removing the selector code if we're in any danger of introducing a v4

This work isn't actually done, I'm just looking at low effort ways of streamlining the navigation.

### DIFF
--- a/pages/agent/v3/upgrading.md.erb
+++ b/pages/agent/v3/upgrading.md.erb
@@ -1,10 +1,23 @@
-# Upgrading to Buildkite Agent v3
+# Upgrading your Buildkite Agents
 
-The Buildkite Agent has changed a lot in v3, but we've tried to keep it as backwards compatible as possible. We'll cover what’s changed and how to upgrade to new agents.
+Upgrade your Agents using your operating system package manager, or by re-running the installation script.
 
 {:toc}
 
-## Overview of what has changed
+## Upgrading Agents on Ubuntu Linux
+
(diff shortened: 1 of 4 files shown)


pages/agent/v3/upgrading.md.erb
```diff
@@ -1,10 +1,23 @@
-# Upgrading to Buildkite Agent v3
+# Upgrading your Buildkite Agents
 
-The Buildkite Agent has changed a lot in v3, but we've tried to keep it as backwards compatible as possible. We'll cover what’s changed and how to upgrade to new agents.
```
yob, line 3: :100:

harrietgrace:
@pda Is there any timeline on releasing v4 of the agent? Is it likely to be this year?

harrietgrace:
Lets say that it's not going to be this year. Could we keep the selector commented out? So remove it visually, but not remove the ability to do the thing?

pda:
> @pda Is there any timeline on releasing v4 of the agent? Is it likely to be this year?

Sorry, I missed this earlier. Nope! No major version bump planned at this point, very unlikely for this year.

PR #756, opened 2020-06-25
Add draft content for Command step type

scopegate:
Adding more representative draft content for Command step type (mostly copied wholesale from existing docs content) to ensure there's nothing we're missing with regards to how we want to serve up and display sidebar content.

### DIFF
--- a/app/views/quick_reference/steps/command/_agents.md.erb
+++ b/app/views/quick_reference/steps/command/_agents.md.erb
@@ -1,0 +1,7 @@
+### agents
+
+A map of [agent tag](https://buildkite.com/docs/agent/v3/cli-start#setting-tags) keys to values to [target specific agents](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting) for this step. 
+
+```yaml
+npm: "true"
+```
--- a/app/views/quick_reference/steps/command/_agents.md.erb
+++ b/app/views/quick_reference/steps/command/_agents.md.erb
@@ -1,5 +1,3 @@
-### agents
-
 A map of [agent tag](https://buildkite.com/docs/agent/v3/cli-start#setting-tags) keys to values to [target specific agents](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting) for this step. 
 
 ```yaml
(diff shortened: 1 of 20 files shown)


plaindocs:
Hey @scopegate but what if we wrote a little bit of code to parse out each H3 from the doc and populate these files from that? Similar to how the `update-agent-help.sh` works? That way we wouldn't be maintaining two sets of content. 

We could even mark out the chunks with kramdown ids, or custom divs which have no effect on the display to make them less error prone. 

I'd be super happy to restructure the docs to help make that work.

CC @harrietgrace do you have any other ideas about ways to do this?

scopegate:
Hey @plaindocs, that could work! @blaknite and I spoke about a slightly different path though, which is essentially flipping it the other way, and updating the docs pages down the track to pull from the structured content, as it'll probably be less prone to error that way and feels a little cleaner. It would also mean we have a single content source feeding:

- Docs pages
- In-app sidebar content
- Pipeline schema/descriptions

Are you ok with this approach? @harrietgrace seemed to think it sounded ok, but I'd be really happy to talk it over and reconsider as a group if anyone has reservations.

plaindocs:
Yeah, that would work also. Miiight be worth double checking that we're happy with the overall structure of that particular doc before we do that, as it'll be a PITA work with for major organizational changes.

app/views/quick_reference/steps/command/_agents.md.erb
```diff
@@ -0,0 +1,7 @@
+### agents
```
ticky, line 1: are we putting titles in this? if so why are we rendering titles in the frontend component?

ticky:
I assembled some “fake” content over in https://github.com/buildkite/buildkite/pull/5176, because I needed more stuff to test the filtering. What’s the plan with integrating existing content?

app/views/quick_reference/steps/command/_agents.md.erb
scopegate: Whoops, good pickup! I've stripped the titles out of the markdown content

scopegate:
> What’s the plan with integrating existing content?

So long as everyone's comfortable with this approach, I'll keep going today and try to get all the other content up, at least in a draft state, which should obviate the need for local placeholder data.

ticky:
> So long as everyone's comfortable with this approach

Has the approach been discussed? The original PR seems to have been merged with no review and I don’t remember any broader discussions of the format?

PR #759, opened 2020-06-30
Reintroduce making h3's linkable

toolmantim:
Reverts #719, and re-implements #710 
- [x] Fix performance problems


toolmantim:
@plaindocs @harrietgrace this is good to go!

PR #760, opened 2020-06-30
mention that BUILDKITE_BUILD_AUTHOR and BUILDKITE_BUILD_AUTHOR_EMAIL can be blank

yob:
These two variables will sometimes be blank, best let folks know


yob:
> We're not super consistent around using Buildkite web interface, Buildkite web UI, web, but I think the first two are slightly more clear than the last

I've gone with `Buildkite web interface`

PR #765, opened 2020-07-07
Parse step attributes out of existing Markdown content

ticky:
This parses attributes' information out of our Markdown pages.

The existing table structures we use in the docs are parsed for attributes, and the title and description are parsed out of the document itself.

In this case, we are using CommonMarker, specifically for its robust support for looking at an abstract syntax tree of the document, as well as for the reliable parsing. This way, we’re not doing any string manipulation or regex matching, but still getting a good result.

It comes with a few rules:

1. The `name` of the step is the text of the first h1 in the document.
2. The `shortDescription` is the text of first paragraph of the document.
3. The `textContent` is the Markdown source of any number of paragraphs up until the first fenced code block (and inclusive of the code block), or paragraphs up until a second-level or deeper heading is found. The `{:toc}` directive, if present, is stripped.
4. `attributes` is an array of the attributes found in `table` elements which have a `data-attributes` attribute on them.
   - Each `tr` is treated as an individual attribute
   - The `name` is the text of the first child of the row
   - The `textContent` is the HTML content of the last child of the row, wrapped in a `<div>` element.
   - `isRequired` will be true if the containing `table` element also has a `data-attributes-required` attribute

This results in this transformation, from rendered Docs page to sidebar:

image


eleanorakh:
This is super nice @ticky! I can't see anything that stands out to me as a blocker for merging, but I'd like another set of eyes on it! cc/ @blaknite

ticky:
@harrietgrace ah yes, thanks for the reminder - I will give that one more pass today before merging this!

PR #764, opened 2020-07-06
Replace web steps with YAML steps

harrietgrace:
Turns out there are heaps of places we talk about the web UI step in our docs! This PR removes them and updates the existing YAML examples or adds them in if they were missing. 

Going to hold this one as a draft until the end of the week so that it doesn't get accidentally merged before YAML steps goes live. 

@plaindocs could you please cast your eyes over all this and check that it all still makes sense? 🙏🏻

PR #766, opened 2020-07-07
Remove just from targeting a queue section

JuanitoFatas:
https://buildkite.com/docs/agent/v3/queues#targeting-a-queue


plaindocs:
Hey @JuanitoFatas thanks for yet another pull request!

@harrietgrace has an open PR for a ton of stuff over in #764 and I've made a much more simplified suggest for this exact text in https://github.com/buildkite/docs/pull/764#discussion_r450351978 -> what do you think?

PR #770, opened 2020-07-09
Fix the emoji rendering on the new Emojis page

toolmantim:
This fixes the Emojis page so it's rendering the new(ish) emojis from the updated submodule.


ticky:
aahhh I’ve been really trying to fight the “emojis” name, it’s not a good plural of it! can we keep it as emoji everywhere we mention it? the one thing that’s been left is the emoji repo name

toolmantim:
Ah, the emoji/emojis change was included in here because I was on a call with @plaindocs when we were doing this.

Maybe we do in fact just go with "Emoji" @plaindocs @harrietgrace? Do we know from search what customers are using to refer to it?

plaindocs:
Algolia search logs have an appalling data retention policy. Recent searches are for `emoji` not `emojis`. I don't know how much that has to do with intentionally searching for a term that covers both.

toolmantim:
Sounds like emoji it is!

Issue #772, opened 2020-07-13
Add caveat that a timeout is not considered soft failure


niceking:
In resolving [this customer query](https://secure.helpscout.net/conversation/1218997068/10917?folderId=1113725), the customer pointed out that our docs on [soft failure](https://buildkite.com/docs/pipelines/command-step#soft-fail-attributes) don't reflect the fact that timeout is not considered a soft failure type.

PR #776, opened 2020-07-15
Clarification of soft-fail / timeout

plaindocs:
I'm not super keen on the double negative, but I think this is an improvement. 

Is the `*` the same as `true` here? Regardless, we should pick one.

Fix for #772


niceking:
I had a look through the code, and it looks like specifying `*` and `true` map to the same behaviour!

plaindocs:
Thanks for checking @niceking ! :-)

PR #786, opened 2020-08-01
Page description

sj26:
Expose a meta description tag for docs pages, and rework the rendering slightly closer to rails convention to make it work, then hook it up to the existing data extractor thingo.

Screen Shot 2020-08-01 at 12 23 14 pm

I'm not entirely happy with the `show.html.erb` being amongst the docs pages, but there's some symlinking and trickery involved which makes it difficult to do otherwise.


app/controllers/pages_controller.rb
```diff
@@ -1,22 +1,18 @@
 class PagesController < ApplicationController
-  append_view_path "app/views/pages"
+  append_view_path "pages"
```
sj26, line 2: `app/views/pages` was already in the view path for this controller, but `pages` isn't.

app/models/page.rb
```diff
@@ -94,6 +94,10 @@ def title
     agentize_title(contents.match(/^\#\s(.+)/).try(:[], 1) || "")
   end
 
+  def description
+    extracted_data.fetch("shortDescription")
```
sj26, line 98: This will raise if a short description doesn't exist, but by design. The tests make sure that every page renders, which means every page has a description. And the data extractor should be able to make a description for every page as long as it has even a little bit of content, and no page should have no content.

PR #788, opened 2020-08-04
Bugsnag

sj26:
Adds Bugsnag for error tracking. The SSM parameter for `BUGSNAG_API_KEY` is already in place, with an appropriate project setup in Bugsnag: https://app.bugsnag.com/buildkite/buildkite-docs/errors

### DIFF
--- a/config/initializers/bugsnag.rb
+++ b/config/initializers/bugsnag.rb
@@ -1,0 +1,3 @@
+Bugsnag.configure do |config|
+  config.api_key = ENV["BUGSNAG_API_KEY"]
+end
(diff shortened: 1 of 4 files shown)


config/initializers/bugsnag.rb
```diff
@@ -0,0 +1,3 @@
+Bugsnag.configure do |config|
+  config.api_key = ENV["BUGSNAG_API_KEY"]
```
sj26, line 2: This does nothing in development and test, but configures error reporting in production where the env is set.

sj26:
Sneaky deploy — this has no user facing functionality, and I want to see what this error is:
https://buildkite.com/docs/pipelines/my-builds

PR #790, opened 2020-08-04
Add test for /docs/pipelines/my-builds

JuanitoFatas:
Follow up of #789.

### DIFF
--- a/spec/features/reading_pages_spec.rb
+++ b/spec/features/reading_pages_spec.rb
@@ -32,6 +32,20 @@ RSpec.feature "reading pages" do
     end
   end
 
+  describe "all pages" do
+    example "render" do
+      root = Rails.root.join("pages")
+      root.glob("**/*.md{,.erb}").each do |path|
+        url = "/docs" + path.to_s.delete_prefix(root.to_s).delete_suffix(".erb").delete_suffix(".md")
+        puts "Visiting #{url}"
+        visit url
+        if !page.status_code.in?([200, 403])
+          raise "#{url} returned #{page.status_code}"
+        end
+      end
+    end
+  end
+
   describe "clicking links" do
     class Check < Struct.new(:path, :source_link, :fragment, keyword_init: true)
     end


spec/features/reading_pages_spec.rb
```diff
@@ -159,6 +159,7 @@ class Check < Struct.new(:path, :source_link, :fragment, keyword_init: true)
       /docs/tutorials/sso-setup-with-graphql
       /docs/webhooks/setup
       /docs/webhooks
+      /docs/pipelines/my-builds
```
sj26, line 162: I'm all for speccing this — yes, please! — but I'm not sure this is the right spot for this. 🤔 This spec is about making sure that redirects work — that old URLs go to current URLs. This isn't a redirect, nor an 'old' url, it's a 'new' url which hasn't been published anywhere yet.

I did consider adding a new spec which iterates all `Rails.root.join("pages").glob("**/*.md{,.erb}")` paths and checks that the matching urls render, but it seems like it had a lot of overlap with the spidery spec, and I couldn't figure out a nice way to do that — plus there were some pages which were broken and I was out of decision making power about what to do with them.

sj26: This is the spec I wrote:
https://github.com/buildkite/docs/compare/spec-all-pages

but it probably needs to be merged with the `clicking links` spec.

plaindocs: I'll defer to @sj26 on the actual implementation of speccing this. 

But as an alternative suggestion, what if we listed all unlisted pages :-p. That way unless you intentionally add the page to an exclusion, a page without a link errors, which is probably (?) a more useful check?

JuanitoFatas: a4d5e2c spec tests 139 paths which 69 of them are covered in `clicking links`. So we are testing 70 extra paths, which I reckon is a win and how to properly marry them, we can think about later... 😅

sj26: @JuanitoFatas yeah, good call, and nice work! Sorry, I ran out of energy to dig deeper last night.

plaindocs:
Awesome stuff! Thanks @sj26 @JuanitoFatas

PR #794, opened 2020-08-12
Update REST API agent docs to clarify when connected/disconnected agents are returned

yob:
I was surprised to see that state is handled differently by the agent list and get endpoints.

I'm not sure if we can fix that in the short term. There are scenarios where it would be helpful to list disconnected agents, but there may be performance implications. Elastic stack style agent handling can result in a very high number of disconnected-and-never-coming-back agents.

While we discuss that, we can at least document the current behaviour.


pages/apis/rest_api/agents.md.erb
```diff
@@ -4,7 +4,7 @@
 
 ## List agents
 
-Returns a [paginated list](<%= paginated_resource_docs_url %>) of an organization’s agents.
+Returns a [paginated list](<%= paginated_resource_docs_url %>) of an organization’s agents. The list only includes running agents - agents in a disconnected state are not returned.
```
plaindocs, line 7: Hey @yob, apologies for resurrecting a merged PR.

Do you think 

> The list only includes connected agents - agents in a disconnected state are not returned.

makes more sense, as it includes connected but idle agents?

yob: yup, that reads much better. Technically agents in a "stopping" state will also be returned, but they're not in that state for very long and if you squint it's almost the same as connected. For this purpose at least.

I'll open a revised PR.

PR #796, opened 2020-08-14
Improve API docs

plaindocs:
Improve the API docs landing page, special collaboration edition.


pages/apis/apis.md.erb
```diff
@@ -0,0 +1,50 @@
...
+2. Open the [GraphQL Explorer](https://graphql.buildkite.com/explorer)
+3. Follow the [GraphQL Tutorial](https://building.buildkite.com/tutorial-getting-started-with-graphql-queries-and-mutations-11211dfe5d64)
+
+More information about the [GraphQL API](/docs/apis/graphql-api).
+
+## Webhooks
+
+Webhooks allow you to monitor and respond to events within your Buildkite organization, providing a real time view of activity and allowing you to extend and integrate Buildkite into your systems.
+
+Webhooks can be added and configured on your [organization's Notification Services settings](https://buildkite.com/organizations/-/services) page.
```
plaindocs, line 48: This link is also present in `webhooks.md.erb` and doesn't work. What should it be instead?

sj26:
Oh, I think we might need to remove this redirect now so that clicking "APIs" goes to the overview page:
https://github.com/buildkite/docs/blob/f3bd3acf0ec995b5db328ea1c66663b6a3fc2db6/config/routes.rb#L70

sj26:
I've made two small changes.

The first means that from the Docs page, clicking the "APIs" link will take you to the API overview page:

image

The second is that I changed the title to "Buildkite APIs" to fit with "The Buildkite Agent", "Buildkite Plugins", "Deployments with Buildkite" etc — "Overview" felt awkward in the new home, and I think it's better Google juice to put "Buildkite" in there:

image

plaindocs:
Thanks for those two great fixes @sj26 , I keep forgetting where that default page setting lives.
:+1: on the API naming also.

plaindocs:
Let's merge this now, and improve later.

sj26:
Nice 🙌

PR #799, opened 2020-08-17
fix: add concurrency_method docs

joscha:
Was added on April 17 by @matthewd :)

cc @keithpitt 

Some examples here: https://github.com/joscha/buildkite-graph/pull/42/

cc @uri-canva @arromer


pages/pipelines/controlling_concurrency.md.erb
```diff
@@ -82,3 +82,22 @@ steps:
...
+This is the default behaviour (`ordered`) and for example most helpful for deploys, etc.
+
+However, in some cases concurrency groups are not used in order to guarantee a limited number of execution to happen at the same time and preserving order, they can be used to restrict access to a limited resource for example - let's say a limited capacity in a SaaS service like Saucelabs or similar.
+In that case, the ordering of the executions can work against you, as one execution is waiting for the next before taking up another concurrency slot. If your resource usage tiem is very different (let's say some cross-browser test executes in pipeline A after 1 minute whereas in another pipeline B it executes after 10 minutes then potentially you could run multiple short tests of pipeline B before the execution of the step in pipeline A is due. Ordering is not helpful because it means that depending on the order in which these tests were queued in, the limited resource you're controlling concurrency for is not utilized fully. For this you can set the concurrency method to `eager`, which removes the ordering condition.
```
joscha, line 93:
```suggestion
In that case, the ordering of the executions can work against you, as one execution is waiting for the next before taking up another concurrency slot. If your resource usage time is very different (let's say some cross-browser test executes in pipeline A after 1 minute whereas in another pipeline B it executes after 10 minutes then potentially you could run multiple short tests of pipeline B before the execution of the step in pipeline A is due. Ordering is not helpful because it means that depending on the order in which these tests were queued in, the limited resource you're controlling concurrency for is not utilized fully. For this you can set the concurrency method to `eager`, which removes the ordering condition.
```

pages/pipelines/controlling_concurrency.md.erb
```diff
@@ -82,3 +82,21 @@ steps:
...
+This is the default behaviour (`ordered`) and for example most helpful for deploys, etc.
+
+However, in some cases concurrency groups are not used in order to guarantee a limited number of execution to happen at the same time and preserving order, they can be used to restrict access to a limited resource for example - let's say a limited capacity in a SaaS service like Saucelabs or similar.
+In that case, the ordering of the executions can work against you, as one execution is waiting for the next before taking up another concurrency slot. If your resource usage time is very different (let's say some cross-browser test executes in pipeline A after 1 minute whereas in another pipeline B it executes after 10 minutes then potentially you could run multiple short tests of pipeline B before the execution of the step in pipeline A is due. Ordering is not helpful because it means that depending on the order in which these tests were queued in, the limited resource you're controlling concurrency for is not utilized fully. For this you can set the concurrency method to `eager`, which removes the ordering condition.
```
joscha, line 93:
```suggestion
In that case, the ordering of the executions can work against you, as one execution is waiting for the next before taking up another concurrency slot. If your resource usage time is very different - let's say some cross-browser test executes in pipeline A after 1 minute whereas in another pipeline B it executes after 10 minutes then potentially you could run multiple short tests of pipeline B before the execution of the step in pipeline A is due. Ordering is not helpful because it means that depending on the order in which these tests were queued in, the limited resource you're controlling concurrency for is not utilized fully. For this you can set the concurrency method to `eager`, which removes the ordering condition.
```

pages/pipelines/controlling_concurrency.md.erb
```diff
@@ -82,3 +82,21 @@ steps:
...
 ```
+
+### Controlling execution order
+
+By default steps that belong to the same concurrency group are executed in the order they are added to the pipeline execution.
+E.g. if you have a step A and it uses the concurrency group `X` with a concurrency of `1` at time 0 and you add another step B with the same concurrency group `X` and also a concurrency of `1` at time 1, then you are guaranteed that step A will run before step B.
+This is the default behaviour (`ordered`) and for example most helpful for deploys, etc.
+
+However, in some cases concurrency groups are not used in order to guarantee a limited number of execution to happen at the same time and preserving order, they can be used to restrict access to a limited resource for example - let's say a limited capacity in a SaaS service like Saucelabs or similar.
```
plaindocs, line 92:
```suggestion
However, in some cases concurrency groups are not used in order to guarantee a limited number of execution to happen at the same time and preserving order, they can be used to restrict access to a limited resource for example - let's say a limited capacity in a SaaS service like Sauce Labs or similar.
```

plaindocs:
Thanks so much @joscha - this one dropped through the cracks. :(

joscha:
> Thanks so much @joscha - this one dropped through the cracks. :(

never too late :)

plaindocs:
:tada:

PR #800, opened 2020-08-17
Add some friendly GraphQL info

plaindocs:
This started as an updated version of https://buildkite.com/blog/getting-started-with-graphql-queries-and-mutations but got bogged down in a lack of variables.

As it stands there isn't much here that isn't also in `pages/apis/graphql_api.md.erb` but might be worth adding anyway. :man_shrugging:


plaindocs:
Woah, I forgot about this.

PR #802, opened 2020-08-19
Steps may run on difference agents

plaindocs:
@sj26 pointed out that folks don't always realize that steps can run on different agents. 
This is a small attempt to clarify that.


JuanitoFatas (approved):
👍

sj26:
Nice one, @plaindocs. 🙌

PR #803, opened 2020-08-20
Offer quoting advice for conditionals

sj26:
One of our customers was bitten by this [today](https://secure.helpscout.net/conversation/1258368214/11765?folderId=1113725) and so I think it might be worth a note in the docs, somewhere. Here's an idea, but I'm not sure I love the words:

Screen Shot 2020-08-20 at 12 00 00 pm

### DIFF
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -18,7 +18,7 @@ steps:
 
 The `if` attribute can be used in any type of step, and with any of the supported expressions and parameters. However, it cannot be used at the same time as the `branches` attribute.
 
-Be careful when defining conditional within YAML. Many symbols have special meaning in YAML and will change the type of a value. You can avoid this by quoting your conditional as a string.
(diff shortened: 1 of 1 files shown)


pages/pipelines/conditionals.md.erb
```diff
@@ -18,7 +18,17 @@ steps:
 
 The `if` attribute can be used in any type of step, and with any of the supported expressions and parameters. However, it cannot be used at the same time as the `branches` attribute.
 
-Multi-line conditionals can be added with the `|` character:
+Be careful when defining conditional within YAML. Many symbols have special meaning in YAML and will change the type of a value. You can avoid this by quoting your conditional as a string.
```
plaindocs, line 21:
```suggestion
Be careful when defining conditionals within YAML. Many symbols have special meaning in YAML and will change the type of a value. You can avoid this by quoting your conditional as a string.
```

plaindocs (approved):
Yup, this looks great.

pages/pipelines/conditionals.md.erb
sj26: Whoops, good spot!

PR #807, opened 2020-08-25
Add `skip` key to Trigger step docs

ngan:
The trigger step supports the `skip` key but it’s not documented.


JuanitoFatas (approved):

JuanitoFatas:
This is great!  Thanks Ngan ❤️💜💛💙💚

PR #808, opened 2020-08-26
Update pipeline update to reflect recent changes.

jayco:
We made some changes here to allow the `configuration` value to be used to set the steps in the pipeline update patch, these are just some small changes to make that a little clearer, we could possibly do an overhaul of all accept values here long term.

Changes:

Adds configuration and tip

Screen Shot 2020-08-26 at 2 04 37 pm

Updates the returned value

Screen Shot 2020-08-26 at 2 04 55 pm


plaindocs (approved):
:+1:

PR #811, opened 2020-08-31
Add blocked builds

harrietgrace:
Add the new `blocked_state` attribute and description to the block step table 🌈

### DIFF
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -53,8 +53,9 @@ _Optional attributes:_
     <tr>
     <td><code>blocked_state</code></td>
     <td>
-      <br>
-      <em>Values:</em> <code>passed</code>, <code>failed</code>, <code>pending</code>
+      The state that the build should be set to when the build is blocked by this block step.<br>
+      <em>Default:</em> <code>passed</code><br>
+      <em>Values:</em> <code>passed</code>, <code>failed</code>, <code>running</code>
     </td>
   </tr>
   <tr>
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -54,7 +54,7 @@ _Optional attributes:_
     <td><code>blocked_state</code></td>
     <td>
       The state that the build is set to when the build is blocked by this block step. If you don't set this to something other than the default (`nil`), the build will show as passed in the Buildkite UI and other integrations.<br>
-      <em>Default:</em> <code>passed</code><br>
+      <em>Default:</em> <code>nil</code><br>
       <em>Values:</em> <code>passed</code>, <code>failed</code>, <code>running</code>
     </td>
   </tr>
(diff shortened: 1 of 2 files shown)


pages/pipelines/block_step.md.erb
```diff
@@ -42,6 +42,14 @@ _Optional attributes:_
       A list of input fields required to be filled out before unblocking the step.<br>
       Available input field types: <code>text</code>, <code>select</code>
     </td>
+  </tr>
+    <tr>
+    <td><code>blocked_state</code></td>
+    <td>
+      The state that the build should be set to when the build is blocked by this block step.<br>
+      <em>Default:</em> <code>passed</code><br>
```
blaknite, line 50:
```suggestion
      <em>Default:</em> <code>nil</code><br>
```

A bit of a technicality in the implementation is that the value defaults to `nil` and Buildkite UI and notifications interpret that as `passed`. If a value isn't set, the user will see `nil` in the API.

JuanitoFatas (approved):
🥇

pages/pipelines/block_step.md.erb
plaindocs: Hey folks, I'm slightly confused about this one. If set to `passed` then is it like `soft_fail`?

blaknite: Similar, yes. The impact is it will show with the green passed colour on Buildkite. On other connected sites it depends on the fidelity of the integration. For some it shows simply as passed, just like a soft fail. On others it'll show as "Passed and blocked".

harrietgrace: Perhaps we need a short section about this, or more info in the description?

plaindocs: Sounds good to me. @harrietgrace you wanna tackle that ;-) ? Or shall I?

pages/pipelines/block_step.md.erb
```diff
@@ -42,6 +42,14 @@ _Optional attributes:_
       A list of input fields required to be filled out before unblocking the step.<br>
       Available input field types: <code>text</code>, <code>select</code>
     </td>
+  </tr>
+    <tr>
+    <td><code>blocked_state</code></td>
+    <td>
+      The state that the build should be set to when the build is blocked by this block step.<br>
```
blaknite, line 49:
```suggestion
      The state that the build should be set to when the build is blocked by this block step. The default behaves the same as `passed` in the Buildkite UI and other integrations.<br>
```

plaindocs: Thanks for pushing this again @blaknite hat do you think about something along the lines of 

> If you don't set this to something other than the default (`nil`), the build will show as passed in the Buildkite UI and other integrations.

blaknite: Yeah love it!

plaindocs, line 49:
```suggestion
      The state that the build is set to when the build is blocked by this block step. If you don't set this to something other than the default (`nil`), the build will show as passed in the Buildkite UI and other integrations.<br>
```

plaindocs (approved):
Once that edit goes in this looks good to me.

PR #815, opened 2020-09-03
Mention GitHub apps and Repository Providers

plaindocs:
- [x] Update GH page
- [x] Fixup screenshot (@scopegate pretty please and thank you?)


yob (commented):
I like it!

My off-the-cuff-not-a-writer first response: what would it be like to specifically call out the App as our recommended integration style, maybe with the advantages from intro moved down to the App section?

... and maybe call out the user OAuth approach as totally legit, but older and less preferred.

plaindocs:
No non-writer-disclaimer needed here! :-)

You need the advantages in the intro so people know to pick the App section. :-p

Do we do anything like 

```
 ## To connect your Buildkite account to GitHub using the GitHub App (Recommended) approach
```

elsewhere I wonder? But with the combo of advantages as listed, and it being the first section I don't think we need to.

yob:
Cool, I don't feel strongly about those thoughts.

With the draft-as-is, I think were I still a customer I'd wonder about the differences in the two connection methods. Is there a reason one is recommended? Are they feature compatible?

I'm not sure about how to fit that in though. The tl;dr that I'm thinking it'd be good to communicate:

* either method can be used, there's no difference in the features they support at this stage
* some future features on buildkite.com may only work  with the app method
* the app approach is preferred because it gives GitHub organization admins more visibility into the permissions we ask for
* we anticipate eventually deprecating and removing the oauth method, but there's no timeline scheduled yet

plaindocs:
I've added a sentence that should address some of your concerns @yob

pages/integrations/github.md.erb
```diff
@@ -8,12 +8,29 @@ To complete this integration, you will need admin privileges for your GitHub rep
 
 ## Connecting Buildkite and GitHub
 
+You can now use our [GitHub App](#to-connect-your-buildkite-account-to-github-using-the-github-app) to connect a Buildkite organization to a GitHub organization, removing the reliance on individual user connections to report build statuses. See the [changelog announcement](https://buildkite.com/changelog/102-github-app-integration).
+
+If you want to [connect using OAuth](#to-connect-your-buildkite-account-to-github-using-oauth) you can still do so from your Buildkite Personal Settings.
+
+## To connect your Buildkite account to GitHub using the GitHub App
+
+Connecting Buildkite and GitHub using the GitHub App lets your GitHub Organization Admins see permissions and manage access on a per-repository basis, instead of having to do so for each user.
```
yob, line 17:
```suggestion
Connecting Buildkite and GitHub using the GitHub App lets your GitHub Organization Admins see permissions and manage access on a per-repository basis.
```

yob (approved):
Thanks for incorporating my feedback, it looks good!

I have one small suggestion :+1:

scopegate:
Screenshot updated @plaindocs — I'm sorry to keep you waiting!

plaindocs:
Thanks for the screenshots @scopegate ! <3

PR #818, opened 2020-09-05
Move `skip` attribute to the right section

ngan:
This is a followup fix for https://github.com/buildkite/docs/pull/807. In the original PR, I had misplaced the `skip` key description in the wrong section. It should have been placed under `Optional Attributes`. I accidentally placed it under "Optional `build` attributes", which is incorrect.


ngan:
cc @JuanitoFatas

JuanitoFatas (approved):
Oops it is totally my fault and thank you for the follow-up fix! 👍

PR #820, opened 2020-09-09
Homebrew install does not currently work with the token argument

jayco:
We have an issue raised in the agent https://github.com/buildkite/agent/issues/1284 around a bug with installing using brew and the config not storing the token value. From initial investigations it does not look like this is something that is going to work with changes to homebrew so this PR removes those references.

### DIFF
--- a/pages/agent/v2/osx.md.erb
+++ b/pages/agent/v2/osx.md.erb
@@ -15,7 +15,7 @@ If you have <a href="http://brew.sh/">Homebrew</a> installed you can use our <a 
 
 ```shell
 brew tap buildkite/buildkite
-brew install --token='INSERT-YOUR-AGENT-TOKEN-HERE' buildkite-agent
+brew install buildkite-agent
 ```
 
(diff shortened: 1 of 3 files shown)


ticky:
I’ve made a complementary change in buildkite/homebrew-buildkite#18. Looking into whether adapting the Unix install script for this would be feasible.

jayco:
Thanks for taking care of that complimentary change @ticky! 👌

ticky:
Okay, I’ve got a proposed change for the Unix `install.sh` script to add support for using buildkite-agent from Homebrew and auto-setting its configuration. buildkite/agent#1290

sj26:
Great!

This will need to be updated over here as well:
https://github.com/buildkite/buildkite/blob/master/frontend/app/components/docs/06-macos.js

sj26 (approved):

sj26:
Let's merge this much and then adjust instructions as they change with @ticky's pulls.

PR #821, opened 2020-09-10
Update Homebrew install docs to use our sed token one-liner

ticky:
This updates the docs which were updated in #820 to follow the pattern from buildkite/buildkite#5523, which in turn borrows from our Linux setup instructions.


jayco (approved):
This is great, thanks @ticky  🎉

sj26 (approved):
Great 👍

PR #822, opened 2020-09-15
API/Webhook authentication clarification

sj26:
I found the opening of [this page](https://buildkite.com/docs/apis) a little confusing after re-reading it today:

Screen Shot 2020-09-15 at 6 24 07 pm

The words a little interchanged, and suggest that webhook tokens and access tokens are interrelated.

A little shuffling of words and reordering of paragraphs makes more sense, to me:

image

### DIFF
--- a/pages/apis.md.erb
+++ b/pages/apis.md.erb
@@ -2,7 +2,7 @@
 
 ## Authentication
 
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Much improved. :+1:

PR #823, opened 2020-09-15
Mention include_retried_jobs works for 'Get a build', ignores value

EnricoMi:
* the flag `include_retried_jobs` also works for 'Get a build'
* the value of `include_retried_jobs` is ignored, so `?include_retried_jobs=false` is interpreted as `?include_retried_jobs=true`.


plaindocs:
Hi @EnricoMi thanks for taking the time to improve the docs. 

As far as I can tell both `include_retried_jobs=false` and `include_retried_jobs=true` are working as you'd expect. Do you have a build showing where that setting is ignored? Thanks!

EnricoMi:
I can confirm that this works now as expected, it did not by the time I raised this issue.

However, the documentation still does not mention this optional parameter, the "Get a build" section does not mention any "Optional query string parameters".

plaindocs:
Thanks for looping back, and  for this second fix! :sparkler:

PR #824, opened 2020-09-16
Strict branch configuration

sj26:
Some updates to the branch configuration docs to demonstrate the changes we're hoping to introduce. We hope this will make branch configurations easier to understand when combining positive and negative patterns.

image

p.s. please don't merge yet — this is for after https://github.com/buildkite/buildkite/pull/5545

### DIFF
--- a/pages/pipelines/branch_configuration.md.erb
+++ b/pages/pipelines/branch_configuration.md.erb
@@ -2,9 +2,7 @@
 
 You can use branch patterns to ensure pipelines are only triggered when necessary. This guide shows you how to set up branch patterns for whole pipelines and individual build steps.
 
-<!--alex ignore just-->
-
-In step-level and pipeline-level branch filtering you can use `*` as a wildcard, and `!` for not, as shown in the [examples](#branch-pattern-examples). If you want a full range of regular expressions that operate on more than just branch names, take a look at the [conditionals](/docs/pipelines/conditionals) page.
+In step-level and pipeline-level branch filtering you can use `*` as a wildcard, and `!` for not, as shown in the [examples](#branch-pattern-examples). If you want a full range of regular expressions that operate on more than branch names, take a look at the [conditionals](/docs/pipelines/conditionals) page.
 
 {:toc}
 
--- a/pages/pipelines/branch_configuration.md.erb
+++ b/pages/pipelines/branch_configuration.md.erb
@@ -28,7 +28,7 @@ steps:
     command:
       - "npm install"
       - "tests.sh"
-    branches: "master feature/* release/*"
+    branches: "master feature/* !feature/beta release/*"
   - block: "Release notes"
     prompt: "Please add notes for this release"
     fields:
--- a/pages/pipelines/branch_configuration.md.erb
+++ b/pages/pipelines/branch_configuration.md.erb
@@ -28,7 +28,7 @@ steps:
     command:
       - "npm install"
       - "tests.sh"
-    branches: "feature/* release/* master !infrastructure"
+    branches: "master feature/* release/*"
   - block: "Release notes"
     prompt: "Please add notes for this release"
     fields:
--- a/pages/pipelines/branch_configuration.md.erb
+++ b/pages/pipelines/branch_configuration.md.erb
@@ -52,7 +52,7 @@ steps:
 
 ## Branch pattern examples
 
-When you define a value in the `branch` attribute of a step, only builds on the specified branch/es will contain that step. Adding more values to the `branch` attribute will expand the set of branches that will include that step.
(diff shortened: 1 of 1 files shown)


pages/pipelines/branch_configuration.md.erb
```diff
@@ -30,7 +28,7 @@ steps:
     command:
       - "npm install"
       - "tests.sh"
-    branches: "feature/* release/* master !infrastructure"
+    branches: "master feature/* release/*"
```
plaindocs, line 31: I'm slightly confused by saying you can do both and removing the 'not' from the first example.

pages/pipelines/branch_configuration.md.erb
```diff
@@ -58,12 +56,12 @@ When you define a value in the `branch` attribute of a step, only builds on the
 
 The following are examples of patterns, and the branches that they will match:
 
-* `master production` will match master and production only
-* `master features/*` will match master and any branch that starts with `features/`
-* `*/release-123` will match any branch that ends with `/release-123` such as `feature-123/release-123`
-* `v*.0` will match branch that begins with a `v` and ends with a `.0`, such as `v1.0`
-* `v*.*.*` will match branch that begins with a `v` and has two periods, such as `v1.2.3`
+* `master` will match `master` only
```
plaindocs, line 59: These are all super clear.

pages/pipelines/branch_configuration.md.erb
```diff
@@ -58,12 +56,12 @@ When you define a value in the `branch` attribute of a step, only builds on the
...
 * `!production` will match any branch that's not `production`
-* `staging !production development` will match staging and development only
+* `master features/*` will match `master` and any branch that starts with `features/`
+* `*-test` will match any branch ending with `-test`, such as `rails-update-test`
+* `stages/* !stages/production` will match any branch starting with `stages/` except `stages/production`, such as `stages/demo`
+* `v*.0` will match any branch that begins with a `v` and ends with a `.0`, such as `v1.0`
+* `v* !v1.*` will match any branch that begins with a `v` unless it also begins with `v1.`, such as `v2.3`, but not `v1.1`
```
plaindocs, line 65: So the change is basically that your disallowed branch can also include one that you've previously wildcarded? Is it worth saying that in an abstract way along with the examples? 

And that it doesn't work the other way round?

pages/pipelines/branch_configuration.md.erb
sj26: I removed the negative because it was redundant, it already did not match the other patterns.

But perhaps we would modify this pipeline yaml to include a meaningful example?

pages/pipelines/branch_configuration.md.erb
sj26: Yeah, perhaps we could mention in the explanatory text that they can be combined. Check out the changelog for some possible words.

pages/pipelines/branch_configuration.md.erb
plaindocs:
```suggestion
    branches: "master feature/* !feature/beta release/*"
```

pages/pipelines/branch_configuration.md.erb
```diff
@@ -58,12 +56,12 @@ When you define a value in the `branch` attribute of a step, only builds on the
```
plaindocs, line 56: When combining positive and negative patterns, any positive pattern must match, and every negative pattern must not match.

plaindocs: I first wrote this as include/exclude, but we're def using positive/ngative?

pages/pipelines/branch_configuration.md.erb
plaindocs: How about that? I too can copy paste. :-)

pages/pipelines/branch_configuration.md.erb
plaindocs, line 56:
```suggestion
When combining positive and negative patterns, any positive pattern must match, and every negative pattern must not match.
```

plaindocs: oops, that was supposed to be a suggestion first time around.

sj26: Great, thank you!

sj26: @plaindocs was this meant to replace the paraphraph? I think it's merged three paragraphs together, and I'm not sure it makes sense 😅

plaindocs: yeah, I suspect I mangled it using the GH suggestion interface. 

Looks like we just need the new text, right?

PR #825, opened 2020-09-16
Clarify webhook triggered build prmission issues

### LINKED ISSUE #798: Trigger permissions docs out of date
This part here:
https://buildkite.com/docs/pipelines/trigger-step#permissions

I don't think it's correct any more. We also allow triggering if the source and target pipelines share a team with "build" permissions (except for public -> private pipelines):
https://github.com/buildkite/buildkite/blob/f1fe2354d66577b9a7d7763e082b75687b1e7351/app/models/step/trigger.rb#L90-L131

I explained this in the forum recently, but perhaps should update the docs too:
https://forum.buildkite.community/t/could-not-find-a-matching-team-that-includes-both-pipelines/1182/3

plaindocs:
Fixes #798 

Looks, I totally misunderstood the fix, will update.

### DIFF
--- a/pages/pipelines/trigger_step.md.erb
+++ b/pages/pipelines/trigger_step.md.erb
@@ -1,6 +1,6 @@
 # Trigger Step
 
-A *trigger* step creates a build on another pipeline. 
+A *trigger* step creates a build on another pipeline.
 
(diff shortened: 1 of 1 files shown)


sj26 (approved):
Yeah, this sounds right to me - great!

PR #826, opened 2020-09-18
Clarify failure

plaindocs:
Minor clarification of what happens when commands fail.


yob (approved):
I just ran a [test build](https://buildkite.com/yob-opensource/gist-test/builds/91#00cea949-9b1c-4b45-b999-2a05f0e325e1) and confirmed this is true.

PR #827, opened 2020-09-18
Add Posthog

toolmantim:
Adds PostHog tracking. The required environment variable values have already been added.

### DIFF
--- a/app/models/posthog.rb
+++ b/app/models/posthog.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Posthog
+  # Default to the free, hosted Posthog account which we send all dev data
+  URL = ENV.fetch("POSTHOG_URL", "https://app.posthog.com").freeze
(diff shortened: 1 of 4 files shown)


toolmantim:
I just double checked this again, and looks good to go! Gonna merge 🚁

app/models/posthog.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Posthog
+  # Default to the free, hosted Posthog account which we send all dev data
+  URL = ENV.fetch("POSTHOG_URL", "https://app.posthog.com").freeze
+  API_KEY = ENV.fetch("POSTHOG_API_KEY") { "fPfkF4tU2OBcVBZhcshmzC7_VglTw078wuqD3tl01ic" }.freeze
```
sj26, line 6: (p.s. Is this a legit key for a dev instance? Do we all have access to it, if so?)

sj26: (keeping in mind this is a public repo)

plaindocs:
Awesomeness

app/models/posthog.rb
toolmantim: Yep, these are public keys (they're exposed in the page source)

PR #830, opened 2020-09-22
Fix inconsistent label for variable

aimran-adroll:
👋🏽


plaindocs (approved):
Great catch, thanks @aimran-adroll !

PR #832, opened 2020-09-27
Add create example with configuration

jayco:
Similar to the issue with the configuration value missing in https://github.com/buildkite/docs/pull/808, this change adds an example with the create API request:

Screen Shot 2020-09-27 at 3 17 34 pm


JuanitoFatas (approved):
Thanks Jason!

plaindocs (commented):
Hey Folks, 

Cheers for the update! Are we doing something different than show in

image

https://buildkite.com/docs/apis/rest-api/pipelines#create-a-pipeline

If we're not, would we be better off linking them both?

jayco:
@plaindocs I have linked in the changelog as discussed, sorry about the slow turn around here. Let me know if you would like me to change any of the wording?

It looks like:

Screen Shot 2020-10-30 at 9 20 12 am

and links to https://buildkite.com/changelog/99-introducing-the-yaml-steps-editor

plaindocs:
Hey @jayco - thanks for the update.

I've looked at this again, and I was still confused, so I think we need to go further. I've pushed up some changes, we can discuss tomorrow.

PR #834, opened 2020-10-01
Add AWS Quick Start examples

toolmantim:
AWS published a number of EventBridge example implementations which could be very useful for people building custom Buildkite event-based integrations. This adds links to all three of them.

- [x] Add the links
- [x] Perhaps we use the example repo style from https://buildkite.com/docs/plugins/tools


toolmantim:
Amazing! Thanks @plaindocs <3

PR #843, opened 2020-10-28
environment_variables.md: BUILDKITE_PULL_REQUEST may be unset

jgehrcke:
We have noted that BUILDKITE_PULL_REQUEST
may be not set at all when a build is triggered by a
branch push (no PR yet created).


jgehrcke:
Sorry, this is like to have been a problem on our side.

PR #846, opened 2020-10-29
Remove Agent v2 selector

plaindocs:
Replaces  #753 and actually does the work instead of showing what should be done.

I'm assuming that upgrading from 3.0 to 3.x is smooth? We also don't go into any detail about successive 3.x update, but I'd rather not  have the burden of putting those in the docs unless there is an obvious advantage to doing so.


pages/agent/v3/upgrading.md.erb
```diff
@@ -24,17 +24,17 @@ Deprecated:
...
 
-## Bootstrap customizations
+### Bootstrap customizations
 
 If you customized your `bootstrap.sh` file, you will need to move the changes to [hooks](hooks), or update your bootstrap.sh to call `buildkite-agent bootstrap`.
 
-## Docker and Docker Compose support
+### Docker and Docker Compose support
 
 In v2 we supported a variety of environment variables like `BUILDKITE_DOCKER_COMPOSE_CONTAINER` and `BUILDKITE_DOCKER`. These are deprecated in favour of the [docker-compose](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin) and [docker](https://github.com/buildkite-plugins/docker-buildkite-plugin) pipeline plugin.
```
yob, line 33: I think this section is only important for folks upgrading from v2 - maybe we can tuck it away in the `Upgrading from a 2.0 agent` section at the bottom?

yob (commented):
This is a great move. Telemetry shows *very* few folks still running a v2 agent, I agree it doesn't warrant a full section for v2. 

Upgrading from 3.<stable> to 3.<stable> should be very straightforward. We've worked hard to keep all those releases backward compatible, and plan a v4 release if we ever want to remove deprecated features.

pages/agent/v3/upgrading.md.erb
plaindocs: @yob wanna take a look at the updated results? If you're happy with that all I am.

PR #848, opened 2020-10-30
Update to alex 9

plaindocs:
Doesn't look like the major Alex version change is what is causing the Alex linting step to hang. CC @yob 

But the combo of latest alpine and latest Alex might be?


yob:
Screenshot from 2020-10-31 16-35-34

oh javascript :disappointed:

yob (approved):
I'm not completely sure how this helped with the hanging builds, but it seems to have down the trick. No harm in keeping semi-up-to-date with dependencies either.

plaindocs:
Yeah, it was a bit of a throw stuff at the wall scenario. :-p

PR #865, opened 2020-11-16
Add job.scheduled to Job Event details docs

codeman9:
This event is present on the overview page: https://buildkite.com/docs/apis/webhooks but not on the detail page: https://buildkite.com/docs/apis/webhooks/job-events


plaindocs:
Great catch, thanks @codeman9

PR #867, opened 2020-11-18
Lets see what indexing these looks like

plaindocs:


plaindocs:
Merging as this only affects the test db

PR #877, opened 2020-12-01
Document soft fail and allowed dependency failure

### LINKED ISSUE #845: clarify how allow_dependency_failures relates to soft_fail
Say I have a build step `step-a` that may soft_fail:

```
...
    key: "step-a"
    soft_fail:
      - exit_status: 42
...
```

Say I have another build step  `step-b` that depends on that with an explicit `allow_dependency_failure: false`:

```
...
    key: "step-b"
    depends_on:
      - "step-a"
    allow_dependency_failure: false
...
```
What's the intended behavior? Is `step-b` meant to execute when `step-a` soft-fails?

The first observation is that I think this is not covered in the documentation. 

I think I would expect this to be discussed in https://buildkite.com/docs/pipelines/dependencies in section [Allowing dependency failures](https://buildkite.com/docs/pipelines/dependencies#allowing-dependency-failures)

Personally, I was thinking that "I mean, it would be sane to _not_ trigger `step-b` because a soft failure is still a failure". Secondly, I was _hoping_ for that to be true, because then this can be used as a pretty powerful technique: "when step A soft-fails, don't trigger step B".

It looks like the actual behavior of today is that `step-b` is executed even when `step-a` soft-fails.

plaindocs:
Fixes #845

### DIFF
--- a/pages/pipelines/dependencies.md.erb
+++ b/pages/pipelines/dependencies.md.erb
@@ -131,3 +131,24 @@ steps:
 {: codeblock-file="pipeline.yml"}
 
 This pattern is often used to run steps like code coverage or annotations to the build log that will give insight into what failed.
+
+## Allowed failure and soft fail
+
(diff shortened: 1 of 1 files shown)


jayco (approved):
LGTM 🎉

PR #896, opened 2021-01-06
[Amazon EventBridge] Fix event pattern rule

shays10:
Using the original example raised the following error in EventBridge:
`
Event pattern is not valid. Reason: "slug" must be an object or an array at [Source: (String)"{ "detail-type": [ "Build Started", "Build Finished" ], "detail": { "pipeline": { "slug": "some-pipeline" } } }"; line: 5, column: 16]
`

It turns out the correct pattern is `"pipeline": { "slug": [ "some-pipeline" ] }` according to [Event Patterns docs](https://docs.aws.amazon.com/eventbridge/latest/userguide/filtering-examples-structure.html)


JuanitoFatas (approved):
Thank you so much ❤️

plaindocs:
Well spotted, thanks for fixing!

PR #903, opened 2021-01-12
Update troubleshooting headers from h1s to h3s

### LINKED ISSUE #878: Fix headings for troubleshooting notes
Hey @scopegate we currently use `h1` for all of our troubleshooting notes which doesn't seem right:

```
<div class="Docs__troubleshooting-note">
<h1>depends_on and block/wait</h1>
<p>Note that a step with an explicit dependency specified with the <code>depends_on</code> attribute will run immediately after the dependency step has completed, without waiting for <code>block</code> or <code>wait</code> steps unless those are also explicit dependencies.</p>
</div>
```

What is the best way to give this a title which isn't outside of the doc structure? Just `p id="title"` ?

scopegate:
Update stylesheet and content for the orange troubleshooting headers from H1s (typically used once on a page, for page title) to H3s.

image

Resolves #878

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -282,7 +282,7 @@ a.Docs__example-repo {
   border-left-color: orange;
   margin-top: 2rem;
   margin-bottom: 2rem;
-  h1 {
+  h1, h3 {
     font: inherit;
     color: inherit;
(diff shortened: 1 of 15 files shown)


JuanitoFatas (approved):

plaindocs:
@scopegate you're a star!

PR #909, opened 2021-01-19
Mention in docs that include_retried_jobs works for 'Get a build'

EnricoMi:
The flag `include_retried_jobs` also works for 'Get a build', but this is missing in the [docs](https://buildkite.com/docs/apis/rest-api/builds#get-a-build). Re-opens #823.


JuanitoFatas (approved):
Woo thanks so much for improving our docs 👍🏼 👍🏼 

🏆

PR #913, opened 2021-01-26
Fix yarn error building docker image

JuanitoFatas:
This fixes the following error:

```
GPG error: http://dl.yarnpkg.com/debian stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 23E7166788B63E1E
```

which is currently breaking our build 😅


JuanitoFatas:
sneaky merge

plaindocs:
Thanks @JuanitoFatas !

PR #918, opened 2021-01-28
Add shutdown hook

plaindocs:


keithpitt:
Here's a few more comments/nuggest on info on this particular hook:

1. It's run outside of the job lifecycle
2. Output appears in the logs of the agent, not in any job logs
3. It's run once, no matter how many agent workers there are
4. It's useful for entire agent after it's finished cleanup, not cleanup on a per-job basis

plaindocs:
Thanks for the extra info @keithpitt - I think that update covers it.

PR #920, opened 2021-01-29
Revert "Fix yarn error building docker image"

JuanitoFatas:
Reverts buildkite/docs#913

The original `ADD` command should have noticed the update, but didn't for some reason, so the change to a `RUN` busted the cache. But the `RUN` will never have the opportunity to notice upstream changes and so will cause the same problem in the future, whereas the `ADD` should notice updates and also creates lighter layers and maintains consistency.


sj26 (approved):
Yes please :-)

JuanitoFatas:
@sj26  Looks like we need to recycle the agent  😅

スクリーンショット 2021-01-29 11 34 21

sj26:
@JuanitoFatas there's another culprit at the heart of it:

https://github.com/buildkite/buildkite/commit/673ab3f0effa3d721ee9668e9416edf78cd2c3a4

PR #941, opened 2021-02-22
Better title

plaindocs:
Hey @JuanitoFatas 

I'm not keen on using Documentations, but I think this change captures what you were trying to fix?


plaindocs:
I also found this https://www.searchenginejournal.com/pipe-or-dash-in-title-tag/378099/#close  :-D

JuanitoFatas (approved):
Awesome!

PR #942, opened 2021-02-23
List meta endpoint

### LINKED ISSUE #940: Document REST API endpoint that lists the IP addresses we use for egress traffic
We have a REST that publishes the IP addresses we use for egress traffic (like webhooks and commit statuses) at this unauthenticated API: https://api.buildkite.com/v2/meta

I can't see it mentioned in the docs, and it'd be nice to do so.

It works like so:

```bash
$ curl "https://api.buildkite.com/v2/meta"
{"webhook_ips":["54.165.103.71"]}
```

plaindocs:
Close #940


yob (approved):
perfect :+1:

PR #944, opened 2021-02-25
Add "created_by" to the Pipelines API docs

ticky:
As added in buildkite/buildkite#6094


plaindocs (approved):
Thanks a ton @ticky !

PR #945, opened 2021-02-25
This endpoint creates webhook for pipeline using GitHub

JuanitoFatas:
The response is `201 Created` (no content).

### DIFF
--- a/pages/apis/rest_api/pipelines.md.erb
+++ b/pages/apis/rest_api/pipelines.md.erb
@@ -565,7 +565,7 @@ Success response: `204 No Content`
 
 ## Add a webhook
 
-Create webhook for an existing pipeline.
+Create SCM webhook for an existing pipeline using GitHub.
 
 ```bash
 curl -X POST "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{slug}/webhook"
--- a/pages/apis/rest_api/pipelines.md.erb
+++ b/pages/apis/rest_api/pipelines.md.erb
@@ -565,7 +565,7 @@ Success response: `204 No Content`
 
 ## Add a webhook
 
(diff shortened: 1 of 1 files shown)


pages/apis/rest_api/pipelines.md.erb
```diff
@@ -565,7 +565,7 @@ Success response: `204 No Content`
 
 ## Add a webhook
 
-Create webhook for an existing pipeline.
+Create SCM webhook for an existing pipeline using GitHub.
```
blaknite, line 568:
```suggestion
Create an SCM webhook for an existing pipeline. Currently supports pipelines configured using our GitHub App.
```

blaknite (approved):

yob (commented):
neat, the extra detail will be useful!

I reckon it's worth waiting for @plaindocs docs to take a look too, just in case there's additional feedback he has.

plaindocs (changes requested):
I think that with this change we can push it out.  

One other possiblity would be to change the title to mention GitHub, as it is the only option.

plaindocs (approved):

pages/apis/rest_api/pipelines.md.erb
```diff
@@ -565,7 +565,7 @@ Success response: `204 No Content`
 
 ## Add a webhook
 
-Create an SCM webhook for an existing pipeline. Currently supports pipelines configured using our GitHub App.
+Create an Githook webhook for an existing pipeline that is configured using our GitHub App. Pushes to the linked GitHub repository will trigger builds.
```
plaindocs, line 568: Pretty sure I meant to write GitHub here. :man_facepalming:  

```suggestion
Create an GitHub webhook for an existing pipeline that is configured using our GitHub App. Pushes to the linked GitHub repository will trigger builds.
```

PR #946, opened 2021-02-27
Fix invalid command substitution in plugin docs.

malclocke:
Fails with `bash: PWD: command not found`.  Should either be $PWD or
$(pwd), I've opted for the former.


plaindocs (approved):
Thanks for fixing!

PR #953, opened 2021-03-03
Clarify IP address policy in meta api docs

sj26:
Also clarify that the Meta API might grow more properties which describe Buildkite over time as required.

image

### DIFF
--- a/pages/apis/rest_api/meta.md.erb
+++ b/pages/apis/rest_api/meta.md.erb
@@ -10,7 +10,7 @@ It does not require authentication.
 
 Returns an object with properties describing Buildkite.
 
-`webhook_ips` is a list of IP addresses that Buildkite uses to send outbound traffic such as webhooks and commit statuses. These are subject to change from time to time. We recommend checking for new addresses daily, and will try to advertise new addresses for at least 7 days before they are used.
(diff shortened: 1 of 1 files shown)


ticky (approved):

pages/apis/rest_api/meta.md.erb
```diff
@@ -1,22 +1,26 @@
...
+Note: The IP addresses shown here are examples. You must query the API to get the current set of IP addresses.
 
 ```bash
 curl "https://api.buildkite.com/v2/meta"
 ```
 
 ```json
 {
-  "webhook_ips": ["100.24.182.113", "35.172.45.249", "54.165.103.71", "54.85.125.32"]
+  "webhook_ips": ["192.0.2.0/24", "198.51.100.12"]
```
pda, line 23: Ooh a sneaky CIDR in the example. I like that. But maybe the documentation words should clarify that each item could be a single address or a CIDR range?

pda: I also like that they're not real IP addresses, so there's less risk of people copy/pasting straight from the docs.

plaindocs (approved):
This looks good to me.

I'm not :100: sold on using example addresses, nothing stops folks hitting the endpoint manually, and copy pasting those. But I guess nudging folks in the right direction is all we can do.

PR #963, opened 2021-03-10
Bump rails (and web console)

plaindocs:


yob (approved):
@plaindocs and I paired on this and it seems to be working well! :+1:

PR #970, opened 2021-03-16
Add parallel job envs

JuanitoFatas:
Add these two from [Parallel Jobs](https://buildkite.com/docs/tutorials/parallel-builds#parallel-jobs).

CleanShot 2021-03-16 at 13 52 20@2x

https://docs-review-more-buildk-qonyjd.herokuapp.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-parallel-job

### DIFF
--- a/pages/pipelines/environment_variables.md.erb
+++ b/pages/pipelines/environment_variables.md.erb
@@ -305,7 +305,7 @@ Example: `"acme-inc"`
 
 <h3 class="h3-caps">BUILDKITE_PARALLEL_JOB</h3>
 
-The number of parallel jobs created from a parallel build step. Starting from 0. For a build step with `parallelism: 5`, the value would be 0, 1, 2, 3, and 4 respectively.
+The index of each parallel job created from a parallel build step, starting from 0. For a build step with `parallelism: 5`, the value would be 0, 1, 2, 3, and 4 respectively.
 
 Example: `"0"`
 
--- a/pages/pipelines/environment_variables.md.erb
+++ b/pages/pipelines/environment_variables.md.erb
@@ -303,6 +303,18 @@ The organization name on Buildkite as used in URLs. The value cannot be modified
 
 Example: `"acme-inc"`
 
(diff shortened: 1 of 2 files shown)


pages/pipelines/environment_variables.md.erb
```diff
@@ -303,6 +303,18 @@ The organization name on Buildkite as used in URLs. The value cannot be modified
 
 Example: `"acme-inc"`
 
+<h3 class="h3-caps">BUILDKITE_PARALLEL_JOB</h3>
+
+The number of parallel jobs created from a parallel build step. Starting from 0. For a build step with `parallelism: 5`, the value would be 0, 1, 2, 3, and 4 respectively.
```
plaindocs, line 308: So this is the "index" for a specific job created from a parallel step? 

We might need something like

The index of each parallel job created from a parallel build step. Starts from 0 and has a maximum of (total parallelism -1 ).  For a build step with `parallelism: 5`, the value for each job would be 0, 1, 2, 3, and 4 respectively.

pages/pipelines/environment_variables.md.erb
```diff
@@ -303,6 +303,18 @@ The organization name on Buildkite as used in URLs. The value cannot be modified
...
+
+Example: `"0"`
+
+<h3 class="h3-caps">BUILDKITE_PARALLEL_JOB_COUNT</h3>
+
+The total number of parallel jobs created from a parallel build step. For a build step with `parallelism: 5`, the value is 5.
```
plaindocs, line 314: This is the max parallel jobs right? There could be a case where there are only 2 parallel jobs running but there theoretically could be 5? Or is it the actual count of parallel jobs created?

plaindocs (changes requested):
Thanks @JuanitoFatas !

Couple of questions inline.

pages/pipelines/environment_variables.md.erb
JuanitoFatas: If `parallelism` says 5 it should be 5 for that step. Otherwise, something went wrong 😅

plaindocs (approved):

PR #973, opened 2021-03-22
Update suggested manifest for deploying the agent to GKE

yob:
Our GKE docs suggest a kubernetes Deployment with type "extensions/v1beta1/Deployment", however Deployments moved to "apps/v1/Deployment" sometime ago and the old namespace is [unavailable from kuberentes 1.16](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) (released mid 2019).

Our GKE docs in general are a bit dated and need an overhaul, but this is a worthwhile quick-fix that'll save customers hitting a confusing error.

I've also added matchLabels, because I think Deployment's are supposed to have them.

This diff is mostly whitespace - I recommend viewing [with whitespace changes hidden](https://github.com/buildkite/docs/pull/973/files?diff=split&w=1).

### DIFF
--- a/pages/agent/v3/gcloud.md.erb
+++ b/pages/agent/v3/gcloud.md.erb
@@ -63,35 +63,38 @@ Create a [deployment](https://kubernetes.io/docs/user-guide/deployments/#creatin
 
 ```shell
 $ cat | kubectl apply -f -
-apiVersion: extensions/v1beta1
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Thanks @yob !

And doubly so for the thoughtful PR.

PR #974, opened 2021-03-22
Show cli help as rendered markdown not as plain text

plaindocs:
I'm kinda excited about this one. :tada: 

Reviewers:

- [x] suggestions on better approaches for the ruby or script welcome

Still do to: 

- [ ] might need to tweak how we show cli options (what do we use elsewhere
- [x] double check all headings still work

# Before 
image

# After
image


toolmantim:
I didn’t know how it would go rendering this to markdown, but I love it! 😍

toolmantim (approved):
Hey, if the Ruby works! 👍🏼 Might be worth renaming the files to `.md` from `.txt` now tho?

plaindocs:
Yeah, we do need to promote all H2 to H3 in pages where we include multiple "man" pages

```
  "artifact download"
  "artifact shasum"
  "meta-data exists"
  "meta-data get"
  "meta-data keys"
  "meta-data set"
  "step get"
  "step update"
```

scripts/cli2md.rb
```diff
@@ -0,0 +1,21 @@
...
+        puts "`#{line.strip}`"      
+    # Code sections
+    elsif /\s{3}\$/ =~ line
+        puts " #{line}"
+    # Lists of parameters
+    #    --config value                         Path to a configuration file [$BUILDKITE_AGENT_CONFIG]
+    elsif /\s{3}(-{2}[a-z0-9\- ]*)([A-Z].*)$/ =~ line
+        command = $1.rstrip
+        desc    = $2
+        desc.gsub!(/(\$BUILDKITE[A-Z0-1_]*)/,"`\\1`")
```
ticky, line 16: My gut says this might be a typo:

```suggestion
        desc.gsub!(/(\$BUILDKITE[A-Z0-9_]*)/,"`\\1`")
```

ticky: (the --no-http2 flag shows the issue with the current implementation)

scripts/cli2md.rb
```diff
@@ -0,0 +1,21 @@
+ARGF.each_with_index do |line, line_num|
+    # Headings 
+    if /^([A-z]*):/ =~ line
```
ticky, line 3: Might be better to just use the "word character" selector here; `A-z` includes a bunch of other characters which aren't alphabetical:

```suggestion
    if /^(\w*):/ =~ line
```

scripts/cli2md.rb
```diff
@@ -0,0 +1,21 @@
...
+    elsif /\s{3}\$/ =~ line
+        puts " #{line}"
+    # Lists of parameters
+    #    --config value                         Path to a configuration file [$BUILDKITE_AGENT_CONFIG]
+    elsif /\s{3}(-{2}[a-z0-9\- ]*)([A-Z].*)$/ =~ line
+        command = $1.rstrip
+        desc    = $2
+        desc.gsub!(/(\$BUILDKITE[A-Z0-1_]*)/,"`\\1`")
+        puts "* `#{command}` - #{desc}"
+    elsif
```
ticky, line 18: Should this just be an else? I'm surprised permits an `elsif` without a condition!

```suggestion
    else
```

ticky (approved):
Looks great! A couple of little code issues but otherwise it's a great little upgrade :)

scripts/cli2md.rb
plaindocs: Ah yes, I fixed that but didn't commit. Took me a while though. Good eyes.

plaindocs:
Thanks @ticky, that is super helpful.

plaindocs:
@ticky would you mind taking another look? If you can suggest a better way of demoting headings in a subset of the files that would be great. The "meta-data" page  for examples, includes 

```
../pages/agent/v3/cli_meta_data.md.erb
13:<%= render 'agent/v3/help/meta_data_set.md' %>
21:<%= render 'agent/v3/help/meta_data_get.md' %>
25:<%= render 'agent/v3/help/meta_data_exists.md' %>
29:<%= render 'agent/v3/help/meta_data_keys.md' %>
```

so the headings need to be H3 not H2 for the page structure to work as is.

ticky:
Hmm, are those used in multiple places or just one?

plaindocs:
@ticky each include is included once, but three of them are included in `/agent/v3/cli_meta_data`, basically, here is the meta-data page, you can set, get, check if it exists and look at keys....

ticky:
Yeah, I think my other thought on how you could approach it is to make the CLI output to Markdown conversion happen on demand rather than manually, and leave the files as txt. Then you can add a parameter to the helper when called in the view to tell it how far to indent. It may be overly optimising though, I’m not sure it’s better than the sed solution if it works! 😂

plaindocs:
Yeah, I tried that over in #975 but couldn't quite make it work. I think I'm just gonna merge it for now, as good enough. :+1:

PR #975, opened 2021-03-22
Trying to demote the partial

plaindocs:
I'm not entirely sure I'm making the conversion in the right place here.

My initial plan was to wrap the demote in an condition for the demote:option but that doesn't quite work as is.


plaindocs:
This doesn't work, so I've put the logic in the conversion over in #974 instead.

Issue #980, opened 2021-03-24
Clarify default queue tagging 


niceking:
This issue came through in a [customer email](https://secure.helpscout.net/conversation/1463525853/17020?folderId=1113730), where the customer was confused by this section of the [Buildkite Agent Job Queue](https://buildkite.com/docs/agent/v3/queues) doc.

```
The default queue
All agents include an extra default tag of queue=default.

Setting no queue value will match every available agent in the 'default' queue.
```

This makes it sound like every agent is tagged with the default tag of queue=default. However, the queue=default is an implicit tag that applies to the agent only if no queue tags are specified. If an agent is started with queue tags, it doesn't get the queue=default tag added on.

PR #982, opened 2021-03-25
Clarify default queue

plaindocs:
Fix for #980 

Thanks for raising @niceking !


niceking (approved):
LGTM! I wonder if it's worth also clarifying that the default queue tag doesn't show up in the UI? Not sure if we normally document our UI like that tho...

plaindocs:
Good point. Let's add it if folks start asking about that.

PR #983, opened 2021-03-26
Build and jobs states diagram and info

plaindocs:
Hey @scopegate 

Here are a few words and links. We definitely need the build states diagram as well when showing the two together.

I'm not 100% sold on the extra lists of progressions yet. 

Oh, and possibly, `accepted ->canceling` is not valid?

### Tables

Merging this will sneakily add [markdown table](https://michelf.ca/projects/php-markdown/extra/#table) support.


harrietgrace:
Such a good addition! 👯

What do you reckon about moving the Job states and Build states sections over into the Conditionals doc? When I went looking for it just now I was surprised to find it in the Notifications doc 🤔

Also @sj26 had some good additional context for the `broken` state, I'll paste it in here in case you'd like to include any of it:

> Jobs become "broken" when their configuration prevents them from running. This might be because their [branch configuration](https://buildkite.com/docs/pipelines/branch-configuration) doesn't match the build's branch, or because a [conditional](https://buildkite.com/docs/pipelines/conditionals) returned false. This is distinct from "skipped" jobs, which might happen if a newer build is started and build skipping is enabled. Broadly, jobs break because of something inside the build, and are skipped by something outside the build.

plaindocs:
@harrietgrace thanks for calling that out! How do you like this solution?

harrietgrace:
Love it! 🙌🏻

pages/pipelines/_job_states.md.erb
```diff
@@ -0,0 +1 @@
+Job state can be one of `pending`, `waiting`, `waiting_failed`, `blocked`, `blocked_failed`, `unblocked`, `unblocked_failed`, `limiting`,  `scheduled_limited`, `limited`, `scheduled`, `assigned`, `accepted`, `running`, `finished`, `canceling`, `canceled`, `timing_out`, `timed_out`, `skipped`, or  `broken`.
```
sj26, line 1: `scheduled_limited` is not a job state, afaict.

plaindocs:
Gawd, this thing is finally going out. 

@scopegate if you get a chance would you take a look at 

image

is it reachable at `<table class="no-formatting">`

PR #986, opened 2021-03-29
Fix rendering of bootstrap docs

pecigonzalo:
This is currently broken
image


plaindocs (approved):
Ooh, so it is, thanks for the fix.

I'll need to see if I can update the regex that generates that page from

```
Example:

   $ eval $(curl -s -H "Authorization: Bearer xxx" \
     "https://api.buildkite.com/v2/organizations/[org]/pipelines/[proj]/builds/[build]/jobs/[job]/env.txt" | sed 's/^/export /')
   $ buildkite-agent bootstrap --build-path builds
``` 

https://github.com/buildkite/docs/blob/f5dc7d412b3209f1fcbca191cc11d862baf0ddf8/scripts/cli2md.rb#L9

PR #993, opened 2021-04-06
301 redirect paths with trailing slashes

toolmantim:
In our ahrefs.com SEO site audit report it shows that we're internally linking to both https://buildkite.com/docs/pipelines and https://buildkite.com/docs/pipelines/ but don't set any canonical URL, so we're leaving up to search engines to figure out what the right URL is.

https://help.ahrefs.com/en/articles/2596742-duplicate-pages-without-canonical-error-in-site-audit

This adds a 301 redirect rule to ensure we only have a single URL for a unique page on the docs.


plaindocs:
Sounds good to me.

PR #996, opened 2021-04-06
Fix errors in linter

plaindocs:
Pronfanity has been there forever I think. Who lints the linters?

More importantly, the missing quotes around the regex do affect things badly, and leaves us an extra 25 "simply" type things to fix up. 

When we're done we can add the `--diff` back, so it only runs on changed files.

### DIFF
--- a/.alexrc.yml
+++ b/.alexrc.yml
@@ -19,4 +19,4 @@ allow:
   - executes
   - executed
   - invalid
-pronfanitySureness: 2
+profanitySureness: 2
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 The source files for the [Buildkite Documentation](https://buildkite.com/docs).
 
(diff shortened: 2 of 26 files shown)


StarfallProjects:
Removed the simplies, apart from one (emojis.md.erb contains a reference to an emoji named :one-does-not-simply:)

StarfallProjects:
@plaindocs have left some things in:

"Unlike artifact upload glob patterns, these operate over the entire path and not just between separator characters." - this could be changed to only, but that reads a little awkwardly I think? I wonder if we could add an exception for "not just"? As I assume the rule is designed to catch things like "Just do xyz".

"The token description should clearly identify the environment the token is intended to be used for,"  - legit use of clearly

plaindocs (commented):
Nice one, thanks @StarfallProjects !

(lol, I can't approve cos it is "my" PR)

PR #998, opened 2021-04-07
Add styling for Experimental/WIP feature callouts

scopegate:
The requirement here was a style to flag when a given feature was experimental/WIP.

Turns out there was an existing style there for pretty much exactly this use case — it wasn't in use anywhere, so I've refactored it very slightly to avoid a dependence on its headings being H1's, and have updated the existing example page ("Managing Log Output") to use this class instead.

image

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -272,7 +272,7 @@ a.Docs__example-repo {
     font-weight: bold;
     color: orange;
     &:before {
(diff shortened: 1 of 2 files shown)


plaindocs:
Oh sweet! Love the colour, but the road sign is giving me very slight flashbacks to Under constructions Geocities sites.

scopegate:
Yeah, I can see that... How about we replace it with 🛠?

PR #999, opened 2021-04-07
Add interactive sidebar

scopegate:
Resurrecting this with the latest sidebar content, and some updated logic to handle the deprecation of the v2 agent docs.

As a part of this, I've split the sidebar content out into its own component, but happy to be guided by anyone on a better way to do this if this doesn't feel like the right approach.

One other question that comes to mind: How do we want to approach top-level category links? We've had some feedback that categories on the homepage not linking feels a little weird. @keithpitt other than the links not being blue, how does this feel for you?


plaindocs:
I'm going to add landing pages for every category, so we can consistently link those titles.

plaindocs:
Logic looks solid to me.

I'm gonna totally bikeshed that green ,which looks odd to me. Might just be the newness though?

plaindocs:
@scopegate I'm happy to merge this if you are?

PR #1002, opened 2021-04-08
Proof of concept of attributes as yaml

plaindocs:
Following on from data as  Ruby constants in `models` this lets you source YAML files from `data` and iterate over them. 

Which would mean you don't need to add the id manually, but can generate it from the attribute name.

```html
    <tr id="build-path">
      <th><code>build-path</code></th>
      <td>
        Path to where the builds will run from.
        <p class="Docs__api-param-eg"><em>Default:</em> (depends on platform)</p>
        <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_BUILD_PATH</code></p>
      </td>
    </tr>
```

which by itself isn't worth making the change. 

But you could also change the html formatting at low effort. And include single attribute data in other places.

### DIFF
--- a/app/models/agent_attributes.rb
+++ b/app/models/agent_attributes.rb
@@ -1,0 +1,1 @@
+AGENT_ATTRIBUTES = YAML.load_file('data/agent_attributes.yaml').freeze
--- a/data/agent_attributes.yaml
+++ b/data/agent_attributes.yaml
@@ -1,0 +1,11 @@
+---
+- attribute: bootstrap-script
+  desc: |
+      <p>Command to invoke the bootstrap process.</p>
+      <p class="Docs__api-param-eg"><em>Default:</em> <code>"buildkite-agent bootstrap"</code></p>
+      <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_BOOTSTRAP_SCRIPT_PATH</code></p>
+- attribute: build-path
+  desc: |
+      <p>Path to where the builds will run from.</p>
+      <p class="Docs__api-param-eg"><em>Default:</em> (depends on platform)</p>
+      <p class="Docs__api-param-eg"><em>Environment variable:</em> <code>BUILDKITE_BUILD_PATH</code></p>
--- a/pages/agent/v3/configuration.md.erb
+++ b/pages/agent/v3/configuration.md.erb
@@ -32,6 +32,20 @@ _Required attributes:_
   </tbody>
 </table>
 
+_ Test attributes:_
+
+<table>
+  <tbody>
+    <% AGENT_ATTRIBUTES.each do |attr| %>
+      <tr id="<%= attr['attribute'] %>">
+        <th><code><%= attr['attribute'] %></code></th>
+        <td><%= attr['desc'] %></td>
+      </tr>
+      <% end %>
+  </tbody>
+</table>
+
+
 _Optional attributes:_
 
 <table>


plaindocs:
Urgh, I'd forgotten about the vale and erb inside html thing. 

```ini
BlockIgnores = (?s) *(<table class="erb".*?>.*?</table>)
```

and marking the tables as containing `erb` works. Or just wrapping in [ignore comments](https://github.com/errata-ai/vale/issues/210).

plaindocs:
@ticky do you have any thoughts on this being a terrible idea generally? From a rails point of view? Anything I'm doing wrong?

ticky:
Seems like a reasonable thing to do, yeah. I'd just double check this still works with the data extractor gear. 👍🏼

lastthyme (approved):

plaindocs:
You can do some simple yaml validation using https://www.npmjs.com/package/yaml-schema-validator#command-usage

```
npx schema validate -f data/agent_attributes.yaml -s data/agent_attributes.schema.yaml
```

plaindocs:
@StarfallProjects I don't know why it isn't running in the build, but `npm run validate-yaml` should work locally.

StarfallProjects:
Currently updating the YAML and saw that the current docs have recently started using an "introduced in" field. Is it worth adding that to the schema @plaindocs ?

plaindocs:
I'll need to tweak the YAML validator here.

plaindocs:
I fixed some of the linting stuff, but the long list of render markdown errors are I think that all `render_markdown` need to be `render_markdown partial`

plaindocs:
@lachlanhardy ok mate, I've cleaned up the merge conflict and the errors introduced by the update from main due to a change in the `partial` syntax.

Let's replace the two tables in https://buildkite.com/docs/agent/v3/configuration#configuration-settings (*Required* and *Optional*) with a single table, with visual indication of the *required* state. And remove the *test* table obv.

Looks like we didn't convert the deprecated settings to YAML, so we can ignore those for now.

Then, first week of January we can merge this thing it and start iterating on the details. Of how to make it sortable, visually more useful, and include better meta data from the structured data.

lachlanhardy:
First take on the table pushed. You can see it here: https://docs-review-pr-1002.herokuapp.com/docs/agent/v3/configuration#configuration-settings

lachlanhardy:
New version of the table: https://docs-review-pr-1002.herokuapp.com/docs/agent/v3/configuration#configuration-settings
table

plaindocs:
I really like this! A few things that spring to mind:

- I'd like the `required` tag to stand out more than the `optional` one. Or put a different way, I'd like that to be an obvious distinction not just a label.
- I think we also need to label the environment variables, 

Let's fix these and merge this. 

@karensawrey We're going to need to double check the quotes on some of these default values, which might have happened on the conversion to YAML.

karensawrey:
> I really like this! 

Same! Love it, amazing work, @lachlanhardy! I only have one small concern - while the required/optional tags look nice, green, and prominent in the screenshot above, they barely look like they are colored differently on a 4K screen in Chrome under Windows for me (tested on 2 different 4K screens).
viewin4k

> @karensawrey We're going to need to double check the quotes on some of these default values, which might have happened on the conversion to YAML.
Please clarify.

plaindocs:
@karensawrey is that a font width thing do you think?

image

karensawrey:
> @karensawrey is that a font width thing do you think?
> 
> image

Could be, but I'm running all defaults, no customization. I also sent @lachlanhardy a photo of the screen where it all looks more or less black, but that's for info only if it renders fine on most machines.

lachlanhardy:
I've revamped again. Viewable directly here: https://docs-review-pr-1002.herokuapp.com/docs/agent/v3/configuration#configuration-settings

image

@karensawrey I made that text larger, has that made a difference for you? 

@plaindocs I've broken required and optional out into separate sections (but still the same table so we keep the widths). Also labelled the env vars and adjusted some wrapping. What do you reckon?

plaindocs:
Looks great! If we tweak the template code to make up the experimental stuff, it should _just work_ right? 

image

plaindocs:
Works just fine

image

lachlanhardy:
Awesome!  You should just be able to stick anything inside the table cells and have them work.

Potentially, some elements might require margin adjustments to look right, but we can fix that easily if you see it.

Let me know if you want anything else on this, otherwise we can merge whenever. Tomorrow is the last day I can do anything for a bit because flights etc

lachlanhardy:
@karensawrey Awesome. Thanks to your feedback, I bumped up the size of that text a bit which is hopefully what made it more readable.

@plaindocs I think this is good to merge then. I'd prefer sooner than later in case we missed something, but I'll hold off until I hear confirmation from you.

lachlanhardy:
@plaindocs Ah, I've seen your internal post now. Yeah, happy to let it settle a bit. After six months, a few more days won't hurt. Hit me up via email if you need!

lachlanhardy:
@karensawrey That's good feedback, though, thank you! I am not much of a visual designer so I always want to hear if I've gone too far with something.

lachlanhardy:
🙌🏼

PR #1003, opened 2021-04-13
adding in missing documentation for using a key as part of a block step

EyeOfMidas:
While setting up dependencies for steps, I found this article which describes an undocumented feature of the block step.

https://forum.buildkite.community/t/step-dependencies-and-block/862

I added in a basic row to call this out


plaindocs:
Thanks @EyeOfMidas !

PR #1004, opened 2021-04-14
Adding an example using depends_on with block steps

EyeOfMidas:
Adding an example of using block _without_ input fields to set subsequent dependencies


plaindocs (approved):
Nice! Thank you.

PR #1006, opened 2021-04-15
Acquire job

StarfallProjects:
Resolves https://trello.com/c/GkxXC0ln/167-docs-for-acquire-job


plaindocs (approved):
Looks good to me. Might be worth using `stop` or `exit` but not both for the agent.

PR #1010, opened 2021-04-19
Update docs__note__heading elements from p to h3

scopegate:
@plaindocs I've gone and updated all of these Ps to H3s as we discussed. The CSS was targeting the classname and didn't mention the element type, so didn't even need to update the CSS 👍🏻


JuanitoFatas (approved):

plaindocs:
Cheers!

PR #1011, opened 2021-04-19
Add build.state to conditional list

harrietgrace:
It looks like `build.state` has been around a while as a conditional, no idea how I managed to leave it off the list 😅 😬

This PR adds the state conditional to the 'Supported Variables' section of the Using Conditionals page. It is already used in examples on the Notifications page.

image


plaindocs (approved):
LGTM

PR #1013, opened 2021-04-21
Remove pipeline `env`

unrealcloud:
A Pipeline doesn't seem to have the concept of `env`s. This field is returned in the API, but is not settable (thus always an empty object `{}`). I have tried with `POST` and `PUT`.

This field is not settable through the UI neither.

If this field is supposed to be settable, then it is an issue with the API!.


plaindocs:
Hey @unrealcloud thanks for the PR. @jayco and I are looking into exactly what is going on here, and what is supposed to be going on.

PR #1014, opened 2021-04-22
Update Algolia config to match new sidebar structure

scopegate:
When I pushed [the update to the sidebar](https://github.com/buildkite/docs/pull/999) last week, the changes in markup broke our Algolia config, and gave us useless categories that look like this:

image

This PR updates the selector that Algolia uses to determine the `lvl0` (ie, root category) for a given record, to match the current markup. Algolia should then update and go back to normal the next time its scraper runs (which happens ["every 24 hours"](https://docsearch.algolia.com/docs/scraper)).

I've also manually excluded the content categories that live on the new docs landing page, so they won't be indexed for search.

### DIFF
--- a/config/algolia.json
+++ b/config/algolia.json
@@ -8,7 +8,7 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": ".Docs__nav__section-heading:not(.Docs__nav__section-heading--link)",
+      "selector": ".Docs__nav__section-container.current > .Docs__nav__section-heading",
       "global": true,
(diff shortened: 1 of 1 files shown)


plaindocs:
Hey @scopegate thanks a ton for looking into this. 

As far as I can tell this doesn't fix the problem. The syntax looks like it should, but it is setting all level 0 headings to the default of `      "default_value": "Documentation"` in the test instance. 

I'm merging anyway, because it less broken than what we have.

PR #1015, opened 2021-04-22
Update docs on webhooks

JuanitoFatas:
Update the Webhook doc to include information about the new `X-Buildkite-Signature` header and how to verify signatures

### DIFF
--- a/pages/apis/webhooks.md.erb
+++ b/pages/apis/webhooks.md.erb
@@ -100,7 +100,7 @@ BuildkiteWebhook.valid?(
 A [replay attack](https://en.wikipedia.org/wiki/Replay_attack) is when an attacker intercepts a valid payload and its signature, then re-transmits them. To mitigate such attacks, Buildkite includes a timestamp in the `X-Buildkite-Signature` header. The timestamp is part of the signed payload, it is also verified by the signature, so an attacker cannot change the timestamp without invalidating the signature. To mitigate against a replay attack, the webhook receiver needs to:
 
 1. Verify the signature
-2. Check that the timestamp is within a reasonable timeframe (e.g. 5 minutes)
+2. Check that the timestamp is within a reasonable time frame (e.g. 5 minutes)
 
 ## Example implementations
 
--- a/pages/apis/webhooks.md.erb
+++ b/pages/apis/webhooks.md.erb
@@ -66,7 +66,7 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
 
 ## Verifying Webhook signatures
 
-The `X-Buildkite-Signature` header included in each signed event contains a timestamp and one signature. The timestamp is prefixed by `timestamp=`, and each signature is prefixed by `signature=`.
+The `X-Buildkite-Signature` header included in each signed event contains a timestamp and one signature. The timestamp is prefixed by `timestamp=`, and the signature is prefixed by `signature=`.
 
 Buildkite generates signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer represents UTC timestamp.
 
--- a/pages/apis/webhooks.md.erb
+++ b/pages/apis/webhooks.md.erb
@@ -38,7 +38,8 @@ The following HTTP headers are present in every webhook request, which allow you
 <table>
 <tbody>
   <tr><th><code>X-Buildkite-Event</code></th><td>The type of event<p class="Docs__api-param-eg"><em>Example:</em> <code>build.scheduled</code></p></td></tr>
-  <tr><th><code>X-Buildkite-Signature</code></th><td>The signature signed from your webhook secret. It's presence in the header of the webhook confirms that you are the originator of the the webhook request<p class="Docs__api-param-eg"><em>Example:</em> <code>timestamp=1619071700,signature=30222eb518dc3fb61ec9e64dd78d163f62c6134a63db768f1d40e0edbe6e43f0</code></p></td></tr>
+  <tr><th><code>X-Buildkite-Token</code></th><td>The token from your webhook configuration settings. It will only be sent if you selected the option to send the token in the token header. Its presence in the header of the webhook confirms that you are the originator of the the webhook request<p class="Docs__api-param-eg"><em>Example:</em> <code>309c9c842d8565adec5d7469160059f9</code></p></td></tr>
+  <tr><th><code>X-Buildkite-Signature</code></th><td>The signature signed from your webhook token. It will only be sent if you selected the option to send an HMAC signature in the signature header. Its presence in the header of the webhook confirms that you are the originator of the the webhook request<p class="Docs__api-param-eg"><em>Example:</em> <code>timestamp=1619071700,signature=30222eb518dc3fb61ec9e64dd78d163f62c6134a63db768f1d40e0edbe6e43f0</code></p></td></tr>
 </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,39 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+
+Buildkite generates signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer represents UTC timestamp.
+
+Example using Ruby:
+
+```ruby
+class BuildkiteWebhook
+  def self.valid?(payload, header, secret)
+    timestamp, signature = get_timestamp_and_signatures(header)
+    expected = OpenSSL::HMAC.hexdigest("sha256", secret, "#{timestamp}.#{payload}")
```
yob, line 79: I know this is only an example in a single language so it's not intended to be a copy-pastable thing folks can use, but maybe we want to include a `require 'openssl'` at the top?

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,39 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+BuildkiteWebhook.valid?(
+  request.body.read,
+  request.headers["X-Buildkite-Signature"],
+  ENV["BUILDKITE_WEBHOOK_SECRET"]
+)
+```
+
+### Preventing replay attacks
+
+A [replay attack](https://en.wikipedia.org/wiki/Replay_attack) is when an attacker intercepts a valid payload and its signature, then re-transmits them. To mitigate such attacks, Buildkite includes a timestamp in the `X-Buildkite-Signature` header. The timestamp is part of the signed payload, it is also verified by the signature, so an attacker cannot change the timestamp without invalidating the signature.
```
yob, line 98: To mitigate against a reply attack, does the webhook receiver need to:

1. verify the signature
2. check that the timestamp (which we can trust after #1) is within a reasonable timeframe (30s? 60s? 5 mins?)

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,39 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
   <p>Note that if a builds transitions between states very quickly, for example from blocked (`finished`) to unblocked (`running`), the webhook may be in a different state from the actual build. This is a known limitation of webhooks, in that they may represent a later version of the object than the one that triggered the event.</p>
 </div>
 
+## Verifying Webhook signatures
+
+The `X-Buildkite-Signature` header included in each signed event contains a timestamp and one signature. The timestamp is prefixed by `timestamp=`, and each signature is prefixed by `signature=`.
```
plaindocs, line 69:
```suggestion
The `X-Buildkite-Signature` header included in each signed event contains a timestamp and one signature. The timestamp is prefixed by `timestamp=`, and the signature is prefixed by `signature=`.
```

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,39 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+
+Buildkite generates signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer represents UTC timestamp.
```
plaindocs, line 71:
```suggestion
Buildkite generates signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer representation of a UTC timestamp.
```

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,39 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+
+Example using Ruby:
```
plaindocs, line 73: maybe state that this is an example of using Ruby to verify the signature and timestamp?

plaindocs (approved):
Nits aside this is all super clear @shevaun @JuanitoFatas , thanks!

pages/apis/webhooks.md.erb
shevaun: should i specify 5 minutes (or other timeframe) here?

yob: The approriate time frame might depend on the customers risk tolerance, but 5 minutes does seem like a pretty good suggestion to me

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,43 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
   <p>Note that if a builds transitions between states very quickly, for example from blocked (`finished`) to unblocked (`running`), the webhook may be in a different state from the actual build. This is a known limitation of webhooks, in that they may represent a later version of the object than the one that triggered the event.</p>
 </div>
 
+## Verifying Webhook signatures
+
+The `X-Buildkite-Signature` header included in each signed event contains a timestamp and one signature. The timestamp is prefixed by `timestamp=`, and the signature is prefixed by `signature=`.
+
+Buildkite generates signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer representation of a UTC timestamp.
```
shevaun, line 71:
```suggestion
Buildkite generates the signature using a hash-based message authentication code ([HMAC](https://en.wikipedia.org/wiki/HMAC)) with [SHA-256](https://en.wikipedia.org/wiki/SHA-2). The timestamp is an integer representation of a UTC timestamp.
```

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,43 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+  request.body.read,
+  request.headers["X-Buildkite-Signature"],
+  ENV["BUILDKITE_WEBHOOK_SECRET"]
+)
+```
+
+### Preventing replay attacks
+
+A [replay attack](https://en.wikipedia.org/wiki/Replay_attack) is when an attacker intercepts a valid payload and its signature, then re-transmits them. To mitigate such attacks, Buildkite includes a timestamp in the `X-Buildkite-Signature` header. The timestamp is part of the signed payload, it is also verified by the signature, so an attacker cannot change the timestamp without invalidating the signature. To mitigate against a replay attack, the webhook receiver needs to:
+1. Verify the signature
```
shevaun, line 101:
```suggestion

1. Verify the signature
```

pages/apis/webhooks.md.erb
```diff
@@ -63,6 +64,44 @@ Each event’s data is sent JSON encoded in the request body. See each event’s
...
+  ENV["BUILDKITE_WEBHOOK_SECRET"]
+)
+```
+
+### Preventing replay attacks
+
+A [replay attack](https://en.wikipedia.org/wiki/Replay_attack) is when an attacker intercepts a valid payload and its signature, then re-transmits them. To mitigate such attacks, Buildkite includes a timestamp in the `X-Buildkite-Signature` header. The timestamp is part of the signed payload, it is also verified by the signature, so an attacker cannot change the timestamp without invalidating the signature. To mitigate against a replay attack, the webhook receiver needs to:
+
+1. Verify the signature
+2. Check that the timestamp is within a reasonable timeframe (e.g. 5 minutes)
```
yob, line 103:
```suggestion
2. Check that the timestamp is within a reasonable time frame (e.g. 5 minutes)
```

shevaun:
`buildkite-console:002> Feature::SIGNED_WEBHOOK.activate_for_all_users
=> true
`

On for everyone now, so merging the docs

PR #1016, opened 2021-04-22
Fix table overflow

JuanitoFatas:
|Before|After|
|------|------|
|CleanShot 2021-04-22 at 16 14 37@2x | CleanShot 2021-04-22 at 16 14 51@2x |


plaindocs (approved):
Thank you!

PR #1019, opened 2021-04-22
Fixing minor typo in docs

fhightower:
I believe the sense of the documentation could be improved by this PR.

I know there is debate whether to use singular or plural verbs when referring to a company's actions, but according to the sources cited [here](https://www.grammarphobia.com/blog/2018/01/it-or-they.html), British English allows either singular or plural and American English uses singular verbs "unless you wish to emphasize the individuals who make up the organization; in that case, use the plural" (quoted [here](https://www.grammarphobia.com/blog/2018/01/it-or-they.html)).


plaindocs (approved):
Thanks for caring, and for the PR! <3

PR #1021, opened 2021-04-22
Tiny typo fix

karensawrey:
Also trying to see if I can commit after the renaming of the main branch.


plaindocs:
Looks good to me. You need to also update https://github.com/buildkite/docs/blob/main/spec/features/reading_pages_spec.rb#L18 I think to make the test aware of the change.

PR #1022, opened 2021-04-22
Fixes for botched code blocks

karensawrey:


pages/pipelines/permissions.md.erb
```diff
@@ -85,9 +86,10 @@ query FindOrganizationMember {
     }
   }
 }
-```  
+```
```
plaindocs, line 89: If you were being really picky you could delete the trailing spaces also.

karensawrey: Done. As I AM picky ;)

PR #1023, opened 2021-04-26
Algolia fix, part two: Add class to each sidebar container based on path match

scopegate:
The selector I used for the fix from last week in #1014 worked fine client-side, but the logic to add the "current" class to the relevant sidebar content was in javascript, so wasn't getting triggered by Algolia's crawler. 

This PR shifts that logic and to happen server-side, which makes the markup a little more complex, but should resolve the Algolia issue properly and let it determine categories again.

### DIFF
--- a/app/views/layouts/_sidebar.html.erb
+++ b/app/views/layouts/_sidebar.html.erb
@@ -1,6 +1,7 @@
+<% current_section = request.original_fullpath.split('/')[2] %>
 <nav class="Docs__nav">
   <div class="Docs__nav__sections">
-    <div class="Docs__nav__section-container" data-docs-path="tutorials">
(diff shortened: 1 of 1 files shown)


plaindocs (commented):
@scopegate you're a star!

This really should work, but still seems not to. 

```
$ curl -s http://localhost:3000/docs/integrations/github | htmlq '.Docs__nav__section-container.current > .Docs__nav__section-heading'
<p class="Docs__nav__section-heading">Integrations</p>
```

image

scopegate:
@plaindocs just catching up on this - those results look pretty much as I'd expect, in that the query selector from the first result returns the expected element without any javascript being run. And if the screenshot is from our existing Algolia db, I wouldn't expect that to update until we merge this PR to the production docs environment.

Unless I'm missing something (which is entirely possible), I reckon we merge this if we're happy with the sidebar behaviour, and then keep an eye on Algolia as its crawler updates and see what results we get.

PR #1028, opened 2021-05-03
Add notification events

plaindocs:
- [x] Adding a section like this to each notification type:

image

- [x] Add PagerDuty

- [x] Tweaked the rest of the page.


yob (commented):
Great! I always found it difficult to wrap my head around which events apply to each type of notification.

The changes look accurate to me, but I think an additional review from someone else more familiar with build notifications would be wise.

/cc @eleanorakh  @sj26

PR #1031, opened 2021-05-04
Library plugin visibility & additional plugin docs updates

StarfallProjects:


pages/agent/v3/hooks.md.erb
```diff
@@ -68,7 +68,7 @@ The following is a complete list of available hooks, and the order in which they
     <tr>
       <td style="white-space: nowrap"><code>command</code></td>
       <td style="white-space: nowrap">Plugin, Repository, Agent</td>
-      <td>Overrides the default command running behavior. <em>Note:</em> If multiple command hooks are found, only the first will be run.</td></tr>
+      <td>Overrides the default command running behavior. If multiple command hooks are found, only the first will be run. Each pipeline must have a command hook available in at least one of the locations (plugin, repository, or agent)</td></tr>
```
plaindocs, line 71: Clarifying this is :+1: :+1: 

> Each pipeline must have a command hook available in at least one of the locations (plugin, repository, or agent

But there is something niggling at me with the sentence. How about

```suggestion
      <td>Overrides the default command running behavior. If multiple command hooks are found, only the first will be run. A pipeline must have a command hook defined defined in the plugin, the repository or the agent.</td></tr>
```

Although, you can just define the command in the pipeline as well, 

```suggestion
      <td>Overrides the default command running behavior. If multiple command hooks are found, only the first will be run. A pipeline must have a command hook defined in one or more of the pipeline, the plugin hook, the repository hook, or the agent hook.</td></tr>
```

pages/pipelines/defining_steps.md.erb
```diff
@@ -60,7 +60,7 @@ Starting your YAML with the `steps` object, you can add as many steps as you req
 
 Before getting started with a `pipeline.yml` file, you'll need to tell Buildkite where it will be able to find your steps.
 
-In the YAML steps editor, add the following YAML:
+In the YAML steps editor in Buildkite, add the following YAML:
```
plaindocs, line 63: We don't do a good job of documenting how to talk about the Buildkite dashboard, but it is better than "in Buildkite"

pages/plugins/writing.md.erb
```diff
@@ -15,17 +17,17 @@ steps:
 ```
 
-## Create a new git repository
+### Create a new git repository
```
plaindocs, line 20: I'm not sure we gain a lot by nesting everything as H3, other than moving it out of the in page TOC, which doesn't seem great.

plaindocs: If it is to avoid clashing with the design topic, we could number the tutorial steps?

pages/plugins/writing.md.erb
```diff
@@ -208,12 +210,19 @@ You can now run the tests using the following command:
...
+### Add a readme
 
 Next we'll add a `README.md` file to introduce the plugin to the world:
 
 <%= render 'plugins/tutorial_readme' %>
 
-## Publish to the directory
+## Publish to the Buildkite Plugin Directory
+
+To add your plugin to the [Buildkite Plugin Directory](https://buildkite.com/plugins), publish your repository to a public GitHub repository and add the `buildkite-plugin` repository topic tag. For full instructions, see the [Plugin Directory documentation](/docs/plugins/directory).
```
plaindocs, line 221: :+1:

plaindocs (commented):
Looks great. I've left a few comments inline.

pages/plugins/writing.md.erb
StarfallProjects: I nested them because they're subsections of the tutorial. Can make them steps instead I guess (will work visually, not so nice from an HTML hierarchy pov)

pages/agent/v3/hooks.md.erb
StarfallProjects: As I understand it, in the pipeline, you could say which command to run, but it would look for the command in either a plugin, agent, or the repo? You don't actually write commands in the pipeline yaml? And there's no pipeline directory setup with command files that I just haven't come across?

pages/pipelines/defining_steps.md.erb
StarfallProjects: What do you suggest? "In the Buildkite dashboard"?

pages/agent/v3/hooks.md.erb
StarfallProjects: I'm also really not sure "defined" is the right word? That suggests you set it in a config/describe it somewhere? But you literally just stick a command script in the hooks directory?

plaindocs (approved):
:shipit:

PR #1032, opened 2021-05-04
add bitbucket server to the list on source_control

StarfallProjects:
I missed a menu when adding the bitbucket server docs: https://buildkite.com/docs/integrations/source-control


plaindocs (approved):

plaindocs:
Yup - good candidate for pulling out to data.

PR #1036, opened 2021-05-06
Add JSON-specific depends_on documentation

wesmangum:
Include messaging about using `null` in JSON files instead of `~`, which is YAML specific.

### DIFF
--- a/pages/pipelines/dependencies.md.erb
+++ b/pages/pipelines/dependencies.md.erb
@@ -86,6 +86,8 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks @wesmangum , I did not know this. 

What do you think about rewording the previous paragraph to something like 

```
To ensure that a step is not dependent on any other step, add an explicit empty dependency with the `~` character (YAML) or `null` (JSON). This also ensures that the step will run immediately at the start of the build:
```

It is a bit less jarring than the note.

wesmangum:
Thanks for the suggestion @plaindocs !

plaindocs (approved):
Cheers! :-)

Issue #1012, opened 2021-04-21
REST API: querying by multiple commit sha's is supported 


yob:
The [docs for listing builds via the REST API](https://buildkite.com/docs/apis/rest-api/builds) say:

>  commit - Filters the results by the commit (only works for full sha, not for shortened ones). 
> Example: ?commit=long-hash

It's also possible to query for multiple commits by using the rails array syntax:

https://api.buildkite.com/v2/organizations/<org-slug>/builds?commit[]=sha2&commit[]=sha1`

This might be an accidental implementation, but it works and is useful.

PR #1040, opened 2021-05-12
Multi sha

StarfallProjects:
Resolves https://github.com/buildkite/docs/issues/1012


plaindocs (approved):
Awesome!

Do you think `commit=long-hash` would be better with a real example?

PR #1043, opened 2021-05-14
Update cli_annotate.md.erb

jsharpe:
Commonmark img tags don't render if they are closed.


plaindocs:
Thanks @jsharpe - do you have a source for that?

jsharpe:
It was based on trying the code in the docs on Buildkite and realising that it only rendered as an image if you removed the trailing slash.

plaindocs (approved):

PR #1049, opened 2021-05-24
Update GHEnterprise Firewalled Installation IP

amukherjeetwilio:
The docs for GitHub Enterprise Firewalled Installations referenced one
specific IP address, which I think is the old egress IP address from
before [the new IPs](https://buildkite.com/changelog/118-new-outbound-ip-addresses) were
allocated.

This PR adds the new (current) IPs, as well as a reference to where one
can find the updated list (the meta API)


plaindocs:
Thanks so much for spotting this @amukherjeetwilio and for suggesting a fix.
I've removed the list of IPs to encourage people to look them up from the API directly.

PR #1055, opened 2021-05-31
Mention Datadog metric unit

plasticine:
These are sent in milliseconds according to the agent codebase, however without trawling through a bunch of Go this is not very clear.


plaindocs (approved):
Good point, thanks!

PR #1057, opened 2021-06-02
Mobile experience improvements (and other Google lighthouse fixups)

toolmantim:
This PR cleans up a number of warnings from Google Lighthouse that I spotted when viewing the Google crawler warnings for our site via Google Search Console.

The biggest change is that it improves the navigation on mobile, so it isn't as large and expanded by default… which means people can get to the documentation they're reading much faster (especially when navigating between pages, which would involving scrolling past the navigation each time).

| Before | After |
| -- | -- |
| image | image |

It also makes sure the up arrow is available visible in the nav when it's expanded, showing you can collapse it and hinting that the menu is exapando/collapsable for non-hover devices too (like touch screens).

image

It also adds the word "Buildkite" back into the top nav, so "Buildkite Docs" instead of just "Docs"

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -3,11 +3,13 @@ $color-green: rgb(3, 165, 98);
 
 .Docs__page-container {
   padding-top: 20px;
+  @media (max-width: $Docs__bp_stacked) { padding: 0; }
 }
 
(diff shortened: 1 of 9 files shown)


plaindocs (approved):
Love it.

scopegate (approved):
🚢🎉

PR #1064, opened 2021-06-08
Update agent help script to run on macOS

keithduncan:
I’ve added support for invoking the script outside the script directory so that it still finds the `cli2md.rb` script and updated the `sed` arguments so it works on macOS.

@plaindocs I’ll have to lean you to check what options sed takes on Linux to make sure this is compatible


plaindocs:
Cheers, nice update.

Looks like what you have doesn't work for me, but either of these do. 

```bash
  sed -Ei -e 's/^##/###/' "$file"
  sed -i'' -e 's/^##/###/' "$file"
```

keithduncan:
Looks like neither of those work on macOS either 😅

I’ve switched it to use a `.bak` extension and ignored that extension in git, let me know what you think 😄

plaindocs (approved):
Nice one. Yeah, ignore the bak was going to be my next suggestion, I've fixed the silly linter in a separate PR.

PR #1072, opened 2021-06-14
Document the new unarchive pipeline REST API endpoint

shevaun:
Implementation here: https://github.com/buildkite/buildkite/pull/6627

### DIFF
--- a/pages/apis/rest_api/pipelines.md.erb
+++ b/pages/apis/rest_api/pipelines.md.erb
@@ -822,7 +822,7 @@ Error responses:
 
 ## Unarchive a pipeline
 
-Restore an archived pipeline.
+Unarchived pipelines are editable, and are shown on the Pipeline pages.
 
 ```bash
 curl -X POST "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{slug}/unarchive"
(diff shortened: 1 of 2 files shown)


pages/apis/rest_api/pipelines.md.erb
```diff
@@ -820,6 +820,88 @@ Error responses:
 </tbody>
 </table>
 
+## Unarchive a pipeline
+
+Restore an archived pipeline.
```
plaindocs, line 825: Unless restore has more meaning than unarchive, lets stick to the same wording so folks don't worry if there is a difference. And it maybe makes sense to reword the description as well

> Unarchived pipeles are editable, and are shown on the Pipeline pages.

plaindocs: Otherwise, great stuff!

shevaun:
@plaindocs not sure what to do with this - the linting check does not like `unarchive` apparently 😅

JuanitoFatas:
> @plaindocs not sure what to do with this - the linting check does not like `unarchive` apparently 😅

Add `unarchived` after https://github.com/buildkite/docs/blob/f09b3e5ded6a9b1e403e0e78469e2c291846c041/vale/vocab.txt#L123 should work

plaindocs (approved):
Awesome! Thanks for kicking the linter @JuanitoFatas !

PR #1073, opened 2021-06-15
Remove 'experiment' from output redactor docs

keithduncan:
This will land as default feature in v3.31 so we can remove the experimental caveats from these docs.

@plaindocs should we note the agent minimum version required for the feature anywhere?

### DIFF
--- a/pages/agent/v3.md.erb
+++ b/pages/agent/v3.md.erb
@@ -74,11 +74,6 @@ Please note that there is every chance we will remove or change these experiment
 
 The Buildkite agent can redact strings that match the value of environment variables whose names match common patterns for passwords and other secure information before the build log is uploaded to Buildkite.
 
-<div class="Docs__wip-note">
-  <p class="Docs__wip-heading">Experimental feature</p>
-  <p>To use it, set <code>experiment="output-redactor"</code> in your <a href="/docs/agent/v3/configuration#experiment"> agent configuration</a>.</p>
-</div>
-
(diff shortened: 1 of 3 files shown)


plaindocs:
This is great!

We've not done so much in the past, but especially where the behavior changes like this we should.
The real question is whether we should document both states? I guess because it is experimental we don't commit to that? 
So yeah, :+1: to `introduced in Agent v3.31`

plaindocs:
@StarfallProjects this will affect #1002

keithduncan:
> I guess because it is experimental we don't commit to that?

🤔 I could see including a link to the release that the experiment is available from too, if we’re documenting the experiment it’s probably worth pointing folks to which version they’d need.

> So yeah, 👍 to `introduced in Agent v3.31`

I’ve tried putting this in a new `<p>` tag after the description to see how it looks in the preview deploy, let me know what you think.

keithduncan:
Screenshot 2021-06-16 at 09 18 51

toolmantim:
I wonder if we could put it underneath the name of the config on a new line, rather than in the description? Perhaps just "v3.31 and above" in dark grey?

plaindocs:
@scopegate wanna factor this information design into the parameters rework in #1002 ? 

@toolmantim do you see that needing an introduction of that field for all backwards params? Or starting now for all going forward?

keithduncan:
3.31.0 is live now so this is ready for final review 🙌

plaindocs (approved):
@keithduncan happy for this to be merged and work on the version tracking stuff separately.

PR #1076, opened 2021-06-17
Update screenshot with latest syntax

harrietgrace:
I noticed that the screenshot in the conditionals doc was using the old style conditional syntax (env vars) and have updated it to use the newer style (in-built values).

This PR only touches that one screenshot 😊  

Before:
image

After:
image

A screenshot of a screenshot! ✨


plaindocs (approved):
SO META!

image

PR #1078, opened 2021-06-18
Less noisy CSP

sj26:
We're getting _lots_ of CSP reports about docs pages containing inline style attributes, like `<... style="...">`. I'm not exactly sure why. It might be because it's falling back to `default-src`, so I'm proposing removing the default here as a trial. It might also be because once you add a nonce the browser expects all things to have nonces and ignores `unsafe-inline`. Let's find out?

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -17,13 +17,8 @@
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
 Rails.application.config.content_security_policy do |policy|
-  policy.default_src    :self
-  policy.font_src       :self, "https://www2.buildkiteassets.com/"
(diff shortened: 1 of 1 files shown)


matthewd:
> It might also be because once you add a nonce the browser expects all things to have nonces and ignores unsafe-inline

Yeah, I believe it's that one.

sj26:
Yeah neat, let's try adding an explicit style-src-attr instead then.

sj26:
An example violation report:

image

sj26:
After this change, I still get an error in the Safari I'm using:

> [Report Only] Refused to apply a stylesheet because its hash, its nonce, or 'unsafe-inline' does not appear in the style-src directive of the Content Security Policy.

sj26:
I think maybe bang-for-buck we're best off focusing on object and script usage, and ignoring most everything else.

sj26:
This might be a bit of a fools errand — algolia also needs `eval`, so the current script-src is violated regularly.

sj26:
I'm declaring too-hard and giving up for now 😅

PR #1083, opened 2021-06-24
Update agent OS support policy

keithduncan:
This changes the docs for each operating system (I only changed the agent v3 docs and didn’t touch the v2 ones) to steer folks towards which operating systems we support rather than the oldest operating system the agent can be installed on.

This doesn’t practically change where the agent can be installed, e.g. it probably still works on Windows XP we’re just saying that if it _happens to_ we can’t guarantee support when it goes wrong as we have to focus our efforts.

For Debian, Red Hat, and Ubuntu we can also drop the docs for using Upstart, as presently in service and supported versions of these all use systemd 🎉

### DIFF
--- a/pages/agent/v3/windows.md.erb
+++ b/pages/agent/v3/windows.md.erb
@@ -1,6 +1,6 @@
 # Installing Buildkite Agent on Windows
 
-The Buildkite Agent is supported on Windows Server 2012 and newer.
+The Buildkite Agent is supported on Windows 8, Windows Server 2012, and newer.
 
 {:toc}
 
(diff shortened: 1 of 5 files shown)


yob:
> For Debian, Red Hat, and Ubuntu we can also drop the docs for using Upstart, as presently in service and supported versions of these all use systemd

the best

pages/agent/v3/windows.md.erb
```diff
@@ -1,14 +1,14 @@
 # Installing Buildkite Agent on Windows
 
-The Buildkite Agent can be installed on both 64bit and 32bit editions of Windows XP and later.
+The Buildkite Agent is supported on Windows Server 2012 and newer.
```
yob, line 3: Is it worth mentioning the equivalent consumer windows version as well

yob (approved):
Looks technically accurate to me, but probably best to wait for a review from a docs person as well :+1:

plaindocs (approved):
Love it.

PR #1089, opened 2021-06-28
update keyservers

bobzoller:
sks-keyservers.net is dead, long live keys.openpgp.org

https://sks-keyservers.net/
image

### DIFF
--- a/pages/agent/v2/debian.md.erb
+++ b/pages/agent/v2/debian.md.erb
@@ -30,7 +30,7 @@ sudo apt-get install -y apt-transport-https dirmngr
 Now you can add our signed apt repository:
 
 ```shell
-sudo apt-key adv --keyserver ipv4.pool.sks-keyservers.net --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198
+sudo apt-key adv --keyserver keys.openpgp.org --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198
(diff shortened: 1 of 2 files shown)


yob (commented):
Thanks, I've confirmed the proposed changes work :+1: 

```
$ sudo apt-key adv --keyserver keys.openpgp.org --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
Executing: /tmp/apt-key-gpghome.yXInvOOoPW/gpg.1.sh --keyserver keys.openpgp.org --recv-keys 32A37959C2FA5C3C99EFBC32A79206696452D198
gpg: key A79206696452D198: "Buildkite <dev@buildkite.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
```

The deprecation warning is interesting. From the man page:

```
Use of apt-key is deprecated, except for the use of apt-key del in maintainer scripts to remove existing keys from the main keyring. If such usage of apt-key is desired the additional installation of the GNU Privacy Guard suite (packaged in gnupg) is required.

apt-key(8) will last be available in Debian 11 and Ubuntu 22.04.
```
...and:

```
Note: Instead of using this command a keyring should be placed directly in the /etc/apt/trusted.gpg.d/ directory with a descriptive name and either "gpg" or "asc" as file extension.
```

Seems like something we should follow up. We'll need to check if the new approach works in all supported debian/ubuntu versions though, and in the interim this change will keep things working for folks :+1:

plaindocs:
Thanks @bobzoller @yob :+1:

PR #1097, opened 2021-07-14
Update secretmanager docs to include adding permission for the scaling lambda role

keithduncan:
I missed this in my first pass on these because I used a static auto scale group of one instance and only tested that the agent booted. To allow the scaling Lambda to poll the Buildkite API when using a token in Secrets Manager we also need to grant permission to the Lambda execution role.

cc @lox


plaindocs (approved):
:+1: Cheers!

PR #1100, opened 2021-07-15
Osx -> macos

plaindocs:
- [x] Needs #1099 to be merged first.

This PR changes the v3 `osx.md.erb` page to `macos.md.erb` and updates links and routes.

PR #1102, opened 2021-07-21
Osx -> macos

plaindocs:
Change OSX to MACOS (ignoring Agent v2) 

Replaces #1100 which I broke the conflict resolution on.


keithduncan (approved):
Looks good to me!

PR #1109, opened 2021-08-03
Remove posthog and matomo

yob:
Historically we've used these for visitor analytics, but not any more


pda (approved):
Looks good, I haven't run the code though.

Do we have a plan to remove the environment variables from production?
(Or, is that part of the infrastructure tear-down?)

sj26 (approved):

plaindocs (approved):
Looks fine to me.

PR #1112, opened 2021-08-03
updated details on running locally

StarfallProjects:
Added a little more detail, to flag Ruby versions and Node requirement.

Have also spelled out need to be in docs directory - possibly too basic for our users though?

Also: I never got it running on Windows directly, had to resort to WSL. Wondering if we should flag this? You might get another windows user in a year or two . . .

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,12 +8,14 @@ To contribute, send a pull request! :heart:
 
 ```bash
 git clone https://github.com/buildkite/docs.git
+cd docs
 git submodule update --init
 ```
 
 If you have Ruby installed:
 
 ```bash
+# Navigate into the docs directory
 # Install the dependencies
 bundle
 # Run the specs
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +24,8 @@ bundle exec rspec
 bin/rails server
 ```
 
+> **Note**: The documentation uses Ruby 2.7.2. You also need Node installed. The current LTS (long term support) version should be ok.
```
plaindocs, line 27: How about pointing to https://github.com/buildkite/docs/blob/main/.ruby-version for the ruby version instead? :p

plaindocs (approved):

README.md
StarfallProjects: Yeah I was on the fence between friendly and maintainable. I shall edit.

plaindocs: Write a custom ssg to insert the value at commit time?

PR #1113, opened 2021-08-04
Differentiate agent lifecycle vs job lifecycle hooks

keithduncan:
Presently, the term **agent hooks** can be used to refer to either the _buildkite-agent feature_ or the _location that hooks are stored in the file system_.

With the new `pre-bootstrap` and arguably `agent-shutdown` hooks we have introduced a new _category_ of hook, one that is run outside the step/job lifecycle and that is _executed_ rather than _sourced_.

I’d like to make that distinction clear and think introducing new terminology is the way to do that.

This feels like a significant change to the product area. ~I’m **not** tied to this particular terminology or this particular slicing of things, but I wanted to have something written down for us to discuss pros and cons 😄~ Edit: I have since discussed this with others and am fairly tied to this new naming. How I’ve named and sliced this up is elaborated on 👇 and the doc diff reflects this.

---

This change continues to use the term **agent hooks** to refer to the all encompassing **feature** of the Buildkite Agent.

There are **two categories of agent hook**: Agent Lifecycle Hooks and Job Lifecycle Hooks.

- Agent Lifecycle Hooks are **executed** as part of the agent lifecycle, presently `pre-bootstrap` and `agent-shutdown`.
- Job Lifecycle Hooks (the _new name_ for the _old feature_) are **sourced** as part of the bootstrap lifecycle, all our old favourites `environment`, `pre-command`, `post-artifact` etc

There are **three locations that agent hooks are defined**:

- Agent hooks (the _new name_ for the _old location_) are stored on the agent host. This location can define **both** agent lifecycle hooks, and job lifecycle hooks.
- Repository hooks, as before, can define job lifecycle hooks
- Plugin hooks, as before can define job lifecycle hooks

---

Next steps:

- [x] Update strict checks example to use `BUILDKITE_ENV_FILE` for vetting rather than env vars
- [x] Reach consensus on naming
- [x] Find other places to update naming

### DIFF
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -119,11 +119,12 @@ they are run as part of each job:
 
 ## Creating Lifecycle hooks
 
-Lifecycle hooks are executed and can be written in any programming language of
-your choice. See the specific lifecycle hook documentation for details on the
-interface between them and the Buildkite Agent.
+Lifecycle hooks can be written in the programming language of your choice and
+are executed by the Buildkite Agent. See the documentation for each lifecycle
+hook for details on the interface between the Buildkite Agent and the
+Lifecycle hook.
 
-Lifecycle hooks can only be defined in the [global hooks](#agent-hook-locations)
+Lifecycle hooks can only be defined in the [global hooks](#agent-hook-locations-global-hooks)
 directory.
 
 ## Creating Job hooks
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -30,7 +30,7 @@ environment variables are carried to the job’s subsequent phases and hooks. Fo
 example, the `environment` hook can modify or export new environment variables
 for the job’s subsequent phases and hooks.
 
-## Available Lifecycle hooks
+## Lifecycle hooks
 
 <table>
   <thead>
@@ -60,7 +60,7 @@ for the job’s subsequent phases and hooks.
   </tbody>
 </table>
 
-## Available Job hooks
+## Job hooks
 
 The following is a complete list of available job hooks, and the order in which
 they are run as part of each job:
@@ -191,9 +191,8 @@ platform’s installation documentation.
 To get started with global hooks copy the relevant example script and remove the
 `.sample` file extension.
 
-See the lists of [lifecycle hooks](#available-lifecycle-hooks) and
-[job hooks](#available-job-hooks) for the hook types that can be included in the
-global hooks directory.
+See [lifecycle hooks](#lifecycle-hooks) and [job hooks](#job-hooks) for the hook
+types that can be defined in the global hooks directory.
 
 ### Repository hooks
 
@@ -205,8 +204,8 @@ To get started, create a shell script in `.buildkite/hooks` named
 `post-checkout`. It will be sourced and run after your repository has been
 checked out as part of every job for any pipeline that uses this repository.
 
-You can include any of the [available job hooks](#available-job-hooks) whose
-`Order` includes *Repository*.
+You can define any of the [job hooks](#job-hooks) whose `Order` includes
+*Repository*.
 
 ### Plugin hooks
 
@@ -214,5 +213,4 @@ Plugin hooks allow plugins you’ve defined in your Pipeline Steps to override
 default behavior.
 
 See the [plugin documentation](/docs/plugins) for how to implement plugin hooks
-and [available job hooks](#available-job-hooks) for the list of hook types that
-a plugin can include.
+and [job hooks](#job-hooks) for the list of hook types that a plugin can define.
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -1,40 +1,89 @@
 # Buildkite Agent Hooks
 
-Every agent installer comes with a hooks directory which can be used to override and extend the built-in agent behavior. For example, you could create an agent `checkout` hook which speeds up a fresh `git clone` on a new build machine, an agent `environment` hook which exports secret API keys as environment variables, or a repository `pre-command` hook which sets up repository-specific environment variables.
+Every agent installer comes with a hooks directory which can be used to override
+and extend the built-in agent behavior. For example, you could create an agent
+`checkout` hook which speeds up a fresh `git clone` on a new build machine, an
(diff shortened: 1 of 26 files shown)


pages/agent/v3/hooks.md.erb
```diff
@@ -1,40 +1,69 @@
...
+<table>
+  <thead>
+    <tr>
+        <th>Hook</th>
+        <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td style="white-space: nowrap"><code>pre-bootstrap</code></td>
```
plaindocs, line 44: The description of this reads like it should be called "strict check" hook, not "pre-bootstrap"? Are there any other envisaged use cases?

pages/agent/v3/hooks.md.erb
```diff
@@ -1,40 +1,69 @@
...
-It will be sourced and run every build, after the Agent checks your repository out.
-You can also use one of the other available *Repository* hooks, according to when you want your custom script to run.
+## Lifecycle hooks
 
-### Plugin hooks
-
-Plugin hooks allow plugins you’ve defined in your Pipeline to override default behavior.
-
-See the [plugin documentation](/docs/plugins) for how to implement plugin hooks.
+<table>
```
plaindocs, line 35: I work err on the side of including a mostly redundant "Order" column in this table, to make it clear that both are Global hooks, in the same format as the Job hook table.

pages/agent/v3/hooks.md.erb
```diff
@@ -128,3 +168,50 @@ An example of a windows `environment.bat` hook:
 ECHO "--- \:house_with_garden\: Setting up the environment"
 SET GITHUB_RELEASE_ACCESS_KEY='xxx'
 ```
+
+## Agent Hook Locations
```
plaindocs, line 172: I found the mapping slightly hard to parse here, although don't yet have a better suggestion.

plaindocs (changes requested):
This is great stuff. Did you discuss "types" as an alternative to "categories", and was it decided against *because* of the multiple hooks in Lifecycle hooks? 

You've got the following H2 headings

```
    Categories of hooks
    Lifecycle hooks
    Job hooks
    Creating Lifecycle hooks
    Creating Job hooks
    Job hooks on Windows
    Agent Hook Locations
```

How do you feel compacting that down to something like 

```
    Categories of hooks -> move this into the intro hooks
    Lifecycle hooks (and Creating Lifecycle hooks)  (you'd probs need to include some text in both places    Job hooks (Creating Job hooks) (you'd probs need to include some text in both places
    Job hooks on Windows
    Agent Hook Locations
```

It _might_ be easier to parse. I'll have another look at this later on.

I'd also like a super short section on "in July 2021 we slightly changed how we refer to Agent hooks, although we could fold that into the changelog (a better approach or even both).

pages/agent/v3/hooks.md.erb
keithduncan: We covered the naming a bit in the [Basecamp Message Board post](https://3.basecamp.com/3453178/buckets/11763568/messages/3899813064) about the feature and I admit I’d be hesitant to redo the naming in a new agent release. Though there aren’t any other envisaged use cases right now, I do think the naming fits in with how the rest of our hooks are named.

pages/agent/v3/hooks.md.erb
keithduncan: I spent a while thinking about this and I think it’s the right call :+1: What do you think about naming both columns "Location Order"? I’m going to make that change and we can always revert it 😄

pages/agent/v3/hooks.md.erb
keithduncan: I’ve tweaked this a bit in my local changes to be pushed up, let me know what you think when I push it up

keithduncan:
> This is great stuff. Did you discuss "types" as an alternative to "categories", and was it decided against because of the multiple hooks in Lifecycle hooks?

I went with category because we already use type to refer to the existing collection of hooks and I wanted to preserve that while introducing a new grouping. This way we have two categories, and in those two categories we have multiple types. What do you think?

> How do you feel compacting that down to something like

I think these suggested header changes are great, I’ll incorporate those now 😄 

> I'd also like a super short section on "in July 2021 we slightly changed how we refer to Agent hooks, although we could fold that into the changelog (a better approach or even both).

I think both is good, I’ll work that into the intro too :+1:

keithduncan:
What do you think of the relative ordering of "Agent hook locations" in the doc? It was higher but I pushed it to the end, maybe it should come before the discussion of the different hook categories and types though?

keithduncan:
> I went with category because we already use type to refer to the existing collection of hooks and I wanted to preserve that while introducing a new grouping. This way we have two categories, and in those two categories we have multiple types. What do you think?

Re-reviewing this we used to use "type" to refer to the location of the hook rather than the hook’s function 🙈 

I do still think this new naming names makes sense though, that there are two categories across three locations, in those locations we have different hook types.

pages/agent/v3/hooks.md.erb
plaindocs: Yeah, seems fine to me.

pages/agent/v3/hooks.md.erb
plaindocs: I actually wanted to redesign that table a few months ago, but haven't had any design time. Leave it as location order and we'll reexamine later?

plaindocs:
> I do still think this new naming names makes sense though, that there are two categories across three locations, in those locations we have different hook types.

The question I ask myself here is "even if the type / category thing is technically correct, does it matter to the read" and I think here it does make sense to make a distinction.

plaindocs:
> What do you think of the relative ordering of "Agent hook locations" in the doc? It was higher but I pushed it to the end, maybe it should come before the discussion of the different hook categories and types though?

As I see it there are now two minor awkward parts to the doc. The repetition of hook categories follow by the same info in the toc one para down, and the hook location, especially where we're mentioning them in the table. 

How about we just link all the hook locations in the table to the relevant section? 

And ignore the two categories coming before doc as not a real problem.

plaindocs:
@keithduncan where are we at with this? Still deciding on the exact naming?

keithduncan:
We’re ready to go with this naming @plaindocs!

plaindocs (approved):

PR #1123, opened 2021-08-16
Group step type page

karensawrey:
Logical grouping of jobs in the UI.


pages/pipelines/group_step.md.erb
```diff
@@ -0,0 +1,78 @@
+# Group Step
+ 
+A *group* step groups up parallel jobs into logical groups and allows displaying them on the build page in the Buildkite UI.  
+
+{:toc}
+ 
+You can scale your build pipelines in Buildkite to support an unlimited number of jobs. However, a large number of jobs on the build page can look messy. If pipelines contain 4-5 steps that can be logically grouped together, you can put them in a group using the `group` step type.  
+ 
+Job groups allow creating and then displaying logical groups in the Buildkite UI. They also help manage dependencies between a collection of steps in a dag (for example, step X depends on everything in a group Y).
```
plaindocs, line 9: The DAG stuff is also unrelated afaik, it's only mentioned because DAG + Groups isn't 100% tested yet. 

We should ask the team if there is an exhaustive list of what you can use in a group step? `notify`, `env` and `depends` at least?

pages/pipelines/group_step.md.erb
```diff
@@ -0,0 +1,78 @@
+# Group Step
+ 
+A *group* step groups up parallel jobs into logical groups and allows displaying them on the build page in the Buildkite UI.
```
plaindocs, line 3: So, I think we're slightly confusing two different things here. 

The *new* group step is unrelated to parallel steps. But it does *display* group steps in the same way as we already display group steps, in a compact way.

pages/pipelines/group_step.md.erb
```diff
@@ -0,0 +1,78 @@
...
+{:toc}
+ 
+You can scale your build pipelines in Buildkite to support an unlimited number of jobs. However, a large number of jobs on the build page can look messy. If pipelines contain 4-5 steps that can be logically grouped together, you can put them in a group using the `group` step type.  
+ 
+Job groups allow creating and then displaying logical groups in the Buildkite UI. They also help manage dependencies between a collection of steps in a dag (for example, step X depends on everything in a group Y).
+
+<%= image "job-groups-in-UI.png", width: 760, height: 328, alt: "Job groups displayed in the Buildkite UI" %>
+
+ 
+The `group` step type not only groups up parallel jobs and also groups them up by a `group_identifier` if it exists. For example:
```
plaindocs, line 14: Looks like the "group identifier" is "hello group" from line 19? But if so it isn't `group_identifier`. We need to be clear about this.

pages/pipelines/group_step.md.erb
```diff
@@ -0,0 +1,78 @@
...
+      - wait
+      - command: "echo hello"
+  - command: "echo goodbye"
+    parallelism: 2
+```
+ 
+The `group` step type:  
+ 
+* Internally switches the build to `dag` mode
+* Makes groups run in parallel next to each other
```
plaindocs, line 37: Is this true? I might have misunderstood.

plaindocs (changes requested):
Excellent start :+1: 

I think we need some more detail here. 

As we're adding this as a proper "step type" we'll want to mentions in a few places I think:

* https://buildkite.com/docs/pipelines/defining-steps
* image
* https://buildkite.com/docs/pipelines/controlling-concurrency maybe? 
* https://buildkite.com/docs/pipelines/example-pipelines maybe? 
* https://buildkite.com/docs/pipelines/dependencies

:-D 

we can pair on it next week if you like?

karensawrey:
> Excellent start 👍
> 
> I think we need some more detail here.
> 
> As we're adding this as a proper "step type" we'll want to mentions in a few places I think:
> 
> * https://buildkite.com/docs/pipelines/defining-steps
> * image
> * https://buildkite.com/docs/pipelines/controlling-concurrency maybe?
> * https://buildkite.com/docs/pipelines/example-pipelines maybe?
> * https://buildkite.com/docs/pipelines/dependencies
> 
> :-D
> 
> we can pair on it next week if you like?

Yes, we definitely need to discuss it further. I've spotted and fixed some other issues, addressing your comments now.

pages/pipelines/group_step.md.erb
karensawrey: On my TODO list for tonight.

pages/pipelines/group_step.md.erb
karensawrey: Which item on the list? :-D Let's discuss this.

PR #1124, opened 2021-08-17
Add input step example

StarfallProjects:


StarfallProjects:
@plaindocs I've tested both scripts (yay pi), so https://github.com/buildkite/input-step-example should be ok - but a second test never hurt. It's using the Bash example by default.

pages/pipelines/block_step.md.erb
```diff
@@ -18,17 +18,33 @@ steps:
...
-
-The data you collect from these fields is then available to subsequent steps in the pipeline in the [build meta-data](/docs/pipelines/build-meta-data). Block steps with input fields can only be defined using a `pipeline.yml`.
+You can add form fields to block steps with the `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items. Block steps with input fields can only be defined using a `pipeline.yml`.
 
 <%= image "block-step-with-fields.png", width: 944/2, height: 1178/2, alt: "Screenshot of a block step with input fields" %>
 
+The data you collect from these fields is then available to subsequent steps in the pipeline in the [build meta-data](/docs/pipelines/build-meta-data). 
+
+In this example, the `pipeline.yml` defines an input step with the key `release-name`. The Bash script then accesses the value of the step by getting that key from the meta-data.
```
plaindocs, line 27:
```suggestion
In this example, the `pipeline.yml` defines an input step with the key `release-name`. The Bash script then accesses the value of the step using the [meta-data](/docs/agent/v3/cli-meta-data) command.
```

pages/pipelines/block_step.md.erb
```diff
@@ -18,17 +18,33 @@ steps:
...
+
+```yml
+ - block: "Release"
+    prompt: "Fill out the details for release"
+    fields: 
+      - text: "Release Name"
+        key: "release-name"
+```
+{: codeblock-file="pipeline.yml"}
+
```
plaindocs, line 37: Do we need to mention that the bash script isn't run *in the example*? Kinda reinforcing the next para on line 43?

pages/pipelines/input_step.md.erb
```diff
@@ -21,9 +21,29 @@ steps:
 
 You can add form fields to input steps with the `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items.
 
-The data you collect from these fields is available to subsequent steps via the [build meta-data](/docs/pipelines/build-meta-data).
+The data you collect from these fields is available to subsequent steps through the [build meta-data](/docs/pipelines/build-meta-data). 
 
-## Input Step Attributes
+In this example, the `pipeline.yml` defines an input step with the key `name`. The Bash script then accesses the value of the step by getting that key from the meta-data.
```
plaindocs, line 26: And as above.

pages/pipelines/block_step.md.erb
```diff
@@ -18,17 +18,33 @@ steps:
 
 <%= image "block-step.png", width: 944/2, height: 364/2, alt: "Screenshot of a basic block step" %>
 
-You can add form fields to block steps with the `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items.
-
-The data you collect from these fields is then available to subsequent steps in the pipeline in the [build meta-data](/docs/pipelines/build-meta-data). Block steps with input fields can only be defined using a `pipeline.yml`.
+You can add form fields to block steps with the `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items. Block steps with input fields can only be defined using a `pipeline.yml`.
```
plaindocs, line 21: What do you think about? 

```suggestion
You can add form fields to block steps by adding a `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items. Block steps with input fields can only be defined using a `pipeline.yml`.
```

It might be worth adding a short section to our style guide on how to talk about YAML, so we're always using it in the same way.

plaindocs (commented):
Solid stuff! 

Just a couple of suggestions inline, lemme know what you think.

pages/pipelines/block_step.md.erb
StarfallProjects: Resolving this by adding the run script step to the yml.

plaindocs (approved):
This is great. 

Do you want to add the new example to the [manually curated list](https://buildkite.com/docs/pipelines/example-pipelines) as well?

PR #1127, opened 2021-08-24
Adding info about SEEK tool

karensawrey:


pages/agent/v3/securing.md.erb
```diff
@@ -45,6 +45,10 @@ Command line evaluation can be disabled by setting [`no-command-eval`](/docs/age
 
 As plugins execute in the same way as local hooks, they can pose a potential security risk. If you're using third party plugins, you'll be executing the third party's code on your agent.
 
+You can disable plugins with the command line flag: `--no-plugins`.
```
plaindocs, line 48: I think we  should keep both of the "disable plugins" options together. Let's link the "command  line flag one also"

image

Note: we might need to talk about some docs on how flags and settings can be used

pages/agent/v3/securing.md.erb
```diff
@@ -45,6 +45,10 @@ Command line evaluation can be disabled by setting [`no-command-eval`](/docs/age
 
 As plugins execute in the same way as local hooks, they can pose a potential security risk. If you're using third party plugins, you'll be executing the third party's code on your agent.
 
+You can disable plugins with the command line flag: `--no-plugins`.  
+
+If you still would like to use plugins, a third-party tool for signing pipelines will help you do it securely. See [Signing pipelines](https://buildkite.com/docs/agent/v3/securing#signing-pipelines) for more information.
```
plaindocs, line 50: More direct is better. :-)
And links without buildkite.com make it much easier to review PRs, because the links in previews work.

```suggestion
If you still want to use plugins, a third-party tool for signing pipelines will help you do it securely. See [Signing pipelines](/docs/agent/v3/securing#signing-pipelines) for more information.  
```

pages/agent/v3/securing.md.erb
```diff
@@ -102,6 +106,14 @@ do
 done
 ```
 
+## Signing pipelines  
+
+<section class="Docs__troubleshooting-note">
+  <p>Please remember that for maximum security it’s not recommended to use plugins.</p>
+</section>
```
plaindocs, line 113: I _think_ we can remove this? we're pretty clear about the limitations already? What do you think?

pages/agent/v3/securing.md.erb
```diff
@@ -102,6 +106,14 @@ do
...
+
+Suppose using plugins is crucial to your workflow, but you would still like to preserve strong security guarantees. In that case, you can use a third-party tool [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline). It adds extra security guarantees around Buildkite’s jobs by allowing the uploaded steps to be signed with a secret shared by all agents. As a result, plugins can run without concerns of tampering by third parties.
```
plaindocs, line 115:
```suggestion
If using plugins is crucial to your workflow, but you would still like to preserve strong security guarantees, you can use the [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline) plugin. It adds extra security guarantees around Buildkite’s jobs by allowing the uploaded steps to be signed with a secret shared by all agents. As a result, plugins can run without concerns of tampering by third parties. 
```

I'm not sure about "Buildkite’s jobs" here.

plaindocs: Do you have any suggestions for something more specific? What are we signing? The commands in the plugin, so you that you know that they've not been tampered with?

plaindocs (commented):
Nice and simple, I like it.

pages/agent/v3/securing.md.erb
karensawrey: Moved stuff around, but don't understand the part about linking. Let's discuss it today.

pages/agent/v3/securing.md.erb
karensawrey: Oops. Done.

pages/agent/v3/securing.md.erb
karensawrey: We're clear if we assume that the document is read top to bottom. If this piece is the starting point picked out from thee TOC... Better safe than sorry?

pages/agent/v3/securing.md.erb
karensawrey: As for "Buildkite's jobs", it's from the description of the tool, verbatim (https://github.com/seek-oss/buildkite-signed-pipeline).
I'll expand about the way the tool works - I originally wanted to keep it _super_ short here.

pages/agent/v3/securing.md.erb
plaindocs: Agreed that you can't assume that folks read top to bottom. 

But at the same time, saying "here are all the plugins you can use, but don't use them" is kinda...

plaindocs: Especially as we're saying this is a more secure way to use plugins. With that in mind I think we can remove the note. What do you think?

karensawrey: I gave this some thought, and yes, although we can't expect or assume that anyone is going to read top to bottom, it's better to remove the note.

PR #432, opened 2019-03-11
Add cloud identity SSO guide

harrietgrace:
A copy of the other guides, but for Cloud Identity. It's not a complex set up, so this draft has the majority of it in the steps. Just a few bits to clean up, and to decide if we want to add this as a guide 🤷🏻‍♀️

PR #1134, opened 2021-08-29
Update Google Workspace SSO docs

yob:
Today I wanted to setup SSO to a personal Buildkite org from Google Workspace (previously GSuite, previously Google Apps).

I'm familiar with doing so via OpenID, but I went the SAML option instead just to try something new. I got it working without much trouble, but our docs were a bit out of date so this is a first pass at updating them. I assume polish is needed, but I thought I'd open the discussion.

* Google Apps became GSuite, which has since become "Google Workspace"
* It wasn't clear to me what "Google Cloud Identity" was. It's not  mentioned in the Google Workspace admin area, so I changed the title to Google Workspace (SAML). It's clunky  though, and doesn't feel right. @harrietgrace do you have more historical context on what Google Cloud Identity is?
* As far as I can tell, customers using Google Workspace can choose the OpenID or SAML. I'm not sure why they should pick one over the other, but at the very least I think it's nice to communicate that these are alternatives for the same SSO provider.
* The Google Workspace admin panel has changed a bit, so the steps to create a custom SAML app have been updated to match

An open question is the URLs. This leaves them mentioning google cloud identity and gsuite, which doesn't match the text of the pages any more.

Sidenote: I checked all the Buildkite organizations with Google Workspace SSO setup and 93% (no exaggeration) are using OpenID.


harrietgrace:
Goodness Google changes their names around a lot! Having a squiz through the admin.google.com UI, that all looks super. 

@plaindocs I haven't looked at any of the wording or anything, just the technical content.

harrietgrace (commented):
Technical side of this looks 🙌🏻

plaindocs:
Thanks a ton folks. Will :eyes:

yob:
@plaindocs now that the AzureAD changes have shipped, I've given this another look over and rebased it onto master.

I still reckon this is worth considering. The current SSO docs for Google are pretty out of date in terms of the branding used by Google, **and** the UI we're describing.

Now that I've renamed the two pages to `Google Workspace` and `Google Workspace (SAML)` and it's clear these are two alternative ways to setup SSO with the same provider, I'm left with two questions:

1. Why should I customer pick one over the other. The OpenID path is **much** more popular with customers, but I'm nit sure why. One reason might be that the setup is easier.
2. If these are two alternative approaches for the same identity provider, would they make more sense as different sections on the same page?

I see the original google identity docs were added in #432, with comments by @keithpitt and @toolmantim - maybe they know something about the differences between OpenID and SAML with Google, and which one folks should pick?

Oh damn, my simple refresh of some outdated copy has gone down a rabbit hole :grimacing:

plaindocs:
Thanks for the refresh and update. :eyes: for real this time.

plaindocs:
On the **Google Workspace** and **Google Workspace (SAML)** question, moving them to the same page, makes it much easier to have a short paragraph with and overview of the differences and why you should pick one or the other. 

Alternatives include linking to the one page with that same overview which then links to both separate pages (cons: an extra click) or just duplicating the reason to pick one over the other on each page.

plaindocs (approved):
Regardless, best to merge this and iterate if the branding is off.

yob:
thanks :heart:

PR #1138, opened 2021-08-30
Improve link styles

lachlanhardy:
The links in the app were imported from Buildkite app in 2019 and
haven't been updated since.

We had an issue with `code` that are links or part of link text looking
odd or not consistently like a link. This commit re-styles links across
the whole site (!) but makes them consistent, including those with
`code`.

## `code` screenshots
Screen Shot 2021-08-30 at 17 10 05

Screen Shot 2021-08-30 at 17 10 33

## side-effects
The main side-effect of these changes is that all links now have underlines. Except the sidebar menu links. These now highlight with the brand green for consistency. If any of this is a step too far, please let me know!

image

### DIFF
--- a/app/assets/stylesheets/public/_variables.scss
+++ b/app/assets/stylesheets/public/_variables.scss
@@ -1,3 +1,4 @@
 $font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Helvetica, sans-serif;
 $text-on-dark-subdued-opacity: .8;
 $text-on-dark-very-subdued-opacity: .7;
+$color-brand-green: rgb(2, 129, 76);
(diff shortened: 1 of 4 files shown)


lachlanhardy:
I should have pointed out that the new link styles are based on the marketing site

plaindocs:
Just for clarity, this PR shows those as links right, so for markdown output there isn't actually any point in formatting text inside a link?

plaindocs (approved):
I'm happy to merge this as is, or to make an exception for in-page TOCs.
Those might well get a revamp in an upcoming PR also.

lachlanhardy:
@plaindocs In the link for '`X-Buildkite-Token` header', the `code` element still has the monospace font and sizing common to `code` elements elsewhere. Same with the `code` elements for `soft_fail` that are entirely links.

So, yes, marking it up as `code` in links still provides benefits.

lachlanhardy:
@plaindocs I'll hit merge and if you wanna revise the TOCs, we'll do it separately.

PR #1139, opened 2021-08-30
update submodule that points to bk/emojis

yob:
I'm not sure what docs uses this submodule for, however I've just renamed the emojis default branch from master to main so I thought I'd take the opportunity to refresh all the downstream users of it.

It's been over a year since docs refreshed the emojis submodule, so there's a few fun new emojis available to use now.


plaindocs (approved):
Oh, it is a manual refresh? TIL

PR #1140, opened 2021-08-30
configure dependabot to update the emojis submodule once per month

yob:
I did this manually in #1139, but we may as well make it automatic 🤷‍♂️

I'm .... 99% confident this will work. It matches the config we have in bk/bk.


plaindocs (approved):
:heart: ! Thank you

I'll keep an eye out for the first time it runs, and make sure all looks fine.

PR #1142, opened 2021-08-31
Update table styling for more flexibility

lachlanhardy:
## Tables without borders now work with the class `no-formatting`.
Screen Shot 2021-08-31 at 16 05 55

## Tables can now look like two columns with the class `two-columns`.
Screen Shot 2021-08-31 at 16 06 10


plaindocs (approved):
Excellent

PR #1144, opened 2021-09-01
Update list styling to make columns possible

lachlanhardy:
Sometimes we want to make lists more compact by spreading them out in
columns. On small screens (<500px) these will revert to a single column.

Technically this is just a class and so could be used on any element. It'll apply a basic columnar CSS Grid.

These examples are just for the screenshots, I didn't actually commit the addition of these classes to this list:
## List with the class `two-column-list`.
Screen Shot 2021-09-01 at 14 48 14

## List with the class `three-column-list`.
Screen Shot 2021-09-01 at 14 47 43

## Either list on a viewport of less than 500px
Screen Shot 2021-09-01 at 14 47 03


plaindocs (approved):
Perfect.

PR #1145, opened 2021-09-01
Fix typo in Permissions page

JuanitoFatas:
feel free to merge.


plaindocs (approved):
Nice one @JuanitoFatas 

When I've got a spare couple of minutes I'll implement a vale [block list](https://errata-ai.github.io/vale-server/docs/vocab) as well as our accept list, so we can prevent similar mistakes in the future.

PR #1150, opened 2021-09-03
added permissions section to github

StarfallProjects:


StarfallProjects:
@plaindocs Bogan has already taken a look at this and is happy with the info. We've discussed the fact that .buildkite.yml could be slightly misleading - turns out it's part of wider discussions about preferred practices for file-based config. So leaving as-is for now. If that doesn't make sense, let me know!

plaindocs (approved):
Looks good to me!

PR #1151, opened 2021-09-06
Small PagerDuty Change Events integration link improvements

toolmantim:
Now that PagerDuty have [nicer Change Events docs](https://support.pagerduty.com/docs/change-events) we should link to it, so this updates our integration guide to use that, as well as updating our support email address.


plaindocs (approved):
Nice one @toolmantim 

Should we switch out hello@ in other places also?

PR #1155, opened 2021-09-09
increase the CPU and memory we have available on ECS/Fargate

yob:
I've noticed the steady state for the docs Fargate tasks has memory utilization on ~95%. That seems a little close to the wire, so I've doubled the available memory.

CPU utilization is mostly OK,  but we do get occasional spikes up nearly 100%. Docs are important, I reckon it's reasonable to give them a full core to work with in each task.

Screenshot from 2021-09-09 21-41-05

Screenshot from 2021-09-09 21-41-54

### DIFF
--- a/deploy/task-definition-template.json
+++ b/deploy/task-definition-template.json
@@ -63,6 +63,6 @@
   "requiresCompatibilities": [
     "FARGATE"
   ],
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
:purple_heart:

PR #1157, opened 2021-09-10
Bump versions

plaindocs:
Bumped the rails stuff to the fixed railties version.
Confirmed everything works locally.


yob (approved):
nothing scary here :+1:

PR #1160, opened 2021-09-13
minor formatting update

danieluhl:
1. The `tests: ...` line is unnecessary and causes an error if pasted directly into the file. IMO better to omit as we'll be adding tests below anyway.

2. Missing space between words in line 220


plaindocs (approved):
Both nice fixes. Thanks!

PR #1162, opened 2021-09-17
Add words about enterprise-only to SCIM guides

harrietgrace:
Adding/updating the words about the Enterprise account gating for the SCIM providers - Okta and AzureAD.

I've also moved the note under the Okta Supported Features into its own note block.


yob (commented):
Looks good to me :+1:

plaindocs (approved):

PR #1163, opened 2021-09-18
AzureAD SCIM: simplify instructions

yob:
We've had feedback from a customer who has enabled SCIM with AzureAD that there's no need to customise the mapping for username. The default source property of userPrincipalName is an email address and may in fact be preferable to the mail property that we were recommending.

userPrincipalName is the username the user logs into AzureAD with, and apparently the email value can be different in some cases.

I've tested the new instructions with a test account and they work great 🎆

PR #1166, opened 2021-09-20
Additional tweaks to Azure AD docs

yob:
A follow up from #1163 - Remove the image file that was replaced by text

/cc @plaindocs @harrietgrace


plaindocs:
Sorry @yob , that'll teach me to YOLO commit to main. :sun_with_face:

plaindocs (approved):

yob:
all good!

I've rebased and force pushed, so now this is only  removing the unused image file

PR #1167, opened 2021-09-22
Add script and template to generate docs for CloudFormation template

### LINKED ISSUE #1093: Parameter documentation – extract to .md file
Given the complexity of the template it would be worthwhile extracting the parameter information from cloudformation to a markdown file for each release.

This would have the added benefit of making the process of determining changes in parameters by version far more easily discoverable.

keithduncan:
Add ERB template which generates a docs page from the inline CloudFormation template parameter docs.

Fixes #1093

### DIFF
--- a/pages/agent/v3/elastic_ci_aws/aws-stack.yml
+++ b/pages/agent/v3/elastic_ci_aws/aws-stack.yml
@@ -1,0 +1,1134 @@
+---
+AWSTemplateFormatVersion: "2010-09-09"
+Description: "Buildkite stack v5.6.1"
+
+Transform: AWS::Serverless-2016-10-31
+
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Buildkite Configuration
+        Parameters:
+        - BuildkiteAgentTokenParameterStorePath
+        - BuildkiteAgentTokenParameterStoreKMSKey
+        - BuildkiteAgentToken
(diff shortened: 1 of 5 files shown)


keithduncan:
It doesn’t look particularly pretty but the template works 😄 

https://docs-review-pr-1167.herokuapp.com/docs/agent/v3/elastic-ci-aws/parameters

Screenshot 2021-09-22 at 17 59 29

Can I hand this off to you @plaindocs to make it look good?

plaindocs:
This is great stuff @keithduncan .

Over in the ancient #1002 I started some nudges towards standardization on:

* YAML in `/data`
* having a script to generate the data, but leave the page generation to the rails app, do you reckon that would work here? similar to how agent updater works
* At some point we could look at automating the pull request when a new agent is released (or whatever would trigger an update here)

Any thoughts?

keithduncan:
> YAML in /data

:+1:

> having a script to generate the data, but leave the page generation to the rails app, do you reckon that would work here? similar to how agent updater works

Do you know whether these go through any precompiling by the assets pipeline? My aim here was to avoid having the docs frontends read, parse and generate that YAML file for each request since we know it isn’t going to change live.

> At some point we could look at automating the pull request when a new agent is released (or whatever would trigger an update here)

That’s the dream! 😄 depending on what you think of the doc content itself I did consider support for an 'overrides' file that could overwrite or append to the docs coming in from the template so we can add more detail here while leaving the template quite utilitarian.

PR #1168, opened 2021-09-27
BUILDKITE_ARTIFACT_UPLOAD_DESTINATION example path more structured.

pda:
We previously recommended (by way of example) `/$BUILDKITE_JOB_ID` as the prefix within the customer-hosted artifact bucket, which means a top-level prefix is created for each job, with the artifact files therein.

For Buildkite-hosted storage, we nest jobs inside builds inside pipelines inside organizations. This makes the bucket far easier to navigate, which becomes important when (a) there's millions of objects, and (b) everything is an opaque UUID.

This PR changes the documentation to recommend a similar hierarchy to the one we use on our end, but without the top-level Organization, which is omitted mostly because that environment variable doesn't exist in jobs, so can't be used by a hook:

```sh
/$BUILDKITE_PIPELINE_ID/$BUILDKITE_BUILD_ID/$BUILDKITE_JOB_ID
```

As a result, customers will be able to use Pipeline and Build as means to navigate / query / maintain their bucket, e.g. setting pipeline-specific lifecycle policies.

### DIFF
--- a/pages/agent/v2/cli_artifact.md.erb
+++ b/pages/agent/v2/cli_artifact.md.erb
@@ -251,7 +251,7 @@ Options:
 You can use the `buildkite-agent artifact` command to store artifacts in your own private Amazon S3 bucket. To do so, you’ll need to export the following environment variables using an [environment hook](/docs/agent/v2/hooks) (this can not be set via the Buildkite web interface, API, or during pipeline upload):
 
 ```bash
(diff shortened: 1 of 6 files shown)


plaindocs (approved):
Looks good to me.

Same comment in the other PR, that `_bootstrap.md` needs updating in the agent source code as well.

pda:
> Same comment in the other PR, that _bootstrap.md needs updating in the agent source code as well.

Ah yes :)

I've merged it for now anyway, and we'll also fix it in the agent codebase.

PR #1169, opened 2021-09-27
Replace i.e. with e.g. in most cases

pda:
Most of our usage of “i.e.” is to give a specific example, not a generally applicable clarification or restatement in different words, so I believe “e.g.” would be more suitable.

I make no claims of being a grammar/writing expert, so please correct me if I'm wrong.

### DIFF
--- a/pages/agent/v2/osx.md.erb
+++ b/pages/agent/v2/osx.md.erb
@@ -27,7 +27,7 @@ If you don't use Homebrew you should follow the [Linux](/docs/agent/v2/linux) in
 
 ## SSH Key Configuration
 
-SSH keys should be copied to (or generated into) the `.ssh` directory in the users’s home directory (i.e. `/Users/some-user/.ssh`). For example, to generate a new private key which you can add to your source code host:
+SSH keys should be copied to (or generated into) the `.ssh` directory in the users’s home directory (e.g. `/Users/alice/.ssh`). For example, to generate a new private key which you can add to your source code host:
 
 ```bash
 $ mkdir -p ~/.ssh && cd ~/.ssh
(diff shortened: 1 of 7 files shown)


pages/agent/v2/osx.md.erb
```diff
@@ -27,7 +27,7 @@ If you don't use Homebrew you should follow the [Linux](/docs/agent/v2/linux) in
 
 ## SSH Key Configuration
 
-SSH keys should be copied to (or generated into) the `.ssh` directory in the users’s home directory (i.e. `/Users/some-user/.ssh`). For example, to generate a new private key which you can add to your source code host:
+SSH keys should be copied to (or generated into) the `.ssh` directory in the users’s home directory (e.g. `/Users/alice/.ssh`). For example, to generate a new private key which you can add to your source code host:
```
pda, line 30: This one felt halfway between an _example_ and a _clarifying restatement_, so I shifted it more firmly into the _example_ camp.

pages/agent/v3/help/_bootstrap.md
```diff
@@ -41,7 +41,7 @@ See https://buildkite.com/docs/agent/v3/hooks for more details.
 * `--pipeline value` - The slug of the pipeline that the job is a part of [`$BUILDKITE_PIPELINE_SLUG`]
 * `--pipeline-provider value` - The id of the SCM provider that the repository is hosted on [`$BUILDKITE_PIPELINE_PROVIDER`]
 * `--artifact-upload-paths value` - Paths to files to automatically upload at the end of a job [`$BUILDKITE_ARTIFACT_PATHS`]
-* `--artifact-upload-destination value` - A custom location to upload artifact paths to (i.e. s3://my-custom-bucket) [`$BUILDKITE_ARTIFACT_UPLOAD_DESTINATION`]
+* `--artifact-upload-destination value` - A custom location to upload artifact paths to (e.g. s3://my-custom-bucket) [`$BUILDKITE_ARTIFACT_UPLOAD_DESTINATION`]
```
pda, line 44: This is going to conflict with #1168, which is where I originally spotted the i.e. vs e.g. thing.

plaindocs: These are all generated from the agent help as well, so need updating on that side too.

plaindocs:
@pda thank you for this wonderfully specific PR. 

Huge +1 on clarifying whether things are an example or a clarifying restatement, although I do think we should lean towards actually saying "for example" instead of "e.g.", just because it is more clear, especially to non native speakers. 

Do you fancy updating the PR to that effect, or wanna throw it over the wall to us?

pda:
> Do you fancy updating the PR to that effect, or wanna throw it over the wall to us?

I'm happy to come back to it One Day Soon, but if you do it first I'll be stoked :)

PR #1172, opened 2021-09-28
Add warning not to edit the files that are auto generated

keithduncan:
Remains to be seen how this affects the eventual markdown to html pipeline.


keithduncan:
This does end up being passed through to the resulting HTML but I don’t think it causes any problems:
Screenshot 2021-09-28 at 14 52 32

plaindocs (approved):
YOU STAR!

PR #1174, opened 2021-09-29
Add GA Universal tracking ID

toolmantim:
Adds the tracking code for our Google Universal Analytics property


toolmantim:
Given the relatively small scope of this, I'm just going to merge so we can get stuck into testing it!

plaindocs:
Sounds good!

PR #1179, opened 2021-10-03
Update elastic_ci_aws.md.erb

paulca:
I tried the example policy and IAM didn't like it.

I used the visual editor, and it added 3x `:::` semi-colons before the resource name.


plaindocs (approved):
Hi @paulca :wave: 

Yeah, that checks out, thanks for the fix!

PR #1189, opened 2021-10-11
Escape the emoji syntax in a YAML codeblock

keithduncan:


keithduncan:
Renders correctly now:

Screenshot 2021-10-11 at 11 09 24

PR #1190, opened 2021-10-11
Fix sidebar highlight for new AWS pages

keithduncan:
The docs I’ve added for AWS aren’t highlighting in the sidebar when you navigate to this, this fixes the link paths so that they show up highlighted:
- CloudFormation service role
- AWS Secrets Manager


keithduncan:
I’m going to merge this as a non-words change, but as always let me know if you’d rather I hadn’t 😄

PR #1222, opened 2021-11-02
signed pipelines updated

nikwek:
This is an attempt to give the tool a slightly more positive spin. The previous version makes it sound like it would be a terrible idea to use it. SEEK use it in production. 
Given that we will eventually take it over, we can afford to be more positive about it. 
Let me know what you think.

### DIFF
--- a/pages/agent/v3/securing.md.erb
+++ b/pages/agent/v3/securing.md.erb
@@ -107,7 +107,7 @@ done
 
 ## Signing pipelines  
 
-If using plugins is crucial to your workflow, but you would still like to preserve strong security guarantees, take a look at [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline) tool. It is an experimental reference implementation that adds extra security guarantees around Buildkite’s jobs by allowing to sign the uploaded pipelines with a secret shared by all agents. You can use it as an inspiration to build a similar tool that meets your needs, and it not recommended to use this tool 'as is' in a production setting.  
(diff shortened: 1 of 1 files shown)


pages/agent/v3/securing.md.erb
```diff
@@ -107,7 +107,7 @@ done
 
 ## Signing pipelines  
 
-If using plugins is crucial to your workflow, but you would still like to preserve strong security guarantees, take a look at [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline) tool. It is an experimental reference implementation that adds extra security guarantees around Buildkite’s jobs by allowing to sign the uploaded pipelines with a secret shared by all agents. You can use it as an inspiration to build a similar tool that meets your needs, and it not recommended to use this tool 'as is' in a production setting.  
+If using plugins is crucial to your workflow and you would still like to preserve strong security guarantees, take a look at the [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline) tool. This tool allows uploaded steps to be signed with a secret shared by all agents, so that plugins can run without any concerns of tampering by third-parties. Disclaimer: This functionality is not officially supported by Buildkite; use it at your own risk.
```
karensawrey, line 110:
```suggestion
If using plugins is crucial to your workflow and you would still like to preserve strong security guarantees, take a look at the [buildkite-signed-pipeline](https://github.com/seek-oss/buildkite-signed-pipeline) tool. This tool allows uploaded steps to be signed with a secret shared by all agents, so that plugins can run without concerns of tampering by third parties. Disclaimer: This functionality is still in experimental stage. 
```

plaindocs (approved):

plaindocs:
Thanks @nikwek

karensawrey (changes requested):
I suggest we tone down the mentions of risk in the PR. Otherwise - thanks for hugely improving the original description!!

karensawrey:
Oops, this one got merged a few seconds too early (literally). All the suggested changes were then merged in via https://github.com/buildkite/docs/pull/1223 so nothing needs to be done here anymore.
Thanks again, @nikwek!

PR #1224, opened 2021-11-02
Removing the disclaimer for the pipeline signing tool

karensawrey:


karensawrey:
The disclaimer was misleading and got removed.

PR #1225, opened 2021-11-03
WIP Analytics docs (DOC-153)

plaindocs:
DOC-153


pages/analytics/integrations.md.erb
```diff
@@ -0,0 +1,379 @@
...
+The following envionrment variables are generated by Test Analytics and can be used in your build.
+
+* `BUILDKITE_ANALYTICS_KEY` - a unique key for the build that initiated the Test Analytics run.
+* `BUILDKITE_ANALYTICS_URL` - a URL that links to the build
+* `BUILDKITE_ANALYTICS_BRANCH` - the branch or ref that this run is for
+* `BUILDKITE_ANALYTICS_SHA` - the commit sha for the head of the branch
+* `BUILDKITE_ANALYTICS_NUMBER` - the build number of the build
+* `BUILDKITE_ANALYTICS_JOB_ID` - the id of a job within the build
+* `BUILDKITE_ANALYTICS_MESSAGE` - the commit message for the head of the branch.
+
```
niceking, line 79: These environment variables aren't generated *by* Test Analytics (and there's also a typo in environment!) 

These environment variables need to be provided *to* Test Analytics. In the rspec-buildkite-analytics gem we use the following Buildkite env variables for these values:
https://github.com/buildkite/rspec-buildkite-analytics/blob/77a6f29ef8fe9232e2c13649980a841dd16247ae/lib/rspec/buildkite/analytics/ci.rb#L52-L62

Providing these env variables are important for the full TA experience (like the branch filtering on the pages) but none of them are strictly required by the JUnit or JSON API other than the `BUILDKITE_ANALYTICS_KEY`. I think maybe it needs to be communicated that sensible values need to be used for these if coming from another CI?

plaindocs: Cheers @niceking 

I get the suboptimal experience part, it's the same for the JSON ingestor, right? But if you're using RSpec it's sufficient to make sure the envionment variables exist, and they will be picked up, you don't need to pass them directionly, is what we're saying?

niceking: If you're using RSpec AND Buildkite, yeah they'll get picked up automatically. But if you're using RSpec and another CI tool, then you'll have to pass them/put them in an ENV file/however you deal with ENV variables

meiqimichelle:
Just noticed a thing: the url should probably be `/test-analytics` rather than `/analytics` to match the marketing we're doing, and also because there are likely to be other sorts of analytics in our future (such as build analytics and/or agent analytics).

meiqimichelle:
@plaindocs does this still need review, or can this be merged? We've shipped the JSON API 🎉

meiqimichelle (approved):
LGTM 🚀

PR #1228, opened 2021-11-04
Fixing broken emoji syntax

karensawrey:


karensawrey:
That was _not_ a typo.

PR #1230, opened 2021-11-08
Clarify AMIs don’t have to be built

keithduncan:
We build these AMI and mark them as public for customers to boot instances with.


karensawrey (approved):
Thanks for the improvement, @keithduncan!

PR #1241, opened 2021-11-09
Nested UL formatting

plaindocs:
Something like this should work?


plaindocs:
Looks ok for now, might need some finessing by @lachlanhardy another day, but good enough for today.
image

karensawrey (approved):
LGTM.

PR #1243, opened 2021-11-10
Build matrix docs

plaindocs:
This is a first pass

- [x]  `skip` in `adjustments` doesn't seem to work for me, have I misunderstood it?
- [x] I'm super unhappy with the attribute descriptions so far, will revisit
- [x] Matrixes or Matrices? :japanese_ogre: 
- [x] Truly [ugly code example](https://github.com/buildkite/docs/compare/matrix?expand=1#diff-602f51273a3d8b17a43680c2dc21450c63f841fa68e9afbc5d702cfc3e5d59feR138) but we've got work in progress to fix that
- [x] How does it interact with parallelism? You can't use matrix in parallel steps?


plaindocs:
Tracking this in DOC-132

ticky:
> You can't use matrix in parallel steps?

Correct, it's got to be one or the other.

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,96 @@
...
+For example, instead of writing three separate jobs for builds on **macOS**, **Linux** and **Windows**, like the following build configuration (which does not use a build matrix):
+
+```yaml
+steps:
+  - label: "macOS build"
+    command: "echo '.buildkite/steps/build-binary.sh macOS'"
+  - label: "Linux build"
+    command: "echo '.buildkite/steps/build-binary.sh Linux'"
+  - label: "Windows build"
+    command: "echo '.buildkite/steps/build-binary.sh Windows'"
```
ticky, line 16: I'd maybe swap this to a Windows-specific script type

```suggestion
    command: "echo '.buildkite/steps/build-binary.ps1 Windows'"
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,96 @@
...
+          test: "B"
+        soft_fail: true
+      - with:
+          os: "linux"
+          arch: "arm64"
+          test: "B"
+        skip: true
+```
+
+To add an extra combination that isn't present in the `matrix.setup`, use the `adjustments` key and make sure you define all of the elements in the matrix. For example, to add a build for [Plan 9](https://en.wikipedia.org/wiki/Plan_9_from_Bell_Labs) (on `arm64`, and test suite `B`) to the previous example:
```
ticky, line 82: 😄

ticky (commented):
It's probably worth noting that the matrix values can be interpolated into the command or in `env` variables.

Looks good!

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,96 @@
...
+steps:
+  - label: "{{matrix}} build"
+    command: "echo '.buildkite/steps/build-binary.sh {{matrix}}'"
+    matrix:
+      - "macOS"
+      - "Linux"
+      - "Windows"
+```
+
+All jobs created by a build matrix are marked with the Matrix symbol (👾) in the Buildkite interface.
```
eleanorakh, line 31: They're marked with a badge that says 'Matrix'!

plaindocs:
> > You can't use matrix in parallel steps?
> 
> Correct, it's got to be one or the other.

In the entire build? Or could you have "matrix section" and a "parallel" section in the same build?

plaindocs:
Thanks @ticky @eleanorakh

ticky:
> > > You can't use matrix in parallel steps?
> > 
> > 
> > Correct, it's got to be one or the other.
> 
> In the entire build? Or could you have "matrix section" and a "parallel" section in the same build?

They apply to a given step, so you can have parallelism on one step and matrix on another, within the same build. You can't, however, have parallelism within a matrix step.

plaindocs:
This is good to merge whenever the feature goes live.

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
+# Build Matrix
+
+Build Matrices lets you simplify complex build configurations by expanding a step template and array of matrix elements into multiple jobs.
```
karensawrey, line 3:
```suggestion
Build Matrix lets you simplify complex build configurations by expanding a step template and array of matrix elements into multiple jobs.
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+
+{:toc}
+
+For example, instead of writing three separate jobs for builds on **macOS**, **Linux** and **Windows**, like the following build configuration (which does not use a build matrix):
```
karensawrey, line 7:
```suggestion
For example, instead of writing three separate jobs for builds on **macOS**, **Linux**, and **Windows**, like in the following build configuration (which does not use a Build Matrix):
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+  - label: "macOS build"
+    command: "echo '.buildkite/steps/build-binary.sh macOS'"
+  - label: "Linux build"
+    command: "echo '.buildkite/steps/build-binary.sh Linux'"
+  - label: "Windows build"
+    command: "echo '.buildkite/steps/build-binary.ps1 Windows'"
+```
+{: codeblock-file="pipeline.yml"}
+
+You can use a build matrix, to expand a single step into three steps by replacing the `{{matrix}}` variable in the label and command step with each value provided in the `matrix` array. You can interpolate the matrix values into command steps (including step labels) and into environment variables.
```
karensawrey, line 20:
```suggestion
You can use a Build Matrix, to expand a single step into three steps by replacing the `{{matrix}}` variable in the label and [command step](/docs/pipelines/command-step) with each value provided in the `matrix` array. You can interpolate the matrix values into command steps (including step labels) and into [environment variables](/docs/pipelines/environment-variables).
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+    matrix:
+      - "macOS"
+      - "Linux"
+      - "Windows"
+```
+{: codeblock-file="pipeline.yml"}
+
+
+
+All jobs created by a build matrix are marked with the _Matrix_ badge in the Buildkite interface.
```
karensawrey, line 35:
```suggestion
All jobs created by a Build Matrix are marked with the _Matrix_ badge in the Buildkite interface.
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+
+<div class="Docs__note">
+<p class="Docs__note__heading">Matrix and Parallel steps</h3>
+<p>Matrix builds are not compatible with  <a href="/docs/tutorials/parallel-builds">parallel steps</a>. You can use both in the same build, as long as they are on separate steps.</p></div>
+
+For more complex builds, add multiple dimensions to `matrix.setup` instead of the `matrix` array. Each dimension you add multiplies by the other dimensions, so two architectures (`matrix.setup.arch`), two operating systems (`matrix.setup.os`), and two tests (`matrix.setup.test`), creates an 8 job build (2 * 2 * 2 = 8):
```
karensawrey, line 41:
```suggestion
For more complex builds, add multiple dimensions to `matrix.setup` instead of the `matrix` array. Each dimension you add is multiplied by the other dimensions, so two architectures (`matrix.setup.arch`), two operating systems (`matrix.setup.os`), and two tests (`matrix.setup.test`) create an 8 job build (2 * 2 * 2 = 8):
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+        - "windows"
+        - "linux"
+      test:
+        - "A"
+        - "B"
+```
+{: codeblock-file="pipeline.yml"}
+
+
+If you're using `matrix.setup` you can also use the `adjustments` key to change specific entries in the build matrix, or add new combinations. You can set the `skip` attribute to effectively remove them from the matrix, or `soft_fail` attributes to allow them to fail without breaking the build.
```
karensawrey, line 62:
```suggestion
If you're using `matrix.setup`, you can also use the `adjustments` key to change specific entries in the build matrix, or add new combinations. You can set the `skip` attribute to effectively remove them from the matrix, or `soft_fail` attributes to allow them to fail without breaking the build.
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+      - with:
+          os: "linux"
+          arch: "arm64"
+          test: "B"
+        skip: true
+```
+{: codeblock-file="pipeline.yml"}
+
+
+## Adding combinations to the build matrix
```
karensawrey, line 94:
```suggestion
## Adding combinations to the Build Matrix
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+
+To add an extra combination that isn't present in the `matrix.setup`, use the `adjustments` key and make sure you define all of the elements in the matrix. For example, to add a build for [Plan 9](https://en.wikipedia.org/wiki/Plan_9_from_Bell_Labs) (on `arm64`, and test suite `B`) to the previous example:
```
karensawrey, line 96:
```suggestion
To add an extra combination that isn't present in the `matrix.setup`, use the `adjustments` key and make sure to define all of the elements in the matrix. For example, to add a build for [Plan 9](https://en.wikipedia.org/wiki/Plan_9_from_Bell_Labs) (on `arm64`, and test suite `B`) to the previous example, use:
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+    adjustments:
+      - with:
+          os: "Plan 9"
+          arch: "arm64"
+          test: "B"
+```
+{: codeblock-file="pipeline.yml"}
+
+
+This results in 9 jobs, (2 * 2 * 2 + 1 = 9.).
```
karensawrey, line 108:
```suggestion
This results in 9 jobs, (2 * 2 * 2 + 1 = 9).
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+
+## Removing combinations from the build matrix
```
karensawrey, line 110:
```suggestion
## Removing combinations from the Build Matrix
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+          os: "linux"
+          arch: "arm64"
+          test: "B"
+        skip: true
+```
+{: codeblock-file="pipeline.yml"}
+
+## Matrix limits
+
+Each build matrix has a limit of 10 dimensions, 10 elements in each dimension and a total of 10 adjustments.
```
karensawrey, line 126:
```suggestion
Each Build Matrix has a limit of 10 dimensions, 10 elements in each dimension, and a total of 10 adjustments.
```

karensawrey (approved):

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+```
+{: codeblock-file="pipeline.yml"}
+
+
+
+All jobs created by a build matrix are marked with the _Matrix_ badge in the Buildkite interface.
+
+<div class="Docs__note">
+<p class="Docs__note__heading">Matrix and Parallel steps</h3>
+<p>Matrix builds are not compatible with  <a href="/docs/tutorials/parallel-builds">parallel steps</a>. You can use both in the same build, as long as they are on separate steps.</p></div>
```
toolmantim, line 39: @plaindocs This note might be a little confusing, because build steps run in parallel by default in Buildkite unless you add a `wait` step. After thinking "Oh no my example for the announcement blog won't work" I realised this note is specifically about lack of support for the `parallelism` property inside of `matrix` steps. Probably worth clarifying!

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+          test: "B"
+```
+{: codeblock-file="pipeline.yml"}
+
+
+This results in 9 jobs, (2 * 2 * 2 + 1 = 9).
+
+## Removing combinations from the build matrix
+
+It isn't currently possible to remove combinations using the `adjustments` key, but you can skip a combination by setting `skip: true`:
```
toolmantim, line 112: Is the word "currently" here because we're planning to add support for it? Perhaps best just to leave it off until we do?

```suggestion
It isn't possible to remove combinations using the `adjustments` key, but you can skip a combination by setting `skip: true`:
```

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+        - "arm64"
+      os:
+        - "windows"
+        - "linux"
+      test:
+        - "A"
+        - "B"
+```
+{: codeblock-file="pipeline.yml"}
+
```
toolmantim, line 60: I really wanted to see how this expanded out when run, similar to the first example, while authoring my own example of this in action. I had to try to run it in via the dashboard just to be sure, but perhaps we can just show how they're expanded/combined in the docs so they can probably give it a go first time in their pipeline.yaml file?

toolmantim:
Thanks for all the work on this doc, it's looking really great. I added two reviews (sorry, missed some on the first pass) with some things I spotted.

Also, now that https://buildkite.com/docs/pipelines/group-step is live, would it be common/sensible to wrap the command in a `group` step so it's nicely grouped in the UI. Seems like something people should do (unless they're already grouped?)

pages/pipelines/build_matrix.md.erb
```diff
@@ -0,0 +1,126 @@
...
+  - label: "macOS build"
+    command: "echo '.buildkite/steps/build-binary.sh macOS'"
+  - label: "Linux build"
+    command: "echo '.buildkite/steps/build-binary.sh Linux'"
+  - label: "Windows build"
+    command: "echo '.buildkite/steps/build-binary.ps1 Windows'"
+```
+{: codeblock-file="pipeline.yml"}
+
+You can use a Build Matrix, to expand a single step into three steps by replacing the `{{matrix}}` variable in the label and [command step](/docs/pipelines/command-step) with each value provided in the `matrix` array. You can interpolate the matrix values into command steps (including step labels) and into [environment variables](/docs/pipelines/environment-variables).
```
pda, line 20: > to expand a single step into three steps by replacing the `{{matrix}}` variable in the label and **command step**

I think it's in the label and command _attributes_ of a command step, rather than in the label and _command step_.

(Also not sure about the comma after “You can use a Build Matrix”)

pda:
> … would it be common/sensible to wrap the command in a group step so it's nicely grouped in the UI. Seems like something people should do (**unless they're already grouped?**)

@toolmantim FYI matrix jobs are not automatically grouped.

ticky:
I was under the impression the plan was to make them grouped once we got the MVP out and dogfooded?

ticky:
Sorry, misspoke earlier - I was under the impression a decision hadn't yet been made about their grouping behaviour/design, but perhaps I missed it.

pages/pipelines/build_matrix.md.erb
ticky: I think we should change this because `skip` is effectively that behaviour now; the jobs are not created

ticky:
A real-world example of the current state of the matrix gear is here: https://github.com/buildkite/agent/pull/1566, with a green build at [buildkite/agent build #4096](https://buildkite.com/buildkite/agent/builds/4096)

pages/pipelines/build_matrix.md.erb
plaindocs: Took me a moment to wrap my head around that. So good call on clarification.

pages/pipelines/build_matrix.md.erb
plaindocs: See my comment in Basecamp.

plaindocs:
Thanks for the feedback folks!

I'd rather not add Grouping to the example unless we really have to, it means we're talking about two concepts not just one. If anything we can easily add an end section about how to do that if it'll be a common thing.

scopegate:
@plaindocs just checking in again, ahead of launch later this week — are we good to merge these now?

plaindocs:
Good to go

PR #1258, opened 2021-11-22
Adding info about wait and block steps when they are placed together

nithyaasworld:
Added about wait and block steps when they are placed one after another.

Issue context: When a wait step is surrounded by a block step, the wait step is ignored. This has been previously reported by our customers a few times. So updating in our docs, so it is obvious that it's a limitation.

Here is the most recent observation from our customer: https://secure.helpscout.net/conversation/1696682218/27449/

**The following section is added in the PR:**
image

### DIFF
--- a/pages/pipelines/_block_wait.md.erb
+++ b/pages/pipelines/_block_wait.md.erb
@@ -1,0 +1,34 @@
+## Block steps interacting with wait steps
+
+If a block step follows or precedes a wait step in your build, the wait step will be ignored and only the block step will run, like this example:
+
+```yml
+steps:
+  - command: ".buildkite/steps/yarn"
+  - wait: ~
+  - block: "unblock me"
+```
+{: codeblock-file="pipeline.yml"}
+
+But let's consider a different example. Now the wait step will be ignored, but the block step will not be executed because `continue_on_failure` is set to `true`.
+
+
+```yml
+steps:
+  - command: "exit -1"
+  - wait: ~
+    continue_on_failure: true
+  - block: "unblock me"
+```
+{: codeblock-file="pipeline.yml"}
+
+If you need to run a block step after a failed step, set [soft_fail](/docs/pipelines/dependencies#allowed-failure-and-soft-fail) on the failing step:
+
+```yml
+steps:
+  - command: "exit -1"
+    soft_fail:
+      - exit_status: "*"
+  - block: "unblock me"
+```
+{: codeblock-file="pipeline.yml"}
--- a/pages/pipelines/_block_wait.md.erb
+++ b/pages/pipelines/_block_wait.md.erb
@@ -1,6 +1,6 @@
 ## Block steps interacting with wait steps
 
-If a block step follows or precedes a wait step in your build, the wait step will be ignored and only the block step will run, like this example:
+If a block step follows or precedes a wait step in your build, the wait step will be ignored and only the block step will run, like in this example:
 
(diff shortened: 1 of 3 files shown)


plaindocs (commented):
@nithyaasworld this is great work, love the detail and the examples. 

I think the second point isn't super clear though, probably we need to reference this somewhow:

> If there's a failure followed by a regular wait step, nothing after the wait step will run, including any subsequent wait steps with continue_on_failure: true.

Does that sound right?

nithyaasworld:
Thanks @plaindocs and @karensawrey for putting it in a more structured way 💜!

It's Awesome that it has been added to both wait and block step pages! 

>If there's a failure followed by a regular wait step, nothing after the wait step will run, including any subsequent wait steps with continue_on_failure: true.

Yeah, but this is regarding just the wait-step and we already have this in doc.. right?

For the second example marked in the screenshot below, the block step is not executed because the step above it (which is command step as wait-step was ignored) failed (with exit status "-1"). If a step is failed, block and any step following the block step will not be executed.  I especially added that example (that is an example with a command step that fails, followed by a wait step with continue_on_failure set to true, followed by a block step) because most customers try the exact same scenario and wonder why wait was not run.  Hope this provides more context! Please let me know otherwise. 

image

pages/pipelines/_block_wait.md.erb
```diff
@@ -0,0 +1,34 @@
...
+
+```yml
+steps:
+  - command: ".buildkite/steps/yarn"
+  - wait: ~
+  - block: "unblock me"
+```
+{: codeblock-file="pipeline.yml"}
+
+But let's consider a different example. Now the wait step will be ignored, but the block step will not be executed because `continue_on_failure` is set to `true`.
```
plaindocs, line 13:
```suggestion
But let's consider a different example. Now the wait step (with `continue_on_failure` is set to `true) will be ignored, but the block step will also **not run**, because the 'previous' command step failed.
```

plaindocs: @nithyaasworld thanks for the detailed explanation. It look me a surprisingly long time to understand what was going on.

I **love** that you've thought about documenting a common anti-pattern, and explaining it. Just double check that my rewording makes sense as you understand it, and we can hit the merge button.

nithyaasworld: Its looks good @plaindocs 🚀. Thanks for checking!

karensawrey:
I merged the suggestions and corrected a tiny omission in the source file, but the source file is rendered in a weird way - two link anchors (#) are displayed (and they both work). 

duplicate-link

plaindocs:
Haha, that looks like an artifact of including an H3, that is tedious.
@karensawrey do you want to move the heading out of the include file and into the wait and block files, and see if that fixes it?

karensawrey:
Lol, that was the first solution that I thought about, @plaindocs, but I dismissed it as a dirty hack because I also suspected that there'd be some sophisticated hi-tech workaround :) I'll move the H3s out into the pages & if that works, I'm merging.

karensawrey:
Yes, this eliminated the double anchors. 
works-fine

karensawrey:
Yes, we cracked this one! 
Awesome docs PR, @nithyaasworld!
Thanks a million for helping to untangle and troubleshoot whatever needed untangling and troubleshooting, @plaindocs!
Merging.

PR #1260, opened 2021-11-22
Update group step documentation with `commands` use examples

karensawrey:
Following https://forum.buildkite.community/t/group-steps-missing-commands/1790, updating the example to reflect the fact that `commands` does work in group step.

### DIFF
--- a/pages/pipelines/group_step.md.erb
+++ b/pages/pipelines/group_step.md.erb
@@ -185,7 +185,6 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
-Remember that a wait step placed inside a set of `commands` will be ignored because <code>wait</code> is not a command.
 
 ## Group merging
 
--- a/pages/pipelines/group_step.md.erb
+++ b/pages/pipelines/group_step.md.erb
@@ -48,7 +48,7 @@ _Required attributes:_
   <tr>
     <td><code>steps</code></td>
     <td>
-    A list of steps in the group; at least 1 step is required. Allowed step types: `wait`, `trigger`, `command`.<br>
+    A list of steps in the group; at least 1 step is required. Allowed step types: `wait`, `trigger`, `command`, `commands`.<br>
       <em>Type:</em> <code>array</code>
     </td>
   </tr>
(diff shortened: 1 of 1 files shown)


plaindocs (approved):

plaindocs:
This closes DOC-181

pages/pipelines/group_step.md.erb
```diff
@@ -185,6 +185,8 @@ steps:
 ```
 
+Remember that a wait step placed inside a set of `commands` will be ignored because <code>wait</code> is not a command.
```
plaindocs, line 188:
```suggestion
```

Yeah this is the thing that that I don't want to call out *here*. If we're gonna call it out it out it should be here 

https://buildkite.com/docs/pipelines/command-step#command-step-attributes

> The shell command/s to run during this step. This can be a single line of commands, or a list of commands that must all pass. Also available as the alias commands, in which case you can use multiple commands but not other step types. 

kinda thing

pages/pipelines/group_step.md.erb
```diff
@@ -48,7 +48,7 @@ _Required attributes:_
   <tr>
     <td><code>steps</code></td>
     <td>
-    A list of steps in the group; at least 1 step is required. Allowed step types: `wait`, `trigger`, `command`.<br>
+    A list of steps in the group; at least 1 step is required. Allowed step types: `wait`, `trigger`, `command`, `commands`.<br>
```
plaindocs, line 51:
```suggestion
    A list of steps in the group; at least 1 step is required. Allowed step types: [`wait`](/docs/pipelines/wait-step), [`trigger`](/docs/pipelines/trigger-step), [`command`/`commands`](/docs/pipelines/command-step)`.<br>
```

plaindocs (approved):
This is great. 

If we link the step types, it makes it easier for folks to look up how they work.
And I'm gonna go ahead and delete the wait thing, we can revisit updating the `command` docs after.

:tada:

pages/pipelines/group_step.md.erb
```diff
@@ -231,14 +236,14 @@ steps:
 - group: ~
   label: "linting"
   steps:
-    - name: ":rubocop: RuboCop"
+    - name: "\:rubocop\: RuboCop"
```
plaindocs, line 239: THANKS! :-)

PR #1263, opened 2021-11-23
Clarify top level env applies to further pipeline uploads

jradtilbrook:
This clarifies that top level YAML env vars apply to the entire pipeline, even further pipeline uploads not part of the original upload where the env is defined.

Before:

Screen Shot 2021-11-23 at 14 36 56
After:
Screen Shot 2021-11-23 at 14 35 18


karensawrey:
Awesome! Thanks for the contribution, @jradtilbrook!

plaindocs:
:tada:

PR #1267, opened 2021-11-24
Clarification on job priority

pzeballos:
- Bit explanation of what `dispatch` is
- Add a note on job priority

### DIFF
--- a/pages/pipelines/managing_priorities.md.erb
+++ b/pages/pipelines/managing_priorities.md.erb
@@ -1,5 +1,7 @@
 # Job Prioritization
 
+Jobs dispatch is when the job is taken from the queue and assigned to an agent.
+
 Jobs are dispatched to agents in the following order
 
 1. Job priority in descending order, highest number to lowest (`priority`)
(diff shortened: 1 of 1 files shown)


pages/pipelines/managing_priorities.md.erb
```diff
@@ -1,5 +1,7 @@
 # Job Prioritization
 
+Jobs dispatch is when the job is taken from the queue and assigned to an agent.
+
 Jobs are dispatched to agents in the following order
```
plaindocs, line 5: What do you think about using this instead:

> Jobs are dispatched (taken from the queue and assigned to agents) in the following order

This way we're not leading a section on Prioritization with a note on dispatch

pages/pipelines/managing_priorities.md.erb
```diff
@@ -18,7 +20,7 @@ steps:
 ```
 
-Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority will be assigned before jobs with lower priority.
+Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority will be assigned before jobs with lower priority, even if the last one was waiting longer.
```
plaindocs, line 23:
```suggestion
Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority will be assigned before jobs with lower priority, regardless of which has been longest in the queue.
```

plaindocs (changes requested):
Nice clarification. @pzeballos if you agree with the first change do you mind making that?

PR #1270, opened 2021-11-24
Warn about headings in snippets

plaindocs:
Let's see


plaindocs:
Good stuff! It's a little sad, but it works for now.

PR #1274, opened 2021-11-25
Clarify agent events

jradtilbrook:
Adds more details about what each event means and how it occurs.

Before:
Screen Shot 2021-11-25 at 10 19 48

After:
Screen Shot 2021-11-25 at 10 19 24

### DIFF
--- a/pages/apis/webhooks/agent_events.md.erb
+++ b/pages/apis/webhooks/agent_events.md.erb
@@ -8,23 +8,23 @@
 <tbody>
   <tr>
     <th><code>agent.connected</code></th>
-    <td>An agent has connected</td>
+    <td>An agent has connected to the API</td>
   </tr>
   <tr>
     <th><code>agent.lost</code></th>
-    <td>An agent has been marked as lost  </td>
+    <td>An agent has been marked as lost. This happens when Buildkite stops receiving heartbeats from the agent.</td>
   </tr>
   <tr>
     <th><code>agent.disconnected</code></th>
-    <td>An agent has disconnected</td>
+    <td>An agent has disconnected. This happens when the agent disconnects from the API.</td>
   </tr>
   <tr>
     <th><code>agent.stopping</code></th>
-    <td>An agent is stopping</td>
+    <td>An agent is stopping. This happens when an agent is instructed to stop from the API. It first transitions to stopping and finishes any current jobs.</td>
   </tr>
   <tr>
     <th><code>agent.stopped</code></th>
-    <td>An agent has stopped</td>
+    <td>An agent has stopped. This happens when an agent is instructed to stop from the API. It can be grafeul or forceful.</td>
   </tr>
 </tbody>
 </table>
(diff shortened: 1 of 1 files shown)


karensawrey:
The original info repeats in:
https://buildkite.com/docs/apis/webhooks/agent-events#events and
https://buildkite.com/docs/apis/webhooks#events

Should we turn the updated info into reusable snippets, @plaindocs?

pages/apis/webhooks/agent_events.md.erb
```diff
@@ -8,23 +8,23 @@
 <tbody>
   <tr>
     <th><code>agent.connected</code></th>
-    <td>An agent has connected</td>
+    <td>An agent has connected to the API</td>
   </tr>
   <tr>
     <th><code>agent.lost</code></th>
-    <td>An agent has been marked as lost  </td>
+    <td>An agent has been marked as lost. This happens when Buildkite stops receiving heartbeats from the agent.</td>
```
plaindocs, line 15: Worth checking if we talk about heartbeats elsewhere. In this case we don't use the term else where. Is this the term we should be using. Would "pings" work?

pages/apis/webhooks/agent_events.md.erb
```diff
@@ -8,23 +8,23 @@
...
   </tr>
   <tr>
     <th><code>agent.disconnected</code></th>
-    <td>An agent has disconnected</td>
+    <td>An agent has disconnected. This happens when the agent disconnects from the API.</td>
```
plaindocs, line 19: and is different from having no internet connection, say?

plaindocs (commented):
These are EXCELLENT additions!

@karensawrey yeah, definitely. Happy to pick that up?

karensawrey:
I'll be happy to turn these repeated bits into reusable snippets, @plaindocs!
@jradtilbrook let's clarify the 2 remaining unclear parts & I'll make sure it's broadcasted to all the places it's needed, from a single source.

pages/apis/webhooks/agent_events.md.erb
jradtilbrook: Oh yeah "pings" is probably better
```suggestion
    <td>An agent has been marked as lost. This happens when Buildkite stops receiving pings from the agent.</td>
```

pages/apis/webhooks/agent_events.md.erb
jradtilbrook: Or perhaps below is clearer, to distinguish between stopping (API tells it what to do) and disconnecting (agent decides what to do)

```suggestion
    <td>An agent has disconnected. This happens when the agent shuts down and disconnects from the API.</td>
```

jradtilbrook:
@karensawrey @plaindocs sounds good. I've added some suggestions to the comments

karensawrey:
Great work, folks! I'll propagate the changes via snippets ASAP. Merging now.

PR #1278, opened 2021-11-25
no-plugin-validation default value

pzeballos:
The default value for the configuration `no-plugin-validation` is actually [true](https://github.com/buildkite/agent/blob/5335389cb4649b86b6478ac542a40480500c6cca/clicommand/agent_start.go#L389). Change the default behavior on the agent would be more problematic than it seems, so we are updating the docs according in the meantime.
So if you actually want to have plugin validation you should specify `--no-plugin-validation=false` (the double-negative is confusing)

plugin


plaindocs:
Nice work @pzeballos I was going to make a comment on the double negative, but you beat me to it! :sun_with_face:

PR #1284, opened 2021-11-26
Remove deprecated bundler option

plaindocs:
I _think_ this does the right thing.

With the original code:

image

And with the PR:

image


plaindocs:
DOC-161

yob:
I built the `Dockerfile` locally and this works as advertised :tada: 

I was a bit surprised at the hoops we jump through when calling `bundle install` here - I'm guessing it's an attempt to make the `Dockerfile` suitable for production *and* dev/test workflows. All good, I'm confident it'll work fine in prod.

I might rebase and force push to confirm everything still works with current main.

PR #1286, opened 2021-11-29
About executing step conditionally based on the result of another step

nithyaasworld:
We always get questions around executing a step conditionally based on the result of another step. This is not directly possible using the if-conditionals. But we do achieve this by checking the result of the previous step in an ‘if’ condition **within a ‘command’ step** and then adding the other steps into the pipeline.

Please change the words/phrases as needed :)

Added the following in the PR:

image

### DIFF
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -59,9 +59,9 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
-As `'if'` conditions are evaluated at the time of the pipeline upload, its not possible to use a `if` attribute to conditionally run a step based on the result of another step.
+Since `'if'` conditions are evaluated at the time of the pipeline upload, it's not possible to use the `if` attribute to conditionally run a step based on the result of another step.
 
-So in order to run a step based on the result of another step, we upload a new pipeline based on a `if` condition set up in command step like below:
+In order to run a step based on the result of another step, upload a new pipeline based on the `if` condition set up in the [command step](/docs/pipelines/command-step) like in the example below:
 
 ```yml
 steps:
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -59,26 +59,25 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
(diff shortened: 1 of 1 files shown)


nithyaasworld:
I did a minor correction @plaindocs. I changed the phrase 'if attribute of a command step' to 'if attribute' as the if-attribute belongs to other steps too.

karensawrey:
That's an excellent addition, @nithyaasworld! Many thanks. I made minor changes and added a (hopefully :D) useful link.

pages/pipelines/conditionals.md.erb
```diff
@@ -59,6 +59,28 @@ steps:
...
+    depends_on: "sanity-check"
+    command: |
+      if [ $$(buildkite-agent step get "outcome" --step "sanity-check") == "passed" ]; then
+         cat <<- YAML | buildkite-agent pipeline upload
+         steps:
+           - label: "Run Regression"
+             command: ./scripts/regression_tests.sh
+      YAML
+      fi
+```
```
plaindocs, line 81:
```suggestion
  - label: "Validation check"
    command: ./scripts/validation_tests.sh
    key: "validation-check"
  - label: "Run regression only if validation check is passed"
    depends_on: "validation-check"
    command: |
      if [ $$(buildkite-agent step get "outcome" --step "validation-check") == "passed" ]; then
         cat <<- YAML | buildkite-agent pipeline upload
         steps:
           - label: "Run Regression"
             command: ./scripts/regression_tests.sh
      YAML
      fi
```
```

**Review by plaindocs — COMMENTED** [review]
Yep, nice addition, I do love pre-empting support queries with good docs

File: pages/pipelines/conditionals.md.erb [hunk 2]
```diff
@@ -59,6 +59,27 @@ steps:
...
+  - label: "Run regression only if validation check is passed"
+    depends_on: "validation-check"
+    command: |
+      if [ $$(buildkite-agent step get "outcome" --step "validation-check") == "passed" ]; then
+         cat <<- YAML | buildkite-agent pipeline upload
+         steps:
+           - label: "Run Regression"
+             command: ./scripts/regression_tests.sh
+      YAML
+      fi
```
plaindocs, line 80:
```suggestion
      fi
```      
```

Comment (plaindocs) [inline pages/pipelines/conditionals.md.erb:80]:
```suggestion
      fi
\`\`\`      
```

PR #1289, opened 2021-11-29
Note about default expanded group

pzeballos:
I added a note that by default the last group it's going to be expanded (always)
I didn't mention a workaround for this, but if someone wants that **all** groups to be collapsed, an option is to add a new group that does nothing.

default

### DIFF
--- a/pages/pipelines/managing_log_output.md.erb
+++ b/pages/pipelines/managing_log_output.md.erb
@@ -24,7 +24,7 @@ If you want to have the group open by default, use `+++` instead:
 echo "+++ A section of the build"
 ```
 
-If no group is explicitly expanded (`+++`), then the last collapsed regular group (`---`) gets expanded instead. By default, the last group it's going to be expanded.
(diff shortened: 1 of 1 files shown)


pages/pipelines/managing_log_output.md.erb
```diff
@@ -24,6 +24,8 @@ If you want to have the group open by default, use `+++` instead:
 echo "+++ A section of the build"
 ```
 
+If no group is explicitly expanded (`+++`), then the last collapsed regular group (`---`) gets expanded instead. By default, the last group it's going to be expanded.
```
plaindocs, line 27:
```suggestion
If no group is explicitly expanded (`+++`), then the last collapsed regular group (`---`) gets expanded instead. If you really want all groups to be collapsed, add an empty expanded group (`+++`).
```

plaindocs: How about this?

pzeballos: Like it! it's clear

PR #1300, opened 2021-12-07
Step attribute descriptions YAML and snippet reuse

karensawrey:
WIP

Initial commit for turning step attribute descriptions to YAML format and reusing those that can be encountered in different steps and share a description.


karensawrey:
Commit #1: 
- default
- options
- value

plaindocs:
OK, I've done a few things here:

1. I moved the first two attributes in the file list into a single YAML file in the `data` directory
2. While I was there I made the following changes (mostly for compatibility with #1002 :
      * changed `attribute` to `name`
      * split out `default_value` and `example` out of the description text and into their own keys
      * changed HTML to markdown
3. I added a single file in the models directory to expose the contents of the YAML file as the PIPELINE_ATTRIBUTES variable throughout the rails app

You can test this by adding the following code to the Command Step page (for example) which just loops over all of the attributes in that variable:

```
## TEST YAML attributes

<table class="Docs__attribute__table">
  <tbody>
    <% PIPELINE_ATTRIBUTES['attributes'].each do |attr| %>
      <tr>
        <td><%= attr['name'] %></td>
        <td><%= attr['desc'] %></td>
      </tr>
    <% end %>
  </tbody>
</table>

```

As there are only two, and we're not currently converting the markdown to html, it looks like this:

image

Once #1002 is merged we'll be able convert those snippets containing markdown to html instead of just showing it.

plaindocs:
For this work properly across step types we need to do one of two things, either specify in each attribute which step types it applies to, or what is probably a better idea write all (or most) of the the step attributes to apply to any step, and maintain a list of which attributes each step has, without other data.

So something like

```yaml
command_step_attributes:
  - command
  - agents
  - allow_dependency_failure
```

although we'll need to check how much difference there is in terms of what is optional and what isn't across step types, for example.

dannymidnight:
Since this one is WIP and now over a year stale I'm going to close this PR. Feel free to open again @karensawrey if you think it's still relevant.

PR #1303, opened 2021-12-12
Upgrade ruby from 2.7.2 to 2.7.5

yob:
Ruby 2.7.{3,4,5} have some bug and security fixes:

* https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-7-3-released/
* https://www.ruby-lang.org/en/news/2021/07/07/ruby-2-7-4-released/
* https://www.ruby-lang.org/en/news/2021/11/24/ruby-2-7-5-released/

Looking at the changelogs, none of the fixes seem to be hugely important for us. Still, we should update to keep up with security updates just in case there's impacts we aren't across.

I've changed the Dockerfile FROM in two interesting ways:

* I've used the new AWS mirror of docker hub library images (inc. ruby)[1]. We build images a lot (particularly in CI) so it might be nice to avoid the small risk of hitting docker hub rate limits
* I've explictly requested a buster image. The 2.7.2 base image we were using was buster, but the 2.7.5 images is built on bullseye. Upgrading to bullseye is a thing we should do, but lets do it when we'rr ready rather than as part of this ruby bump

I looked up the full digest with dfresh[2]:

    $ dfresh resolve public.ecr.aws/docker/library/ruby:2.7.5-buster
    public.ecr.aws/docker/library/ruby:2.7.5-buster@sha256:d76f1c822df854ecc6a983c63be3bd2e5854a6b47a3cdd140664953877d91651

[1] https://aws.amazon.com/blogs/containers/docker-official-images-now-available-on-amazon-elastic-container-registry-public/
[2] https://github.com/realestate-com-au/dfresh


yob:
@plaindocs if the build is green this should be safe to ship.

I'll leave it to you to merge though - the `.ruby-version` change means anyone using a ruby version manager in dev (like .rbenv) will need to install a new version after merge.

JuanitoFatas (approved):
👍🏼

plaindocs:
Nice one @yob @JuanitoFatas ! I'll ping a few channels and then get this merged.

PR #1305, opened 2021-12-15
Move `build-path` agent config to required attributes

jradtilbrook:
The agent code actually lists the `build-path` configuration option as required but the documentation for it is listed under optional. So this moves it to the required section.
See https://github.com/buildkite/agent/blob/6d052d212ab633e37bdd023a8c17e6a7d955402f/clicommand/agent_start.go#L65

Also mention `BUILDKITE_AGENT_CONFIG` env var for specifying configuration file location.

### DIFF
--- a/pages/agent/v3/configuration.md.erb
+++ b/pages/agent/v3/configuration.md.erb
@@ -15,7 +15,7 @@ debug=true
 ```
 {: codeblock-file="/buildkite/buildkite-agent.cfg"}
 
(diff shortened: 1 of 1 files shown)


plaindocs (approved):

plaindocs:
We definitely need a better system for unifying config options, commandline options and environment variables. CC @karensawrey

karensawrey:
> We definitely need a better system for unifying config options, commandline options and environment variables.
Very much agree, @plaindocs. We'll get there.

PR #1308, opened 2021-12-16
Refresh priorities docs

plaindocs:
- I've tweaked the intro and structure
- Added a mention on the homepage
~- Added a line to the quick reference controller ( @ticky is this all that needs to happen on this end?)~
- Added the command step attribute

### DIFF
--- a/app/controllers/quick_reference_controller.rb
+++ b/app/controllers/quick_reference_controller.rb
@@ -8,7 +8,7 @@ class QuickReferenceController < ApplicationController
     'pipelines/wait_step',
     'pipelines/block_step',
     'pipelines/input_step',
-    'pipelines/trigger_step'
+    'pipelines/trigger_step',
     'pipelines/managing-priorities'
   ].freeze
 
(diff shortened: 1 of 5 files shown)


app/controllers/quick_reference_controller.rb
```diff
@@ -8,7 +8,8 @@ class QuickReferenceController < ApplicationController
     'pipelines/wait_step',
     'pipelines/block_step',
     'pipelines/input_step',
-    'pipelines/trigger_step'
+    'pipelines/trigger_step',
+    'pipelines/managing-priorities'
```
ticky, line 12: This isn't a step, so it probably doesn't belong here! Linking it from the command step is sufficient, and doesn't require an icon 👍🏼

ticky (changes requested):

app/controllers/quick_reference_controller.rb
plaindocs, line 12:
```suggestion
    'pipelines/trigger_step'
```

ticky (approved):

PR #1311, opened 2021-12-20
DO NOT MERGE - PR for investigation only

plaindocs:
I'm just gonna look at what is going on in this diff.


plaindocs:
No longer required.

PR #1314, opened 2021-12-27
Jobs that timeout with an exit status of 0

karensawrey:


pages/pipelines/environment_variables.md.erb
```diff
@@ -496,9 +496,9 @@ Example: `"v1.2.3"`
 
 <h3 class="h3-caps">BUILDKITE_TIMEOUT</h3>
 
-The number of minutes until Buildkite automatically cancels this job, if a timeout has been specified. The value cannot be modified.
+The number of minutes until Buildkite automatically cancels this job, if a timeout has been specified. Jobs that timeout with an exit status of 0 will be marked as "passed". The value cannot be modified.
```
plaindocs, line 499: Just checking there is no default here? 

```suggestion
The number of minutes until Buildkite automatically cancels this job, if a timeout has been specified. Jobs that time out with an exit status of 0 are marked as "passed". The value cannot be modified.
```

plaindocs (approved):

PR #1323, opened 2022-01-11
Remove invalid group step attributes

karensawrey:
These attributes currently do nothing when used with the group step.


karensawrey:
This PR removes `branches`, `env`, and `skip` attributes from the group step page.

PR #1330, opened 2022-01-18
Added an example to deliver slack notification when a step is soft failed

nithyaasworld:
image


plaindocs:
Nice one @nithyaasworld 

I don't think there is a way round marking up the text blob, because according to YAML it is just that.

PR #1331, opened 2022-01-18
Adding a note on a failed blocked state

karensawrey:
@plaindocs this is the right thing to add, but is this a correct place to be adding it?..

### DIFF
--- a/pages/pipelines/_build_states.md.erb
+++ b/pages/pipelines/_build_states.md.erb
@@ -1,2 +1,6 @@
 Build state can be one of `creating`, `scheduled`, `running`, `passed`, `failed`, `blocked`, `canceling`, `canceled`, `skipped`, `not_run`.
 You can query for `finished` builds to return builds in any of the following states: `passed`, `failed`, `blocked`, or `canceled`.
+
+<div class="Docs__note">
+<p>Currently, a triggered build in <code>blocked_state: failed</code> leaves the triggering step in the <code>running</code> state. This may change in the future.</p>
+</div>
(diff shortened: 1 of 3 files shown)


pages/pipelines/_build_states.md.erb
```diff
@@ -1,2 +1,6 @@
 Build state can be one of `creating`, `scheduled`, `running`, `passed`, `failed`, `blocked`, `canceling`, `canceled`, `skipped`, `not_run`.
 You can query for `finished` builds to return builds in any of the following states: `passed`, `failed`, `blocked`, or `canceled`.
+
+<div class="Docs__note">
```
plaindocs, line 4: What about saying something like

> When a [triggered build](LINK) fails, the step that triggered it will be stuck in the `running` state forever.

plaindocs (commented):
Yeah, I think this is the right place. Although it might be worth mentioning in the trigger steps page as well.

pages/pipelines/_build_states.md.erb
karensawrey: That's a pretty ominous way of saying that. I like it! :D

plaindocs: :metal:

karensawrey:
@plaindocs can I merge now?

plaindocs:
Do you think it is worth linking *from* the trigger page to here? Or not?

karensawrey:
> Do you think it is worth linking _from_ the trigger page to here? Or not?

I now think that it's worth replicating the note to the trigger page.

plaindocs:
Thanks for the nudge @pzeballos :+1: :+1:

karensawrey:
Yay, the troubleshooting note color coding now works as expected!

@plaindocs lets 
shipit

plaindocs (approved):
:ship: :it:

Issue #1329, opened 2022-01-17
Fixed state does not exists


Thiryn:
https://github.com/buildkite/docs/blob/22bacc56506e3827c4b92e1126e4b917c7bb4a5c/pages/pipelines/conditionals.md.erb#L269

Hi, I am following this doc and it seems like the `fixed` state does not exist. Looking through other parts of the doc, it seems like this state indeed does not exists.
Getting the following error when building

> invalid: `if` expression failed to parse: "fixed" is not a valid `build.state` — did you mean "failed"?

GMNGeoffrey:
Hello there. I am hitting this same error when trying to use this state in a notify step. 

The link for notification services in the docs is https://buildkite.com/docs/pipelines/notifications#conditional-notifications and I believe I am using the variable exactly as specified there, but am getting an error.


```yaml
notify:
  - email: "llvm-bazel-alerts@googlegroups.com"
    if: build.state == "failed" || build.state == "fixed"
```

![image](https://user-images.githubusercontent.com/5732088/159099468-22b5aefe-8076-43b1-9ebd-21d8f1920d1a.png)

Does this only work with external integrations for some reason?

plaindocs:
Hi @GMNGeoffrey we've made another update to clear this up. in #1422 sorry for the confusion.

PR #1332, opened 2022-01-18
Note on fixed state

karensawrey:
The "fixed state" is only supported for notification services  & means that a build passed, but the previous build on the same branch failed. Adding a note about it as a follow up to https://github.com/buildkite/docs/issues/1329


plaindocs:
No idea if a note inside a table works, FYI :-p

karensawrey:
> No idea if a note inside a table works, FYI :-p

Yes it does!! In https://buildkite.com/docs/agent/v3/configuration

karensawrey:
"Note in a table" magic didn't work as expected, but I'm pretty happy with the current iteration. Shall we merge, @plaindocs?

plaindocs (approved):
Looks good, although I had to fix the spaces inside the `code` though, those hurt my eyes.

PR #1340, opened 2022-01-21
add some extra explanation around concurrency groups

jeremybumsted:
This is just to start adding some extra clarity around concurrency groups

### DIFF
--- a/pages/pipelines/controlling_concurrency.md.erb
+++ b/pages/pipelines/controlling_concurrency.md.erb
@@ -23,7 +23,9 @@ You can add concurrency limits to steps either through Buildkite, or your `pipel
 
 Concurrency groups are labels that group together Buildkite jobs when applying concurrency limits. When you add a group label to a step the label becomes available to all Pipelines in that organization. These group labels are checked at job runtime to determine which jobs are allowed to run in parallel. Although concurrency groups are created on individual steps, they represent concurrent access to shared resources and can be used by other pipelines.
 
-A concurrency group is essentially just a queue; it returns jobs in the queued order (oldest to newest). At runtime there are a number of active `job states` that the concurrency group cares about (limiting/limited/scheduled/waiting/assigned/accepted/running/cancelling/timing out) and the group becomes "locked" when the concurrency limit for jobs in these states is hit. Once a job moves from an active state to a terminal state (finished or cancelled) the job is removed from the queue, opening up a spot for another job to enter.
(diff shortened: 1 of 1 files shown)


nikwek:
I always love adding clarity and concurrency groups are complicated. One minor nit:
image
Would it make sense to add some whitespace in that long list of job states so that the line wraps nicer? 
image
Or maybe the list of states isn't really required if you link to https://buildkite.com/docs/pipelines/defining-steps#job-states ?

pages/pipelines/controlling_concurrency.md.erb
```diff
@@ -23,6 +23,8 @@ You can add concurrency limits to steps either through Buildkite, or your `pipel
 
 Concurrency groups are labels that group together Buildkite jobs when applying concurrency limits. When you add a group label to a step the label becomes available to all Pipelines in that organization. These group labels are checked at job runtime to determine which jobs are allowed to run in parallel. Although concurrency groups are created on individual steps, they represent concurrent access to shared resources and can be used by other pipelines.
 
+A concurrency group is essentially just a queue; it returns jobs in the queued order (oldest to newest). At runtime there are a number of active `job states` that the concurrency group cares about (limiting/limited/scheduled/waiting/assigned/accepted/running/cancelling/timing out) and the group becomes "locked" when the concurrency limit for jobs in these states is hit. Once a job moves from an active state to a terminal state (finished or cancelled) the job is removed from the queue, opening up a spot for another job to enter.
```
plaindocs, line 26: This is a nice suggestion @jeremybumsted ! Always be clarifying :-D 

I've tweaked your wording about, removing "essentially" and "just" which often aren't very helpful (`just` triggers the linter which considers the build broken). I've then tried to simplify things while keeping your explanation. 

```suggestion
A concurrency group works like a queue; it returns jobs in the order they entered the queue (oldest to newest). The concurrency group only cares about jobs in "active" states. and the group becomes "locked" when the concurrency limit for jobs in these states is reached. Once a job moves from an active state to a terminal state (`finished` or `cancelled`) the job is removed from the queue, opening up a spot for another job to enter.

The full list of "active" [job states](/docs/pipelines/defining-steps#job-states) is `limiting`, `limited`, `scheduled`, `waiting`, `assigned`, `accepted`, `running`, `cancelling`, `timing out`.
```

plaindocs: We could also just link to job states like @nikwek suggests, but here it is probably worth saving people a click and giving them the info.

plaindocs (approved):

PR #1342, opened 2022-01-24
Added more info about the order of plugin when there are multiple plugins in the same step

nithyaasworld:
**Now:**
image

**In the PR:**
image

### DIFF
--- a/pages/plugins/using.md.erb
+++ b/pages/plugins/using.md.erb
@@ -6,7 +6,7 @@ Plugins can be used in pipeline [command steps](/docs/pipelines/command-step) to
 
 ## Adding a plugin to your pipeline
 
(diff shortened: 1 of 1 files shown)


nithyaasworld:
Some context to the issue:

In the example below:

```
steps:
  - command: echo 'plugin testing'
    plugins:
      - shellcheck#v1.1.2:
          files: output.sh
      - docker#v3.3.0:
          image: node
          workdir: /app
      - docker-compose#v3.9.0:
          run: app
```

as per our testing, it ran in the order of the hooks (and within each hook, the order it was originally added into the pipeline step was followed). So for the above example, the order was:

image

So updated the doc with more clarification. 
cc: @pzeballos

karensawrey (approved):

PR #1347, opened 2022-01-31
Clarify `concurrency` must be used with `concurrency_group`

jradtilbrook:
There has been some confusion about a default value for the `concurrency` attribute when trying to define a concurrency group. Both `concurrency` and `concurrency_group` must be specified for it to work, and there is no error if one is omitted. This attempts to clarify that.

Existing page: https://buildkite.com/docs/pipelines/command-step

Before:
Screen Shot 2022-01-31 at 09 26 14

After:
Screen Shot 2022-01-31 at 09 25 32

Perhaps the API should enforce this but for now, a docs update would be good. I'm not sure if this change makes enough sense or whether we should add more to the `concurrency` row on this page to say something like: 

| There is no default value. You must specify a limit for concurrency limits to take effect.

### DIFF
--- a/pages/pipelines/command_step.md.erb
+++ b/pages/pipelines/command_step.md.erb
@@ -87,7 +87,7 @@ _Optional attributes:_
   <tr>
     <td><code>concurrency_group</code></td>
     <td>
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Nice one!

PR #1348, opened 2022-01-31
Added real example path to powershell

nithyaasworld:
We had a ticket today asking for the format of the shell path. So updated the doc for future users.

Now:
image

In the PR: 
image


plaindocs (approved):
Much better! :+1:

PR #1350, opened 2022-02-01
Added details on how to access buildkite-build-number when using YAML editor

nithyaasworld:
BUILDKITE_BUILD_NUMBER can be accessed if interpolated at runtime when using YAML editor. Added this clarification to the doc and also updated the flow a little.


plaindocs (approved):
Reads great!

PR #1353, opened 2022-02-01
remove reference to "finished" as a valid build.state

jeremybumsted:
`finished` is an alias for `passed` or `failed` builds, but it seems like it can't be referenced if the build never started (or perhaps some other change has occurred - not sure)

For example, if I create a notification trigger in my pipeline with the `if: build.state == 'finished'` I receive the following error on my pipeline upload step:

```
422 The `webhook` notification is invalid: `if` expression failed to parse: "finished" is not a valid `build.state` (line 1, column 16-25)
--
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1
```

So it makes sense to remove `finished` as the state from the valid states that can be referenced.

### DIFF
--- a/pages/pipelines/_job_states.md.erb
+++ b/pages/pipelines/_job_states.md.erb
@@ -1,1 +1,1 @@
-Job state can be one of `pending`, `waiting`, `waiting_failed`, `blocked`, `blocked_failed`, `unblocked`, `unblocked_failed`, `limiting`, `limited`, `scheduled`, `assigned`, `accepted`, `running`, `canceling`, `canceled`, `timing_out`, `timed_out`, `skipped`, or  `broken`.
+Job state can be one of `pending`, `waiting`, `waiting_failed`, `blocked`, `blocked_failed`, `unblocked`, `unblocked_failed`, `limiting`, `limited`, `scheduled`, `assigned`, `accepted`, `running`, `finished`, `canceling`, `canceled`, `timing_out`, `timed_out`, `skipped`, or `broken`.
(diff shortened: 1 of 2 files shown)


plaindocs:
These 'fake' states have always been a little tricky.

We mention `finished` in these two files: 

```
pages/pipelines/_job_states.md.erb
1:Job state can be one of `pending`, `waiting`, `waiting_failed`, `blocked`, `blocked_failed`, `unblocked`, `unblocked_failed`, `limiting`, `limited`, `scheduled`, `assigned`, `accepted`, `running`, `finished`, `canceling`, `canceled`, `timing_out`, `timed_out`, `skipped`, or  `broken`.

pages/pipelines/_build_states.md.erb
2:You can query for `finished` builds to return builds in any of the following states: `passed`, `failed`, `blocked`, or `canceled`.
```

Which are included in these two files:

```
pages/pipelines/defining_steps.md.erb
117:<%= render_markdown partial: 'pipelines/build_states' %>

pages/pipelines/notifications.md.erb
342:<%= render_markdown partial: 'pipelines/build_states' %>
```

Do we know what the actual use for `finished` ish? Maybe that would make it easier to explain around?

jeremybumsted:
@plaindocs  I'll try and find some clarification on the `finished` state and how it's actually intended to be used

jeremybumsted:
@plaindocs sorry I'm getting back to this so late. `Finished` as a state is an alias used for a group of `build states`, but is not used in all the places, which is where I got pretty confused.

Confirmed that the list of available build states for conditionals is as such:

```Creating
Started
Running
Scheduled
Blocked
Passed
Failed
Canceling
Canceled
Skipped
Not_run```

So, while you can query `finished` as a build state when querying the Builds API for example, it's not a valid state for conditionals.

PR #1354, opened 2022-02-02
Move insertion order section out from environment section

jradtilbrook:
The insertion order section applies to the pipeline upload command and not to the environment subsection so I have moved it out and increased its header.

Docs page: https://buildkite.com/docs/agent/v3/cli-pipeline#environment-variable-substitution-insertion-order

Before:
Screen Shot 2022-02-02 at 10 08 14

After:
Screen Shot 2022-02-02 at 10 08 27


plaindocs (approved):
Simple change is the best change.

PR #1366, opened 2022-02-07
Apt sources require "dearmored" pubkeys

kleetus:
* please see: https://wiki.debian.org/DebianRepository/UseThirdParty for more details on this practice. 
* in short, "The reason why we avoid ASCII-armored files is that they can only be used by SecureApt in version 1.4 or later (which appeared in stretch)." 
* you will see this practice in most other 3rd party sources being added. Check out signal app, docker, etc.


plaindocs (approved):
Nice one, I missed that when updating.
Thanks for the fix :smiley_cat:

PR #1371, opened 2022-02-10
Changed `github_check` to `github_commit_status`... 

karensawrey:
for the `notify` attribute example in the group step docs.


toolmantim (approved):
Thank you!

PR #1372, opened 2022-02-11
Remove an extra character from attribute examples

karensawrey:
Removed the extra colon from the group and command steps.


plaindocs (approved):
:100:

PR #1375, opened 2022-02-13
Indicate algolia api key is intentionally public

sj26:
Be clear that the algolia api key should be public. Exposing this key is not a vulnerability. It is meant to be shared with clients in order to access the search index and write usage data.

The comment here is within the javascript and should be carried through to the browser, so folks browsing rendered pages will see it too.


tessereth (approved):
LGTM but you might want to check why the linter is still running after 23 hours.

sj26:
Hm, yeah, I have no idea what's going on. It seems to be getting stuck on installing "alex".

@plaindocs any ideas?

plaindocs (approved):
Thanks @sj26 !

PR #1376, opened 2022-02-14
Add a note about filesystem security on windows

sj26:
When installing the buildkite agent it's important to be aware that it runs arbitrary scripts from hooks and checkout directories, and so if those directories are writeable by other users (which they are by default on Windows) then that should be a consideration when setting up a secure buildkite agent environment.

I'm not sure this is an appropriate place. @plaindocs do you have a feel for where this sort of advice belongs?

We are also adjusting the "automatic installation" script to do better things by default:

https://github.com/buildkite/agent/pull/1562


plaindocs:
Can we be more specific, and suggest specific permissions on Windows? Or is that not a thing?

sj26:
Yeah I'm happy with whatever you like, but I don't have the expertise to offer more pointed advice. I'd say shipping this is an improvement, and we could perhaps add stronger recommendations as we gain expertise over time.

sj26:
(I'll leave it to you to ship, if and how you like!)

plaindocs:
Thanks!

PR #1380, opened 2022-02-15
Adding a note on emoji escaping

karensawrey:


plaindocs (approved):
Seems good. That last paragraph gets a little rambly, but I'm not sure it's worth tightening up.

karensawrey:
Thanks for the review, @plaindocs!

>That last paragraph gets a little rambly, but I'm not sure it's worth tightening up.
Yeah. If I'm visited by the muse of brevity and clarity while working on the PR on troubleshooting notes, I'll revisit this. For the time being, it serves its purpose.

PR #1381, opened 2022-02-16
Optimization of the Defining Steps docs page

karensawrey:


plaindocs (commented):
Did you move the images to the correct folder?

plaindocs (commented):
This is all excellent, but I'm going to hold off merging till we can integrate the Job and Build states a bit more into the rest of the docs. I'm not sure how yet, But as it is it just feel like it's dangling out there.

PR #1383, opened 2022-02-18
Rework of the notes/troubleshooting notes section in the style guide

karensawrey:


plaindocs:
We can't merge this till we've sorted out the markdown notes. They don't work yet.

styleguide/STYLE.md
```diff
@@ -279,25 +279,46 @@ To generate a table of contents from all your \##\-level headings, use `{:toc}`.
...
+```
+<div class="Docs__troubleshooting-note">
+  <h3>Fast transitions and webhooks</h3>
+    <p>Note that if a builds transitions between states very quickly, for example from blocked (<code>finished</code>) to unblocked (<code>running</code>), the webhook may be in a different state from the actual build. This is a known limitation of webhooks, in that they may represent a later version of the object than the one that triggered the event.</p>
+</div>
+```
+
+Note that these note and troubleshooting note blocks are written in HTML so markdown syntax will not work. Use HTML syntax for links and formatting within 'troubleshooting note' blocks.
+
+It is also not recommended to use the heading level (`<h3 class="Docs__note__heading">Line endings</h3>`) in the warnings as it can introduce errors.
```
plaindocs, line 321: I'm not quite sure what you mean by this? 

The tricky part is that if you're in an H2 you need to use an H3 in the warning, and if you're in an H3 you should probably use an H4. Which is super annoying, especially if you're reorganizing headings. Which is why a solution that is level independent in the markdown would be ideal.

plaindocs (approved):
One comment aside, this is better than not having it in.

PR #1131, opened 2021-08-26
add doc on build skipping and link from vc

StarfallProjects:

PR #1384, opened 2022-02-21
Be more clear about job failure

plaindocs:
Fix for https://github.com/buildkite/docs/pull/1131#discussion_r810798233

### DIFF
--- a/pages/pipelines/skipping.md.erb
+++ b/pages/pipelines/skipping.md.erb
@@ -34,7 +34,9 @@ You can also configure these options using the [REST API](/docs/apis/rest-api/pi
 
 If your pipeline has multiple command steps, you can manually cancel a step, which causes the build to fail.
 
-If you do *not* want the build to fail when you cancel a specific step, you can set [`soft_fail`](/docs/pipelines/command-step#soft-fail-attributes), or [`allow_failure`](/docs/pipelines/dependencies#allowing-dependency-failures) for steps with dependencies.
+If you do *not* want the build to fail when you cancel a specific step, you can set [`soft_fail`](/docs/pipelines/command-step#soft-fail-attributes).
+
+To manually cancel a job:
 
 1. From your Buildkite dashboard, select your pipeline.
 2. Select the running build.


pages/pipelines/skipping.md.erb
```diff
@@ -30,16 +30,17 @@ To cancel running builds on the same branch:
 
 You can also configure these options using the [REST API](/docs/apis/rest-api/pipelines#create-a-yaml-pipeline).
 
-## Manually cancel jobs and skip to next step
+## Manually cancel a job
 
-If your pipeline has multiple command steps, you can manually cancel a step, and Buildkite will proceed to the next step. This does not cause the build to fail.
+If your pipeline has multiple command steps, you can manually cancel a step, which causes the build to fail.
+
+If you do *not* want the build to fail when you cancel a specific step, you can set [`soft_fail`](/docs/pipelines/command-step#soft-fail-attributes), or [`allow_failure`](/docs/pipelines/dependencies#allowing-dependency-failures) for steps with dependencies.
```
pda, line 37: Apart from `soft_fail` on the cancelled step/job, and `allow_failure` on the dependency pointing at it, would either of these be relevant here?

- `allow_dependency_failure` ([docs](https://buildkite.com/docs/pipelines/command-step#command-step-attributes)) on the step depending on the one that was cancelled. I guess this is just a version of `allow_failure` which applies to any/all dependencies.
- `continue_on_failure` ([docs](https://buildkite.com/docs/pipelines/wait-step)) on a wait step immediately after the cancelled step/job?

I intend to answer my own question today by testing these…

pda: Based on some testing, `continue_on_failure: true` on a `wait` step currently _does not_ proceed for cancellation, only for failures. But I think it's in the realm of undefined behaviour, and might change as it becomes defined…

pda: I've since verified that this bit is _not_ true:

> or [`allow_failure`](/docs/pipelines/dependencies#allowing-dependency-failures) for steps with dependencies.

Currently `allow_failure` on a step dependency does not treat cancelled the same as failed. I'm pretty keen to change that undefined behaviour though. Maybe best to leave it undefined/undocumented for now?

plaindocs:
@pda thanks for the feedback, I've fixed and am merging.

PR #1387, opened 2022-02-23
Jobs REST API has retry_type attribute.

pda:
Jobs in the REST API (and also outbound webhooks) now have a new field:

```
"retry_type": "automatic" | "manual" | null
```

I don't think we have such granular field-level documentation that there's somewhere to add more details, but this PR at least adds it to the one example response that also has `"retried": "true"` etc.

The Job Webhook Events docs link back to this Jobs REST API doc for details, so nothing to update over there.


plaindocs (approved):
Thanks Paul.
I'd love to get some more automation in our API docs so don't need to find these by hand, but we're not there yet

PR #1390, opened 2022-02-23
Update YAML heredoc in `notifications.md.erb`

the-real-cphillips:
The misaligned `YAML` for the heredoc causes the example code to fail.

Fixing the alignment resolves the issue and the example build runs as expected.

### DIFF
--- a/pages/pipelines/notifications.md.erb
+++ b/pages/pipelines/notifications.md.erb
@@ -276,7 +276,7 @@ steps:
                    channels:
                      - "#general"
                    message: "Step1 has soft failed."
-        YAML
+      YAML
       fi
 ```
 {: codeblock-file="pipeline.yml"}


plaindocs:
Hey @the-real-cphillips thanks for the report and fix! 

I'm currently looking this trying to wrap my head around the fact that the YAML heredoc is actually a Bash heredoc in a YAML file right? And as a Bash heredoc it should be aligned with the command it is part of, which is the `cat`? 

I'll take a closer look before merging. :-)

plaindocs (approved):
Can confirm the example works though :man_shrugging:

the-real-cphillips:
My pleasure on submitting the change! Happy to chip in :) 

In regard to the heredoc, it's just appears to be a Bash heredoc, the delimiter is `YAML`, but realistically could be anything, like:

```bash
if [ $$(buildkite-agent step get "outcome" --step "step1") == "soft_failed" ]; then
   cat <<- ELEPHANTS | buildkite-agent pipeline upload
   steps:
     - label: "Notify slack about soft failed step"
       command: echo "Notifying slack about the soft_failed step"
       notify:
         - slack:
             channels:
               - "#do-test-channel"
             message: "Step1 has soft failed."
ELEPHANTS
fi
```

This would work all the same, according to the heredoc docs, apparently the closing delimiter has to start in the first character position with no leading whitespace, so while `cat` is what it's associated with, it seems that's not the need with the heredoc:

>The closing limit string, on the final line of a here document, must start in the first character position. There can be no leading whitespace. Trailing whitespace after the limit string likewise causes unexpected behavior. The whitespace prevents the limit string from being recognized. [More Info](https://tldp.org/LDP/abs/html/here-docs.html#FTN.AEN17822)

Thanks for the merge and the questions, made me dig into heredoc more than I'd expected too today lol. Have a good one!

plaindocs:
Super clear. Thanks!

PR #1391, opened 2022-02-26
Clarification for manually cancelling steps and allow_failure dependencies

jeremybumsted:
Currently, manually cancelling steps that are a part of `allow_failure` dependencies results in builds being marked as `Failed`, so this should add some clarification in the docs so future travellers can know and avoid some confusion.


plaindocs:
Cheers Jeremy. 

I think this is covered #1384, although I've not sure that PR has all the info yet.

PR #1393, opened 2022-02-28
Include updated checkboxes to launch stack

MelissaKaulfuss:
There is a new checkbox item in the Elastic Stack Cloudformation template prior to launching the stack.
I've added the additional item to the copy and updated the screenshot.


plaindocs (approved):
Cheers Mel!

PR #1398, opened 2022-03-01
Add Ruby to Gemfile

plaindocs:
Heroku build previews were breaking because it wasn't using `.ruby-version`. This does mean we have two places to update the Ruby version though. :/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# Fix Heroku?
+ruby "2.7.5"
+
 source "https://rubygems.org"
 
 # Choo choo 🚝 (only include the Rails gems we need)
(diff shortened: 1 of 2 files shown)


plaindocs:
Is there a better way to approach this?

yob:
It's unfortunate the this duplicates the info, but I'm a heroku n00b so I'm not sure if there's a better way. If it unblocks you, maybe it's worth holding your nose and doing.

lachlanhardy:
@plaindocs @yob There's a little trick folks add to their `Gemfile` to work around this: 

`ruby File.read(".ruby-version").strip`

Happy to add a fresh PR for it.

lachlanhardy:
Done: https://github.com/buildkite/docs/pull/1401

PR #1410, opened 2022-03-11
include `soft_failed` as an option in build.state

rjl79:
Update to include the `soft_failed` build state that is available to the `build.state` conditional.


plaindocs (commented):
Nice catch, thank you!

PR #1418, opened 2022-03-17
Add a note about agent logs for linux install

nsuma8989:
Currently in Linux installation steps documentation, in file locations we do not have any mention about agent logs so adding a note to explain how to access agent logs in such scenarios.


plaindocs:
Just confirming that it doesn't work like this example on Red Hat docs?
image

I've edited the PR so it's more similar to that, even if the functionality is different.

nsuma8989:
> 

No, if agent was started using `buildkite-agent start` instead being part of `systemd` then we found logs wont be part of /var/log/buildkite-agent.log

plaindocs (approved):
LGTM

PR #1420, opened 2022-03-18
Clarification around artifact download

pzeballos:
What it's stated on the docs it's not exactly correct. If you do multiples uploads of the same artifact, when we do an artifact download specifying a folder path (for example `buildkite-agent artifact download 'folder/*' .` ), it's going to download the most recent one (the complete folder). But if you want to download a specific file (like `buildkite-agent artifact download 'folder/filename1.jpg' .`) it will throw an error because there is an ambiguity.

In the case of duplicates, Buildkite will require disambiguation. If you are going to upload multiples files with the same path/filename, you have to incorporate $BUILDKITE_JOB_ID.

I'm leaving to you the correct wording for this 😊

artifact

### DIFF
--- a/pages/pipelines/artifacts.md.erb
+++ b/pages/pipelines/artifacts.md.erb
@@ -47,13 +47,19 @@ You can download artifacts created by a build job, from inside a running build, 
 buildkite-agent artifact download logs/* local-logs/
 ```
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks @pzeballos , I've made more changes than expected, can you take a look and verify?

plaindocs:
@pzeballos did you manage to take a look at this?

PR #1421, opened 2022-03-18
Note on `no-local-hooks`

pzeballos:
If you enable `no-local-hooks`, it will also affect plugins.

local-hook
config

### DIFF
--- a/data/agent_attributes.yaml
+++ b/data/agent_attributes.yaml
@@ -192,7 +192,7 @@ attributes:
   default_value: "false"
   required: false
   desc: | 
-      Do not execute any local hooks.
+      Do not execute any local hooks, including plugins (both local and hosted).
 - name: no-plugins
   env_var: BUILDKITE_NO_PLUGINS
   default_value: "false"
--- a/data/agent_attributes.yaml
+++ b/data/agent_attributes.yaml
@@ -192,7 +192,7 @@ attributes:
   default_value: "false"
   required: false
   desc: | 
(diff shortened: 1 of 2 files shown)


plaindocs (commented):
"Hosted" plugins feels confusing to me. 

> Non-Buildkite plugins hosted on GitHub require you to include the organization name as well as the plugin name, for example:

But they are still downloaded and run, they aren't "run in the cloud".  So maybe 

> Disabling local hooks also disables all plugins.

What do you think?

pzeballos:
For me is clear. A user read: "Disabling local hooks will also disable plugins" and asked "Only repository plugins or no plugins in general?". But maybe adding the `all` it will make it clearer

plaindocs:
Yeah, I guess my main objection is referring to them as "hosted plugins". If we want to call them out specifically we could do something like

> Disabling local hooks also disables plugins from all sources.

or 

> Disabling local hooks also disables plugins from all sources, including local plugins and plugins hosted on GitHub.

I'd tend towards the first option.

pzeballos:
I like the first option! :)

data/agent_attributes.yaml
```diff
@@ -192,7 +192,7 @@ attributes:
   default_value: "false"
   required: false
   desc: | 
-      Do not execute any local hooks.
+      Do not execute any local hooks, or plugins from any sources..
```
plaindocs, line 195:
```suggestion
      Do not execute any local hooks, or plugins from any source.
```

plaindocs (approved):
LGTM

PR #1422, opened 2022-03-19
fixed state not available in conditionals

pzeballos:
`fixed` state is not available on conditionals. We have the concept of "fixed" on Notifications (to indicate that a build on the same branch passed after failing), but only through this page:

notification-service

You can't use `build.state = 'fixed'`


plaindocs (approved):
These not quite real states seem to cause a lot of trouble.

GMNGeoffrey:
> These not quite real states seem to cause a lot of trouble.

It would be very nice to make it a real state that could be used in conditionals :-) I'd love to be able to send simple email notifications based on newly-failing or newly-fixed. I'm probably going to roll my own using the Buildkite API, but native support would be nice.

PR #1423, opened 2022-03-23
fix indentation issue around step-level notifications examples

pzeballos:
step-level notification examples had wrong indentation. The pipeline was not correct.


plaindocs (approved):
Thanks @pzeballos 

I'd love to be able to lint our YAML code samples.

PR #1425, opened 2022-03-25
Update docs for agent version 3.35.0

moskyb:


pda (approved):

pda:
https://docs-review-pr-1425.herokuapp.com/docs/agent/v3/cli-artifact#fetching-the-sha-of-an-artifact seems good

pda:
I'm _slightly_ concerned about public documenting the SHA-256 feature when it's not supported server-side yet.

> `--sha256`  Request SHA-256 instead of SHA-1, errors if SHA-256 not available

The “errors if SHA-256 not available” is technically true, and will happen all the time.

I think that's fine though — it's not like we're changelogging it here, and if somebody notices and asks/complains that will be super valuable product usage insight :)

🚀 IMO

Happy to defer to @plaindocs though.

plaindocs:
Hey folks, cheers for the PR!

@moskyb just checking you're generating this from the agent CLI not writing it manually? 

@pda I'm not quite sure what you mean by "not supported server side"? 

```
The --sha256 argument requests SHA-256 instead of SHA-1; this is only
available for artifacts uploaded since SHA-256 support was added to the
agent.
```

Are we saying the agent doesn't have this? Or are we saying it does but most artifacts don't? 

(side note, would it be worth adding a flag for "get SHA-256 if it's available, if not get SHA-1"?)

pda:
Currently, agent v3.35.0 `buildkite-agent artifact upload` will calculate and upload SHA-256, but the server dose not save it.
So then agent v3.35.0 `buildkite-agent artifact shasum --sha256` will never find it.

However, server-side support will begin shortly (next week?)… unless we decide against it, and then we'd remove it from the agent and docs instead :)

plaindocs (approved):
@pda Gotcha.

I have a broad preference for not telling folks about stuff that doesn't work, but in this case we have enough caveats for it not to be a big deal. I'll merge and if we get a bunch of questions we can reevaluate.

PR #1430, opened 2022-03-28
ssh key location clarification

nithyaasworld:
Doc says to generate the ssh key on the logged-in user's root folder instead of where the buildkite-agent is installed. So modified those examples to point to buildkite-agent installation location.

Below are the 2 places where I did the change. Others are just copied from here.

image

image

These are the places affected by the above change:
https://buildkite.com/docs/agent/v3/ssh-keys#finding-your-ssh-key-directory
https://buildkite.com/docs/agent/v3/ssh-keys#creating-a-single-ssh-key
https://buildkite.com/docs/agent/v3/ubuntu#ssh-key-configuration
https://buildkite.com/docs/agent/v3/debian#ssh-key-configuration
https://buildkite.com/docs/agent/v3/redhat#ssh-key-configuration
https://buildkite.com/docs/agent/v3/freebsd#ssh-key-configuration
https://buildkite.com/docs/agent/v3/linux#ssh-key-configuration

### DIFF
--- a/pages/agent/v3/_ssh_key_with_buildkite_agent_user.md.erb
+++ b/pages/agent/v3/_ssh_key_with_buildkite_agent_user.md.erb
@@ -2,7 +2,6 @@ SSH keys should be copied to (or generated into) `/var/lib/buildkite-agent/.ssh/
 
 ```bash
 $ sudo su buildkite-agent
-$ mkdir -p ~/.ssh && cd ~/.ssh
 $ mkdir -p (echo ~/.buildkite-agent)/.ssh #ssh-key should be present inside buildkite-agent installation location which is /var/lib/buildkite-agent/ in this case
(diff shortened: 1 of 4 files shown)


nithyaasworld:
@jradtilbrook I would like your review on this before passing this to @plaindocs. Thanks!

jradtilbrook (commented):
Good thinking trying to improve the docs here. But I'm thinking they are okay as they are. It all mentions the user the agent is running as, and then referring to home through `~` which is all pretty standard linux terminology.  
Actually, are you sure the changes are accurate? I don't think "buildkite-agent installation location" is the same as the home directory for the buildkite-agent 🤔 
Perhaps leave it up to @plaindocs?

nithyaasworld:
will check on this and re-open with clean commits :)

PR #1437, opened 2022-03-31
Fix Authorization header for Agent API

nsuma8989:
Agent API uses HTTP Authorization header with "Token" instead of "Bearer". This PR fixes this in the agent API docs


nsuma8989:
@plaindocs Here is the pull request to fix the authorization header for agent API as we discussed

plaindocs (approved):
Nice one @nsuma8989

PR #1439, opened 2022-04-01
Update using.md.erb

es-arif:
While developing a plugin, it is logical that if you can pin a commit or release you can also reference a branch, but it isn't _obvious_.

Trying to help others that may not notice!

### DIFF
--- a/pages/plugins/using.md.erb
+++ b/pages/plugins/using.md.erb
@@ -148,7 +148,7 @@ steps:
           run: app
 ```
 
-## Testing plugin versions
+## Referencing plugins from a specific branch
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Hey @es-arif thanks for the suggestion. 

My main comment is that the heading might be misleading, so how about saying something like

```
## Loading plugins from a branch
```

instead of 'Testing'?

es-arif:
I agree "Testing" isn't the best word.

How about "Sourcing" or "Referencing" instead?

plaindocs (approved):
Thanks for the suggestion.

PR #1442, opened 2022-04-02
Added retry attributes to the API response examples

nithyaasworld:
Retry attributes were added to the API response recently. Updated the examples to reflect the change.

### DIFF
--- a/pages/apis/rest_api/jobs.md.erb
+++ b/pages/apis/rest_api/jobs.md.erb
@@ -34,10 +34,10 @@ curl -X PUT "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pi
       "runnable_at": null,
       "started_at": null,
       "finished_at": null,
-      "retried": "true",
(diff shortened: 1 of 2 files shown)


pages/apis/rest_api/jobs.md.erb
```diff
@@ -34,10 +34,10 @@ curl -X PUT "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pi
...
       "started_at": null,
       "finished_at": null,
-      "retried": "true",
-      "retried_in_job_id": "b63254c0-3271-4a98-8270-7cfbd6c2f14e",
-      "retries_count": "1",
-      "retry_type": "automatic",
+      "retried": false,
+      "retried_in_job_id": null,
+      "retries_count": 1,
+      "retry_type": null,
```
pzeballos, line 40: I think it was good to leave the example with "automatic"  instead of "null" It was added in #1387

plaindocs: Oops. 

I'd really like to find the time to look into automating the API bodys.

PR #1452, opened 2022-04-11
Added details about BUILDKITE_* variables supported by build.env()

nithyaasworld:
Added details about BUILDKITE_* variables supported by build.env()

image


plaindocs:
Hi @nithyaasworld - thanks for this.

I'm folding the changes into my #1455 PR. My only question is does the function work **only** with the list of variables? Or only with those BUILDKITE_* variables but also other ones, for example defined by the user?

plaindocs:
Thanks.

PR #1455, opened 2022-04-11
Refactor conditionals info

plaindocs:
I think this is more clear.

### DIFF
--- a/pages/pipelines/conditionals.md.erb
+++ b/pages/pipelines/conditionals.md.erb
@@ -224,7 +224,30 @@ The following variables are supported by the `if` attribute. Note that you canno
 	<tr>
 		<td><code>build.env()</code></td>
 		<td><code>String</code>, <code>null</code></td>
-		<td>This function returns the value of the environment passed as the first argument, otherwise <code>null</code> if the environment variable has not been set</td>
+		<td>This function returns the value of the environment passed as the first argument if that variable is set, or <code>null</code> if the environment variable is not set.<br>
+		<code>build.env()</code> works with the following variables only:<br>
+		<code>BUILDKITE_BRANCH</code><br>
+		<code>BUILDKITE_TAG</code><br>
+		<code>BUILDKITE_MESSAGE</code><br>
+		<code>BUILDKITE_COMMIT</code><br>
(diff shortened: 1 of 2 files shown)


pages/pipelines/conditionals.md.erb
```diff
@@ -85,9 +85,30 @@ steps:
...
+```yaml
+notify:
+  - email: "dev@acmeinc.com"
+    if: build.state == "passed"
+```
+{: codeblock-file="pipeline.yml"}
+
+## Conditionals and the broken state
+
+Jobs become `broken` when their configuration prevents them from running. This might be because their [branch configuration](/docs/pipelines/branch-configuration) doesn't match the build's branch, or because a conditional returned false. This is distinct from `skipped` jobs, which might happen if a newer build is started and build skipping is enabled. Broadly, jobs break because of something inside the build, and are skipped by something outside the build.
```
karensawrey, line 103:
```suggestion
Jobs become `broken` when their configuration prevents them from running. This might be because their [branch configuration](/docs/pipelines/branch-configuration) doesn't match the build's branch, or because a conditional returned `false`. This is distinct from `skipped` jobs, which might happen if a newer build is started and build skipping is enabled. A rough explanation is that jobs break because of something _inside_ the build and are skipped by something _outside_ the build.
```

karensawrey (approved):

pages/pipelines/conditionals.md.erb
```diff
@@ -218,7 +224,30 @@ The below variables are supported by the `if` attribute.
 	<tr>
 		<td><code>build.env()</code></td>
 		<td><code>String</code>, <code>null</code></td>
-		<td>This function returns the value of the environment passed as the first argument, otherwise <code>null</code> if the environment variable has not been set</td>
+		<td>This function returns the value of the environment passed as the first argument if that variable is set, or <code>null</code> if the environment variable is not set.<br>
+		<code>build.env()</code> works with the following variables only:<br>
```
plaindocs, line 228:
```suggestion
		<code>build.env()</code> works with variables you've defined, and the following <code>BUILDKITE_*</code> variables:<br>
```

PR #1466, opened 2022-04-25
Extra detail on using a bootstrap script

jeremybumsted:
We had a support ticket where we were informed that the details provided in the `Customizing Instances with a bootstrap script` is a bit confusing as it doesn't say where to get the ARN for the Instance role, or how to specify it when creating the IAM Policy.

Words may still need some tweaking to capture what I'm going for here, but my goal for this change is to make things easier to understand and get running for anybody trying to set up a bootstrap script and/or s3 bucket


plaindocs:
I confess I don't really the relation between the `arn iam` you added in the example, the new paragraph and the one after it "After creating the policy, you must specify the policy's ARN in the `ManagedPolicyARN` stack parameter."

PR #1478, opened 2022-05-03
Remove all the BETA

plaindocs:
This should do what it says on the tin.


meiqimichelle (approved):
Outta beta!! 🚀 -- well, prepped to 🚀 when the time comes 🙂

PR #1246, opened 2021-11-12
Add documentation for quirks when building tags

jradtilbrook:
This adds a section about a quirk when triggering builds against tags through the API or a schedule (anything but a webhook from GitHub).

The section looks like:

I also updated the agent configuration page in the `BUILDKITE_GIT_FETCH_FLAGS` section to link to this page and the same from the API page for creating a build section

PR #1488, opened 2022-05-11
Update documentation for building from tags

jayco:
Much like this PR https://github.com/buildkite/docs/pull/1246, I have just tweaked the explanation slightly to be explicit about the event we trigger the build off as it came through as a question in support.

It does mention it in the pipeline, you can see the pick from the changelog https://buildkite.com/changelog/116-build-only-branches-tags-or-both, but I think it's worth making it a little clearer.

I think the wording will need some tweaking, I put in a ref to the Github docs. I will leave it in your capable hands

Looks like:

Screen Shot 2022-05-11 at 12 00 41 pm

### DIFF
--- a/pages/integrations/github.md.erb
+++ b/pages/integrations/github.md.erb
@@ -82,7 +82,7 @@ If you want to run builds only on pull requests, set the _Branch Filter Pattern_
 
 ## Running builds on git tags
 
-Builds are only run for tags when a 'push' event is triggered, see [webhook-events-and-payloads](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#push), to enable this edit the GitHub settings for your Buildkite pipeline and choose the _Build Tags_ checkbox.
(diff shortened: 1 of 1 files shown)


pages/integrations/github.md.erb
```diff
@@ -82,7 +82,8 @@ If you want to run builds only on pull requests, set the _Branch Filter Pattern_
 
 ## Running builds on git tags
 
-To run builds for commit tags, edit the GitHub settings for your Buildkite pipeline and choose the _Build Tags_ checkbox.
+Builds are only run for tags when a 'push' event is triggered, see [webhook-events-and-payloads](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#push), to enable this edit the GitHub settings for your Buildkite pipeline and choose the _Build Tags_ checkbox.
```
plaindocs, line 85:
```suggestion
Builds are only run for tags when a [`push` event is triggered](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#push). To enable builds for `push` events for git tags, edit the _GitHub settings_ for your Buildkite pipeline, and choose the _Build Tags_ checkbox.
```

plaindocs (approved):
Addition looks good to me @jayco - see if you're happy with the changes.
I did a bit more looking around for "commit tags" and reckon we can settle on "git tags".

PR #1491, opened 2022-05-11
Prefix `brew install` with explicit `tap`

ddbeck:
If you run `brew install buildkite/buildkite/buildkite-agent` without having tapped `buildkite/buildkite` before, then you get a warning while Homebrew taps anyway. I've put an explicit `tap` command in to suppress this warning and to make it a little more clear that Buildkite's tooling is not part of the core Homebrew formulae.

Truthfully, I'm not sure this is a real suggestion, but it was the smallest thing I could think of to change so far and I'm really here just exercising the docs contribution workflow for @plaindocs.


plaindocs (approved):
Looks good.

This is an picture perfect example of the step that folks forget to add because they don't test the docs on a brand new machine, only on a brand new contributor.

PR #1492, opened 2022-05-11
Fix broken style guide English variants link

ddbeck:
The link to the article about the differences between en-US, en-GB, and en-AU seems to have gone missing. I've replaced the link to the same article on the Wayback Machine.


plaindocs (approved):
:+1:

PR #1493, opened 2022-05-11
add various combinations of "skip ci" that are accepted by Buildkite

jeremybumsted:
Some customers are not aware that we support a few different forms of `skip ci` to tell Buildkite to not create a build on a commit, such as:

```
[skip-ci]
[ci-skip]
[ci skip]
[skip ci]
```

It'd be great if we could add that extra bit of clarification for future travellers!

### DIFF
--- a/pages/pipelines/skipping.md.erb
+++ b/pages/pipelines/skipping.md.erb
@@ -45,7 +45,7 @@ To manually cancel a job:
 
 ## Ignore a commit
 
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Honestly, I wish we just picked *one* and used it, there isn't much gain from have 4 or 6 here. But sure, it's a good PR :-) Thanks @jeremybumsted

Issue #1487, opened 2022-05-11
💄 Side menu is too narrow only on the Webhooks page


olyism:
How to replicate:
1. Go to the [Webhooks overview page](https://buildkite.com/docs/apis/webhooks) 
1. Resize the browser to around 1024px wide and look at the side menu 👀  the plus icons overlap the text in the menu
1. Resize the browser to around 1440px wide and look at the side menu 👀  the plus icon overlaps the beta badge in Test Analytics
1. Browse to other docs pages and you'll see the widths are different and don't have the aforementioned visual glitches
 
Screenshots:

![Screen Shot 2022-05-11 at 11 23 07 am](https://user-images.githubusercontent.com/7202667/167750233-80dcddc0-0ff8-4b5b-9f59-35a0ba09e0e0.png)

![Screen Shot 2022-05-11 at 11 22 59 am](https://user-images.githubusercontent.com/7202667/167750239-0fe2eb8e-88ea-4329-87d2-fce7371177b2.png)

preflightcookie:
Hey @olyism - I think this specific issue is resolved. Unsure if it was @lachlanhardy 's referenced PR or the nav change but the nav looks right to me now, tested at multiple widths.

<img width="1137" alt="Screen Shot 2022-09-06 at 4 47 22 pm" src="https://user-images.githubusercontent.com/1191571/188566072-19a47f4e-bc6f-483c-b3c6-c3a8688267e4.png">


NB: there is an overflow issue going on in the main body (stay tuned for a PR from me on that one) but it's not impacting the nav at present.

Closing this - feel free to reopen if you can reproduce 👍

olyism:
thanks @preflightcookie. I didn't know this PR existed. Yes this has been resolved with the new sidebar design :)

PR #1469, opened 2022-04-27
Revert "Merge pull request #1368 from buildkite/env-vars"

plaindocs:
This reverts commit 65f9c59812ce0380f2b8342f28b0cfdb0ccf7b5c, reversing
changes made to 57cedfa4b8dc47eef40cec388649b40bacd05a23.

We noticed (thanks @keithpitt) the table wasn't playing nice with the sidebar. FAO @lachlanhardy 

image

I suggest merging this now, and going back to fixing the table issue later.

PR #1368, opened 2022-02-10
New environment variable data format and table

plaindocs:
Move environment variables to YAML for more consistency.

Change display from 

image

to 

image

Before merging 

- [x] some design love
- [x] add examples
- [ ] figure out the "modifiable issue"
- [x] fix the * anchor rspec thing

PR #1498, opened 2022-05-13
Doc-308: revert env vars revert

lachlanhardy:
This resolves [DOC-317](https://linear.app/buildkite/issue/DOC-317):
[https://linear.app/buildkite/issue/DOC-317/💄-side-menu-is-too-narrow-only-on-the-webhooks-page](https://linear.app/buildkite/issue/DOC-317/%F0%9F%92%84-side-menu-is-too-narrow-only-on-the-webhooks-page)

and [DOC-308](https://linear.app/buildkite/issue/DOC-308):
https://linear.app/buildkite/issue/DOC-308/fix-and-revert-the-env-vars-business

The issue was occurring on pages where article content (usually a `table`) overflowed the column. This is why #1469 reverted #1368. And why @olyism found #1487.

Turns out Grid Blow-outs are a thing: https://css-tricks.com/preventing-a-grid-blowout/ 
Correcting that lead to a flow-on issue where the table of contents was no longer sticky on wide screens, because absolute positioning doesn't work in overflow containers. But I found a solution of that too: https://www.bennadel.com/blog/3409-using-position-absolute-inside-a-scrolling-overflow-container.htm

The necessity for the article content to x-scroll on overflow is lame and we should fix it, but that'll take a more comprehensive approach. Meantime, I constrained the sidebar nav width more than it would normally just to give those tables some more space and reduce the scroll.

This PR reverts the #1469 reversion (ie puts the shiny env var code back in) and handles the `table` overflow issue, thus correcting the sidebar width.

You can see the two pages with identified issues here to see how much you hate the x-scroll:
- https://bk-docs-pr-1498.onrender.com/docs/pipelines/environment-variables
- https://bk-docs-pr-1498.onrender.com/docs/apis/webhooks


plaindocs:
The table doesn't handle the in-page-toc now. :/ Is this a case of these three things, pick two? 

image

plaindocs:
I guess I'd expect the content to wrap around the in page TOC. I guess you have to chose between wrapping and sticky? I literally do not know how CSS works.

lachlanhardy:
Wrapping wouldn't work for a table anyway. (it'd wrap around the entire table, rather than particular lines that were sticking out)

Yeah, you can basically choose to preserve the whitespace of the column the whole way down the page or have a floaty hover effect as per my screenshot above with the red. Either way, we just have to workaround content that breaks out of our column width until such time as we design for it holistically (And even then there will always be some edge case that does not work particularly gracefully)

plaindocs:
Ah, I see the problem here. On Firefox I don't have a horizontal scroll bar on this page https://bk-docs-pr-1498.onrender.com/docs/pipelines/environment-variables

plaindocs:
I think I'd rather preserve the white space so main content is never hidden, although I guess that still leaves scope for problematic lon non breaking lines.

@olyism maybe we need to bump up looking at this more holistically.

lachlanhardy:
Based on discussion with @plaindocs, I've gone for the TOC as an overlay on the overflowed content rather than cut off the content. This is a temporary solution while we design something better.

See screeny and here: https://bk-docs-pr-1498.onrender.com/docs/pipelines/environment-variables

image

@olyism I'd love an eye on my box-shadow if you've time for an opinion.

app/assets/stylesheets/public/_docs.scss
```diff
@@ -419,6 +421,9 @@ label[for='navicon'] span { // hide alt text
     top: 0;
 
     .Docs__toc__sticky {
+      background: #fefefe;
+      box-shadow: 2px 2px 4px 5px #eee;
```
olyism, line 425:
```suggestion
      box-shadow: 2px 2px 5px 3px rgba(0 0 0 / 10%);
```

2 tweaks please 💄 💅
- Soften the box-shadow with more blur than spread (5px then 4px)
- Use transparency in shadow over opaque colour (replace `#eee` with `rgba(0 0 0 / 10%)`)

olyism (changes requested):
Thanks @lachlanhardy looking good. Please see my suggestion on the box shadow :)

olyism:
ah… this is the first time I'm using the suggesting feature in GitHub. I did a commit. Not sure if this is the right etiquette (or should the person who raised the PR commit)? Hope I've done this correctly. Thanks 🙏

plaindocs:
Here are some things to integrate

- [ ] https://github.com/buildkite/docs/pull/1515

app/assets/stylesheets/public/_docs.scss
plaindocs: Looks like there is an issue with that one @olyism 

```
May 20 08:32:09 AM   #18 4.204 SassC::SyntaxError: Error: overloaded function `rgba` given wrong number of arguments
May 20 08:32:09 AM   #18 4.204         on line 425:35 of app/assets/stylesheets/public/_docs.scss
May 20 08:32:09 AM   #18 4.204         from line 4:9 of app/assets/stylesheets/application.scss
May 20 08:32:09 AM   #18 4.204 >>       box-shadow: 2px 2px 5px 3px rgba(0 0 0 / 10%);
May 20 08:32:09 AM   #18 4.204
May 20 08:32:09 AM   #18 4.204    ----------------------------------^
May 20 08:32:09 AM   #18 4.204 /app/app/assets/stylesheets/public/_docs.scss:425
May 20 08:32:09 AM   #18 4.204 /usr/local/bundle/gems/sassc-2.4.0/lib/sassc/engine.rb:50:in `render'
```

lachlanhardy:
I've corrected the syntax error in @olyism's suggestion. Result looks like:

image

I find that intense. Is that what you were after, Oly?

lachlanhardy:
@plaindocs Do you want to sort out the copy integration?

plaindocs:
yeap, will do. I was just looking at it and wondering why I was seeing a different merge commit than expected. :-D

plaindocs:
Where are we at with this one? @lachlanhardy @olyism I think if one of you sorts out the SCSS and we're all happy we can merge this back in?

lachlanhardy:
@plaindocs @olyism I still reckon that dropshadow should be lighter. Did you want me to change it?

lachlanhardy:
@plaindocs Copy changes from #1590 added and I've lightened the box-shadow

Screen Shot 2022-06-15 at 17 47 20
.

plaindocs (approved):
:shipit:

olyism (approved):

PR #1499, opened 2022-05-16
💄 Better consistency and legibility for body text

olyism:
The app and marketing site uses `1.4` ratio for text line-height. I've also tweaked this on Docs for consistency. Furthermore, although subtle, `1.4` line-height is more legible as it gives the different lines of text more breathing room.

Example:

| Before | After |
| ------ | ----- |
| Screen Shot 2022-05-16 at 5 18 12 pm | Screen Shot 2022-05-16 at 5 18 19 pm |
| Screen Shot 2022-05-16 at 5 14 18 pm | Screen Shot 2022-05-16 at 5 14 54 pm |

Note: I've kept the table of contents right nav at `1.2` as the width is quite narrow and tighter line-height looks better. Furthermore, some pages have really long right nav so I'd like to keep this shorter if possible. Will look into a better UX solutions for addressing navigations later.

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -389,6 +389,8 @@ label[for='navicon'] span { // hide alt text
 }
 
 .Docs__toc {
+  line-height: 1.3;
+
   p { margin-bottom: .5em; }
   .Docs__toc__list {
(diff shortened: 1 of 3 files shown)


lachlanhardy (approved):
Nice!

PR #1511, opened 2022-05-19
Add settings conveniences for Markdown 

ddbeck:
This PR is a bit speculative. Feel free to reject some or all of this!

The Markdown files in this repo are `*.md.erb`. In normal operation, neither GitHub nor VSCode recognize these as Markdown. This PR sets configuration to override ~both~ one of them. That is to say, this PR:

1. Supplies a VSCode workspace settings file with `files.associations` set for the Markdown files.
2. ~Adds a `.gitattributes` file with [github-linguist](https://github.com/github/linguist) directives to do a similar thing to GitHub.~

~The latter change is hard to demonstrate and verify ahead of time. Right now, we don't get GitHub's rich diffs and the like for Markdown files. This is supposed to fix that, but it [looks kinda… off](https://github.com/buildkite/docs/blob/ddbeck/linguist-attributes/pages/integrations/git.md.erb) on my branch, but perhaps that's because rendering is a very low priority background job. If it remains busted a few hours after merging, then we'll have to revert it. (And, though it should go without saying, if you don't like the rich previews then we shouldn't do it.)~


karensawrey:
Not something to be deployed on a Friday :D

plaindocs:
The rich preview is still busted. Shall we remove that? 

I can't see any downsides to the vscode thing though. Does it load automatically?

ddbeck:
> I can't see any downsides to the vscode thing though. Does it load automatically?

That's right. The only problem this might pose is if someone already has a settings file (or wants to extend it and finds Git's tracking a nuisance). There are ways of dealing with that (e.g., a command to generate the settings file, only if it doesn't exist). But odds are… we ain't gonna need it.

I rolled back the attributes change. No rich diffs for now.

plaindocs (approved):
:+1:

karensawrey (approved):

PR #1517, opened 2022-05-20
Adding info on 202 Accepted by REST

karensawrey:
Added the info to the only document in TA that mentions the 202 response. @meiqimichelle - is this what we're looking for? Should I add a note to the REST docs, too?

### DIFF
--- a/pages/test_analytics/importing_json.md.erb
+++ b/pages/test_analytics/importing_json.md.erb
@@ -238,6 +238,20 @@ The JSON format is not yet documented, but the following example is what is prod
 ]
 ```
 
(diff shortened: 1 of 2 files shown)


meiqimichelle:
Yes, please add this to the REST docs as well @karensawrey ! Thank you, looks really good!

meiqimichelle:
@karensawrey I may have gotten wires crossed -- I meant add to the JSON part of the Test Analytics docs. Not sure there's a need to start updating the Buildkite-wide API docs yet! (I think that might be more difficult than we need right now, since the entire information architecture is build around there only being one product at Buildkite)

meiqimichelle (approved):
🚀

PR #1521, opened 2022-05-23
Add or explicitly omit TOCs from all pages

ddbeck:
I started outlining a page and ended up shaving this yak instead. This PR requires all pages to have a `{:toc}` or to explicitly opt-out of having a table of contents.

Specifically, this requires pages to have one and only one of:

- `{:toc}`
- `{:notoc}`

The partial `_*.md.erb` files are exempted from this requirement.


ddbeck:
Ah, so this is a little interesting: the linting fails because it's running an older version of Vale, which was [fixed in v2.15.0](https://github.com/errata-ai/vale/releases/tag/v2.15.0). I'll open a separate PR with the upgrade.

plaindocs:
@ddbeck I'm not really here, but what do you think about `{:notoc}` instead?

plaindocs:
`pages/pipelines/_*` and `pages/agent/*/help/*` are all includes and can be excluded also. Could probably just ignore anything starting with `_` in fact.

ddbeck:
Merged the Vale upgrade and I'm finally getting that green check mark 🚀

karensawrey:
Looks very valid to me, but I'd rather have @plaindocs look at the current implementation if/when possible.

plaindocs (approved):
Love it! 

Care to add a mention in the style guide as well so folks aren't caught out.

ddbeck:
> Care to add a mention in the style guide as well so folks aren't caught out.

Good idea! Take a look at the proposed text?

styleguide/STYLE.md
```diff
@@ -323,6 +323,10 @@ Adding the class `has-beta-pill` to any element will append the beta pill. This
 To generate a table of contents from all your \##\-level headings, use `{:toc}`.
 Make sure there are no spaces after the `{:toc}` - spaces immediately after this custom element are known to break the script.
 
+To omit a table of contents, use `{:notoc}`. Typically `{:notoc}` is for pages where the text immediately following \#-level is the body of the page or where there's only one \##\-level heading, which looks lonely in a table of contents.
```
plaindocs, line 326: <3 lonely <3

styleguide/STYLE.md
```diff
@@ -323,6 +323,10 @@ Adding the class `has-beta-pill` to any element will append the beta pill. This
...
+
+A page can have one and only one of `{:toc}` or `{:notoc}`.
```
plaindocs, line 328: I have a person preference for 

> A page must have either `{:toc}` or `{:notoc}`

plaindocs (approved):
Looks good. Optional suggestion below.

styleguide/STYLE.md
ddbeck: Glad you appreciated this (which also caused me to notice this oddity)

```suggestion
To omit a table of contents, use `{:notoc}`.
Typically `{:notoc}` is for pages where the text immediately following the \#-level heading is the body of the page or where there's only one \##\-level heading, which looks lonely in a table of contents.
```

styleguide/STYLE.md
ddbeck: I like that too. 👍 

```suggestion
A page must have either `{:toc}` or `{:notoc}`.
```

PR #1522, opened 2022-05-23
Upgrade Vale to v2.17.0

ddbeck:
This upgrades Vale to the latest version (the one that's likely to be available through your package manager) and fixes a bug that's blocking https://github.com/buildkite/docs/pull/1521.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -32,8 +32,8 @@ steps:
     steps:
       - label: ":lint-roller: Linting text and markdown"
         commands:
-          - wget https://github.com/errata-ai/vale/releases/download/v2.12.1/vale_2.12.1_Linux_64-bit.tar.gz
-          - tar --transform='s/vale/vale-bin/'  -xf 'vale_2.12.1_Linux_64-bit.tar.gz'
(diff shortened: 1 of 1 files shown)


ddbeck:
Ugh, v2.16.1 introduces… something that breaks the escaping for erb? Investigating.

https://github.com/errata-ai/vale/compare/v2.16.0...v2.16.1

plaindocs (approved):
Nice one!

Weird little bug eh? 

I wonder if one can go about configuring dependabot for this kind of extremely manual dependency? :thinking:

ddbeck:
> I wonder if one can go about configuring dependabot for this kind of extremely manual dependency? 🤔

The options here aren't great, as far as I can tell. The only thing that looks likely: if the Buildkite docker plugin can refer to a `Dockerfile` instead of pulling a named image, then you could use `FROM jdkato/vale:v2.17.0` in something like `vale.Dockerfile`. Dependabot could then upgrade Vale via the via Dockerfile.

Otherwise, I think we're out of luck, since there's no package ecosystem that Vale directly reports into (that's golang for you).

plaindocs:
@ddbeck looks like https://github.com/buildkite-plugins/docker-compose-buildkite-plugin would do the trick.

PR #1531, opened 2022-05-27
💄 Link logo to homepage

olyism:
Link Buildkite logo in the header to the homepage. 

This makes it easier for customers to find the link back to our homepage. This is an interim design improvement until we have a new header design (which we may adopt from the Marketing site, but to be discussed or confirmed later).

https://www.loom.com/share/101d5aaa93644424b0c665cb4cd7523c

Changes:
- Add hyperlink to logos
- "Buildkite Docs" header text still links to the Docs index page
- a11y improvements
- Hover effect on logo to indicate these are linkable
- Refactor: use `$transition-speed` variable for all transition instances

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -333,7 +333,7 @@ label[for='navicon'] span { // hide alt text
   font-weight: bold;
   border-radius: 3px;
   text-decoration: none;
-  transition: background-color 300ms;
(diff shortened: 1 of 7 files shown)


plaindocs (approved):
YAY to A11y :tada: 
The thing does what the loom says, so :+1:

lachlanhardy (approved):
Lovely!

:shipit:

PR #1536, opened 2022-05-30
Refactor Test Analytics collector environment variables docs

ddbeck:
This moves the test collector environment variable coverage out of the _Ruby Collectors_ page, into a standalone reference about what kind of information test collectors read from the environment and upload to Buildkite.

In this framing, the Buildkite collectors do ✨magic✨ and this doc discloses what is or is not magical, but not how the magic works (to transform certain environment variables into URLs, for instance).

This approach raises new and troubling questions; I'll self review to highlight some interesting issues.


pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,55 @@
+# CI Environments
+
+Buildkite Test Analytics collectors automatically detect common continuous integration (CI) environments.
+If available, test collectors gather information about your test runs, such as branch names and build IDs.
```
ddbeck, line 4: There's an open product/style question: "collectors" or "test collectors"? Both have been suggested in Basecamp discussions.

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,55 @@
+# CI Environments
+
+Buildkite Test Analytics collectors automatically detect common continuous integration (CI) environments.
```
ddbeck, line 3: This whole thing sort of implies we ought to have a _collectors_ page that describes the concept in general, then lists all of the collectors: RSpec, Jest, `collector.sh`, and more that come along. It'd also be a good place to link to third-party collectors, when those eventually happen.

```suggestion
Buildkite Test Analytics [collectors](/docs/test-analytics/collectors) automatically detect common continuous integration (CI) environments.
```

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,55 @@
...
+If available, test collectors gather information about your test runs, such as branch names and build IDs.
+Test collectors gather information from the following CI environments:
+
+- [Buildkite](#buildkite)
+- [CircleCI](#circleci)
+- [GitHub Actions](#github-actions)
+
+If you use a test collector in any other environment, the test run will be assigned a unique identifier and attributed to a "generic" CI system.
+Other information, such as branch names or commit hashes, will not be collected.
```
ddbeck, line 12: If we want to emphasize this piece a little more, we could move this to a section — "Using other CI systems" perhaps — so that it's less "only these a supported" and more "we make it easy for these, but it's possible with others."

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,55 @@
...
+
+If you're not using a test collector, see [Importing JSON](/docs/test-analytics/importing-json) and [Importing JUnit XML](/docs/test-analytics/importing-junit-xml) to learn how to provide run environment data.
+
+{:toc}
+
+## Buildkite
```
ddbeck, line 18: For each of the CI provider sections, I've tried to adopt the terminology that gets used by the CI provider, rather than Buildkite's, providing a little bit of a bridge between various terminologies.

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,55 @@
...
+- [Buildkite](#buildkite)
+- [CircleCI](#circleci)
+- [GitHub Actions](#github-actions)
+
+If you use a test collector in any other environment, the test run will be assigned a unique identifier and attributed to a "generic" CI system.
+Other information, such as branch names or commit hashes, will not be collected.
+
+If you're not using a test collector, see [Importing JSON](/docs/test-analytics/importing-json) and [Importing JUnit XML](/docs/test-analytics/importing-junit-xml) to learn how to provide run environment data.
+
+{:toc}
```
ddbeck, line 16: The text above sorta creates two ToCs. I'm not in love with that — it's probably a bit clunky if you're using a screen reader — but the lack of a sidebar `{:notoc}` makes the lack of a sidebar ToC a bit of a loss.

pages/test_analytics/ruby_collectors.md.erb
```diff
@@ -49,79 +49,4 @@ If you're already using RSpec for your tests, add the `rspec-buildkite-analytics
     $ git push
     ```
 
-5. Make sure that the Test Analytics [environment variables](#environment-variables) are set so that the RSpec collector can use them in your Test Analytics dashboard.
```
ddbeck, line 52: One less thing to do! 🚀

pages/test_analytics/ci_environments.md.erb
plaindocs: Makes sense. Not sure how much there is to say about them though..

pages/test_analytics/ci_environments.md.erb
plaindocs: Let's stick with the first for now, as other wise we'll have "Test Analytics Test Collectors".

pages/test_analytics/ci_environments.md.erb
plaindocs: Here I'd be tempted to move the Others into a heading, use the built in TOC only, and leave the intro text more generic. What do you reckon?

pages/test_analytics/ci_environments.md.erb
plaindocs: See above.

pages/test_analytics/ci_environments.md.erb
plaindocs: :+1:

plaindocs (commented):
Definitely better than what we currently have.

The other thing we dropped is mentioning that if you're using containers, you need to make sure the container has access to the relevant variables. I think that fits into the refactor well enough, although I've not figured out how exactly.

plaindocs:
@meiqimichelle we're looking at this approach to fix Environment Variables and CI Environments in one fell swoop.

pages/test_analytics/ci_environments.md.erb
meiqimichelle: +1 to @plaindocs. We've been trying to align on 'collectors' (but have likely not eradicated other words). In calls with customers, folks seem to be able to understand what we mean with that as well when its in the context of Test Analytics already.

pages/test_analytics/ci_environments.md.erb
meiqimichelle: Ah right. There's no collectors homepage, is there? In one of Oly's mockups, we do have subheadings in the left nav, and could easily make a page that has like one sentence -- 'Use collectors to tend test information to Test Analytics' or some such, re-use what I think is a reusable template Rails thingo that Oly made to list collectors on the TA homepage.

There's no a lot to say about them, but this ^^ might help with UX traveling around this part of our docs. 🤔

pages/test_analytics/ruby_collectors.md.erb
meiqimichelle: About this one -- I think we *do* need something on the individual collector pages that help people understand more about env vars. The biggest customer support reach out now is people believing that TA is broken, when really they're missing env vars.

(And then there's the group of people that don't even bother reaching out to customer support, and I just happen to notice that things aren't working right when I look at their setups. They're just sitting there thinking that Test Analytics sucks 😆)

So, while TA will technically work with these instructions (ie, you will see graphy type things in the UI) using only the one required var, if the magic doesn't work (which is not uncommon on first setup because containers), TA will break -- ie, parallel runs won't de-dupe, etc -- and then people either give up on Test Analytics, or contact support.

🤔 Maybe it's a final step that's like "confirm things are working properly"?

🤔 Also, part of the solution here may be to change this sentence:
`Add the Test Analytics code to your application in spec/spec_helper.rb, and set the environment variable [securely](http://localhost:3000/docs/pipelines/secrets) on your agent or agents.`

"The" env var, to me, implies that there are no more to set, when in reality, there are -- even if we're setting them magically, there's more going on behind the scenes. 🤔

pages/test_analytics/ci_environments.md.erb
meiqimichelle: Yes, for sure! Marketing has been emphasizing that Test Analytics will work with any CI. Should find a home for the generic version of CI, and your suggestion to put this is a new section called Other makes sense to me.

`If you're using other CI providers, you need to configure these environment variables yourself. The only required environment variable is the unique key for the build that initiated the Test Analytics run (BUILDKITE_ANALYTICS_KEY), but some features of Test Analytics aren't available if the other environment variables are not set:`

```
Variable | Description and source
-- | --
BUILDKITE_ANALYTICS_KEY | a unique key for the build that initiated the Test Analytics run
BUILDKITE_ANALYTICS_URL | URL that links to the build
BUILDKITE_ANALYTICS_BRANCH | the branch or reference that this run is for
BUILDKITE_ANALYTICS_SHA | the commit SHA for the head of the branch
BUILDKITE_ANALYTICS_NUMBER | the build number of the build
BUILDKITE_ANALYTICS_JOB_ID | the id of a job within the build
BUILDKITE_ANANLYTICS_MESSAGE | the commit message for the head of the branch
```

meiqimichelle:
This is looking so good @ddbeck and @plaindocs ! I left comments but also -- perhaps address the more straightforward comments, ship, and then have another go at the others in a separate PR. All good!

pages/test_analytics/ci_environments.md.erb
meiqimichelle: LOLz so this convo progressed today to where.....we've renamed the actual collector packages to `buildkite-test-collectors` and we're in the process of renaming repos to things like `buildkite/test-collector-javascript` 🙃 .

Whether or not we're able to, or want to, carry those words through all our docs before launch -- I leave in your capable hands. (ie, Test Analytics Test Collectors still sounds silly, maybe it's not necessary to use the full name when used in certain contexts -- but I haven't tested this in docs to have a useful opinion.)

plaindocs: I saw that update. I still think we can stick with `collectors` in the wording when the package is `test-collector-javascript`. At least for now.

pages/test_analytics/ci_environments.md.erb
plaindocs: Yeah. I don't feel strongly about it, but would suggest merging without for now.

pages/test_analytics/ci_environments.md.erb
ddbeck: OK, I can work with this.

data/test_analytics_run_env.yaml
```diff
@@ -0,0 +1,45 @@
+keys:
```
ddbeck, line 1: OK, I know the table is a bit gnarly and also we're likely to recapitulate everything in an endpoint reference, so I'm future proofing by stuffing this info into YAML. You're welcome/I'm sorry.

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,108 @@
...
+    --env BUILDKITE_JOB_ID \
+    --env BUILDKITE_BRANCH \
+    --env BUILDKITE_COMMIT \
+    --env BUILDKITE_MESSAGE \
+    --env BUILDKITE_BUILD_URL \
+    bundle exec rspec
+```
+
+Review the following sections for the environment variables expected by test collectors.
+
```
ddbeck, line 38: 🚨  New section on containerization! 🚨 

Since we can't really know what ridiculous things people are doing with containers (or even what container engine they're using), I've illustrated the concept with plain Docker. Hopefully this gets the point across?

pages/test_analytics/ci_environments.md.erb
```diff
@@ -0,0 +1,108 @@
...
+| Environment variable | Description                                             |
+| -------------------- | ------------------------------------------------------- |
+| `GITHUB_ACTION`      | the name of the action running or its step ID           |
+| `GITHUB_REF`         | the ref (branch or tag) that triggered the workflow run |
+| `GITHUB_REPOSITORY`  | the repository owner and repository name                |
+| `GITHUB_RUN_ATTEMPT` | the numbered attempt of the workflow run                |
+| `GITHUB_RUN_ID`      | the unique number for the workflow run                  |
+| `GITHUB_RUN_NUMBER`  | the cumulative number of runs for the workflow          |
+
+## Other CI providers
```
ddbeck, line 78: 🚨  New section on manually providing environment variables! 🚨 

This was more interesting than I first imagined. Unlike the other sections, where we just need to list out the preexisting variables, here we need to tell folks what to set those variables _to_, which calls for examples. And then I realized that we'll do need to the correspondence to the API. And, well, this table got a little out of hand. There's probably a prettier way to show this stuff (perhaps Oly will have thoughts), but this is close to good enough for now, I think.

pages/test_analytics/ruby_collectors.md.erb
ddbeck: I think I understand the container things now that I've written about it. 😅 

I've partly addressed this with ba2bd0278a2a78134521a6a00104ca0425693175, but it would be nice to have the post-procedure success crtieria (though not strictly blocking).

ddbeck (commented):
I've pushed a bunch of changes and I've highlighted some new and changed things with a self-review.

pages/test_analytics/ci_environments.md.erb
plaindocs: Agree

pages/test_analytics/ruby_collectors.md.erb
plaindocs: My understanding is that it is as simple as folks not seeing corect links to their CI builds (in whatever supported CI) because those vars weren't available when they expected them to be, because they weren't passed through to the container running the collector.

data/test_analytics_run_env.yaml
plaindocs: <3

pages/test_analytics/ci_environments.md.erb
plaindocs: :+1:

plaindocs:
Nice work @olyism - we're all cowboys and cowgirls this week.

meiqimichelle:
This all looks amazing, y'all! 😍😍😍

@ddbeck -- you were asking about how folks can confirm that env vars are coming through.  One give-away is the metadata displayed throughout the Test Analytics UI, and the easiest place to see this all in one screenshot is at the top of individual run pages:

Screen Shot 2022-06-01 at 9 16 12 PM

When env vars aren't set correctly, folks will only see the pass/fail chiclet, timestamp directly next to it, and a UUID for the run name (instead of the human-readable title for the run name you see above). The build name/link, branch, and commit sha will not display at all.

I'm not sure how/if this should be incorporated into docs -- but sharing in case it's helpful.

plaindocs:
Any opinions on this as a table format? 
image

PR #1537, opened 2022-05-30
Fix errant `Github` to `GitHub`

ddbeck:


karensawrey (approved):
Nice catch!

plaindocs:
Again, why isn't the linter catching this?

karensawrey:
But is it case sensitive? 🤔

plaindocs:
Should be.

ddbeck:
As far as I can tell, the spell check doesn't work against words containing `'` (apostrophe). Try misspelling any word you like, then make it possessive: it passes Vale's spell check.

vale/Buildkite/existence.yml
```diff
@@ -7,3 +7,5 @@ swap:
   whitelist: allowlist
   blacklist: blocklist
   oAuth: OAuth
+  'github\.com': github.com
+  github: GitHub
```
ddbeck, line 11: This addition to `existence.yml` will catch this in the future (and others like it) but not `github.com` (which appears in a bunch of places.

plaindocs:
That check's (!) out but is weird. I'm surprised that nobody else has spotted this already. So could it be something to do with our erb format? 

The only apostrophe related issues I'm see are `adoc` related. https://github.com/errata-ai/vale/issues/206 

I also mildly curious as to whether updating the really old dictionary style to the newer one would give us better results. https://vale.sh/docs/topics/styles/#spelling (but not curious enough to do it)

karensawrey:
Yes, from experience ``` will make Vale turn a blind eye. I, however, thought that it was intentional so as not to make Vale take apart code snippets & freak out.

plaindocs:
@karensawrey we're talking `'` not `` ` `` though, right?

(don't ask how long it took me to escape those backticks correctly)

ddbeck:
OK, don't merge this yet. I'll see if I can't reproduce this in a very plain environment. If it's a bug in Vale, we can report it; if it's a bug in our config, I'm guessing we can slice away at the config until we find the cause.

ddbeck:
While trying to sort this out, I found I was getting different behavior locally versus CI (it might be fixed in the latest release). This revealed that Dependabot wasn't upgrading us; I opened https://github.com/buildkite/docs/pull/1580 to fix _that_.

PR #1545, opened 2022-05-31
Add Test Analytics JSON import format reference

ddbeck:
This is a WIP, a skeleton really, to show one possible approach for describing the JSON format. It breaks it down into three component types: a result object, one or more history objects, and (presumably many) span objects. Treat this as a concept and then we can proceed toward filling in all the gaps (and moving it to better format, probably YAML, to give us flexibility to change the presentation).

### DIFF
--- a/pages/test_analytics/importing_json.md.erb
+++ b/pages/test_analytics/importing_json.md.erb
@@ -47,196 +47,7 @@ curl --request POST \
     }
 ```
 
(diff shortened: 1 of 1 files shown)


plaindocs:
@blaknite @meiqimichelle if you have any more info we can use to complete this, or can point at anything that needs fixing before we can get this out we can work on it tomorrow.

meiqimichelle:
@plaindocs @ddbeck -- very happy to have what you've got here as a great improvement on the existing docs! Could we ship as a v1? (Also, we're unlikely to have much time to offer significant before GA.)

PR #1548, opened 2022-06-01
Add specific Buildkite examples

plaindocs:
What do we think about this way of factoring in these specific examples from https://gist.github.com/toolmantim/f4760749851df031fb0eae69e989a685 

@ddbeck @meiqimichelle 

- [ ] GA
- [ ] CircleCI


plaindocs:
Definitely planning to link them in as well, I'll tag you for review when I've looked at where that makes sense to do.

toolmantim:
👍🏼

It's sort of related to https://github.com/buildkite/docs/pull/1526#issuecomment-1143494792 too — if not we're recommending people slurp our shell script that does the magic env figuring out, which I'm recommending we don't for now, it's even more important to provide people the right snippet for the CI system on the JUnit page itself I reckon.

PR #1556, opened 2022-06-02
Change TEST_ to BUILDKITE_

meiqimichelle:
Quick search and replace across docs to remove an outdated env var name:

~~TEST_ANALYTICS_TOKEN~~

BUILDKITE_ANALYTICS_TOKEN


scopegate (approved):
🚀

PR #1566, opened 2022-06-03
Start of a FAQ for TA

jayco:
We need an FAQ to help with questions rolling in from customers - here is a rough draft that you can take and expand on. It needs a lot of love - but feel free to make it you own.

Screen Shot 2022-06-03 at 2 32 43 pm


meiqimichelle:
Convo about this in Basecamp here: https://3.basecamp.com/3453178/buckets/1713315/chats/269048534@4999498889

tl;dr I'd like to avoid an FAQ section in favor of in-context improvement to other areas of docs and the product (ie, this could be better solved with an improved error message). However, the customer support team want to be able to action things quickly. Open question: what's the best way to do this?

plaindocs:
Folks, the history for this is lost in the chat - @jayco let's chat about this next week if there is anything still outstanding.

PR #1568, opened 2022-06-05
💄 Improve responsive table to fix layout bug

olyism:
Currently page layout distorts when an inline string exceeds the container width, especially when used within a table. This makes the Docs site look visually broken and adds unnecessary horizontal scrolling.

Here's the Loom 🎦 going through the rationale https://www.loom.com/share/0bcf4a7256cb459e83ed8eb654b68720

Or you can see the TL;DR bullet list here:

- **Improved responsive table behavior** where the table is stacked on small screens
  - Inspired by the stacked mode in https://github.com/filamentgroup/tablesaw
  - Uses JavaScript to find all instances of `table.responsive-table` and render responsive effect
- Use code blocks, instead of inline `<code>`, when we have a very long string
- Increase the stacking breakpoint
- Refactor: gradually introduce Sass variables with screen size

### Current behavior

Screen Shot 2022-06-06 at 9 57 57 am

### After

Screen Shot 2022-06-06 at 9 58 42 am

Screen Shot 2022-06-06 at 9 58 56 am


app/assets/stylesheets/public/_variables.scss
```diff
@@ -3,3 +3,5 @@ $text-on-dark-subdued-opacity: .8;
 $text-on-dark-very-subdued-opacity: .7;
 $color-brand-green: rgb(2, 129, 76);
 $transition-speed: 300ms;
+$screen-sm: 768px;
+$max-w-prose: 65ch;
```
olyism, line 7: `65ch` width makes each line of text roughly 9–12 words, which is the optimal reading in tradition typography design.

This is inspired by TailwindCSS's `$prose` length. Setting character length is way simpler and interoperable than using JavaScript to set word width.

app/assets/javascripts/responsive_table.js
```diff
@@ -0,0 +1,25 @@
+(function () {
```
lachlanhardy, line 1: Did you know the app typically uses a custom Markdown rendering file for changes like this: https://github.com/buildkite/docs/blob/6859c165c5cf38cae6575bf8efad9ee8bead1ad3/app/models/page/renderer.rb

It looks like not all of them are Markdown tables, though, so maybe the JS is the answer. Happy to talk this through if you want to hash out the pros and cons.

lachlanhardy (approved):
One significant question, otherwise this looks great to me.

app/assets/javascripts/responsive_table.js
olyism: thanks @lachlanhardy no I wasn't aware of `renderer.rb`

This is handy! I would prefer to avoid JavaScript. I'll put this PR back into draft and have another go.

karensawrey (approved):
From the content standpoint - LGTM. Great job!

lachlanhardy (approved):
Looking awesome, Oly!

PR #1570, opened 2022-06-06
💬 Link to offical collector GitHub repo

olyism:
Pushing up a suggestion to link Collectors to the respective GitHub repo.

I've chatted with different people who agree that linking Collectors on Docs to the respective GitHub repo is helpful. Customers can see the Collectors' source code if needed and there are additional information on the repo as well.

That being said, there's an argument to having a 'single source of truth' for documentation, which is helpful for docs findability and maintainability.

I'm trying out a small improvement here to simply add a text link, nothing fancy for now.

Open to your thoughts on this.

### DIFF
--- a/pages/test_analytics/js_collectors.md.erb
+++ b/pages/test_analytics/js_collectors.md.erb
@@ -1,13 +1,14 @@
 # JavaScript Collectors
 
-Test Analytics ships with a Jest test collector.
+Test Analytics ships with JavaScript collectors :github: [`test-collector-javascript`](https://github.com/buildkite/test-collector-javascript).
+
 You can also upload test results by importing [JSON](/docs/test-analytics/importing-json) or [JUnit XML](/docs/test-analytics/importing-junit-xml).
 
 {:toc}
 
 ## Jest collector
 
-[Jest](https://jestjs.io/) is a JavaScript testing framework.
+Test Analytics ships with a [Jest](https://jestjs.io/) test collector.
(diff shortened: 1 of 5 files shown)


karensawrey (approved):
Awesome! Got confused at why "collectorS" for JavaScript, but it checks with the Readme we have there. 
Approved, let's merge!

pages/test_analytics/js_collectors.md.erb
```diff
@@ -1,13 +1,14 @@
...
-Test Analytics ships with a Jest test collector.
+Test Analytics ships with JavaScript collectors :github: [`test-collector-javascript`](https://github.com/buildkite/test-collector-javascript).
+
 You can also upload test results by importing [JSON](/docs/test-analytics/importing-json) or [JUnit XML](/docs/test-analytics/importing-junit-xml).
 
 {:toc}
 
 ## Jest collector
 
-[Jest](https://jestjs.io/) is a JavaScript testing framework.
```
ddbeck, line 10: If we're dropping the _describe what testing thing we're talking about_ language here, then perhaps we should do this for every framework (e.g., the RSpec docs has this same sort of text).

Perhaps this is worth saving for a separate PR. Really, we ought to have a template + checklist in the style guide for , which I'm happy to get started on, but the hard part is consensus.

karensawrey: Let's offload that to a different PR, plz. You can create a newticket for this.

ddbeck: (Though all other things being equal, I'd prefer to see the line change omitted as part of deferring the overall fix.)

PR #1574, opened 2022-06-07
Set identifier as required

pzeballos:
The object identifier is actually required.
I'm also not so sure about the current description, but I think it will need some changes so it's clear that is required


karensawrey:
@pzeballos ship it ;)! 🚢

PR #1578, opened 2022-06-08
👔 Should use support+analytics email

olyism:
Missed this template logic. The support email address should be `support+analytics@buildkite.com` when we're on any on the Test Analytics pages.


ddbeck (approved):
👍

karensawrey (approved):

PR #1579, opened 2022-06-08
🚸 Holisic navigation improvements

olyism:
I wanted to address several problems holistically in this PR (which also resulted in a facelift):

Please see the below 🎦 Loom link or description for summary:

https://www.loom.com/share/42e191943eb94139939a38f6477c5067

1. Left nav too narrow on some breakpoints that causes unpleasant word-breaks
2. Reduce left nav font size so it doesn't take away attention from heading
3. Use typography that's more consistent with the app
4. Make it easier to link back to dashboard or homepage
5. Overall responsive design improvements

Tasks:

- [x] Typography changes so the left nav looks like the app's nav
- [x] Increase stack breakpoint 
- [x] Sticky left nav
- [x] Move home button to header
- [x] Increase burger menu hotspot
- [x] Left nav use chev icon

## Screenshots

Screen Shot 2022-06-08 at 3 22 19 pm

Screen Shot 2022-06-08 at 3 24 03 pm

Screen Shot 2022-06-08 at 3 22 55 pm

Screen Shot 2022-06-08 at 3 23 32 pm


olyism:
Moved back to draft for requested fixes:

- [ ] Prevent main content scrolling once I scroll to the end of the left nav
- [ ] Remove green bar highlighting current section

karensawrey (approved):
Looks great!

olyism:
Thanks @meiqimichelle for taking a look 🙌 I've reduced the chevron size.

Screen Shot 2022-06-08 at 8 44 17 pm

With regards to:

> I've also found the "opening" chevron direction interaction clearer than the up-down one.

If you don't mind I'll have a think about this chevron change 🤔 and do it in an upcoming PR. To start, I'll play it safe and use the convention. I'm unsure about the right arrow as it may be misinterpreted as a link rather than an expando/collapso.

That being said, I think the left nav can be better and speak more about the type of content rather than just a conventional nav. For example [XCode](https://developer.apple.com/documentation/xcode/) and [Notion](https://developers.notion.com/reference/intro) docs have interesting treatment (chev on left, pill and icon affordances). I'll need to make more sense of how our doc content can better serve the developers' needs.

PR #1538, opened 2022-05-30
Make Vale upgradable by Dependabot

ddbeck:
This PR does a few things with the aim of making Vale eligible for upgrading by Dependabot:

- It declares the Vale version via a Dockerfile, so that Dependabot can spot when there's a new version of Vale.
- It uses Docker Compose to run Vale, so that the Buildkite pipeline can use the custom Dockerfile (if this feels a bit roundabout, it's because the Buildkite docker plugin doesn't handle building images)
- It creates a shell script, `vale.sh`, to encapsulate the args for `vale` itself. This has a side-effect of making it more convenient to run Vale locally.

PR #1580, opened 2022-06-08
Exclude some Dockerfiles from Dependabot upgrades

ddbeck:
Dependabot errors out on one of the root Dockerfiles (because `public.ecr.aws` doesn't support an API it requires, see
https://github.com/dependabot/dependabot-core/issues/4212). To avoid this, I've changed the dependabot configuration to only upgrade Vale.

This is a follow-up to https://github.com/buildkite/docs/pull/1538.


karensawrey (approved):
Great! Let's get this rolling.

PR #1584, opened 2022-06-09
🚑 HOTFIX: Search pop up should not be clipped

olyism:
Position absolute children within a `overflow-y: scroll;` element get clipped. The search pop up is wider than the scrolling sidebar and got clipped, making the search unusable.

To fix this, I've moved the search box into the header.


JuanitoFatas (approved):
Confirmed the search box moved to header that is better than currently being clipped 👍🏼 
CleanShot 2022-06-09 at 10 50 51@2x

PR #1587, opened 2022-06-10
Make TimeHelpers from ActiveSupport in RSpec

JuanitoFatas:
This Pull Request sets up RSpec to use https://edgeapi.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html in the tests.

### DIFF
--- a/spec/support/time_helper.rb
+++ b/spec/support/time_helper.rb
@@ -1,0 +1,5 @@
+require "active_support/testing/time_helpers"
+
+RSpec.configure do |config|
+  config.include ActiveSupport::Testing::TimeHelpers
+end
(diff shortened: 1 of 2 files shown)


spec/support/time_helper.rb
```diff
@@ -0,0 +1,5 @@
+require "active_support/testing/time_helpers"
+
+RSpec.configure do |config|
+  config.include ActiveSupport::Testing::TimeHelpers
+end
```
JuanitoFatas, line 5: This is a RSpec way to include a module (module has methods) to all the describe/context/it blocks

https://relishapp.com/rspec/rspec-core/docs/helper-methods/define-helper-methods-in-a-module

shevaun:
Oh, came here to say we should probably stick to TimeCop to be consistent, but I didn't realise this was the docs repo and we don't use TimeCop in here! Carry on then :)

shevaun (approved):

PR #1588, opened 2022-06-10
Structure Test Analytics JSON format reference

ddbeck:
I'm not sure this is genuinely good, but it does eliminate some Markdown tables (🤮) and CSV-filled comments with something a little less awful. There are more opportunities to refactor here, but this is my attempt at starting that process.

### DIFF
--- a/app/models/test_analytics_json.rb
+++ b/app/models/test_analytics_json.rb
@@ -1,0 +1,5 @@
+TEST_ANALYTICS_JSON = {
+  "history" => YAML.load_file('data/test_analytics_json_fields_history.yaml'),
+  "span" => YAML.load_file('data/test_analytics_json_fields_span.yaml'),
+  "test_result" => YAML.load_file('data/test_analytics_json_fields_test_result.yaml')
+}.freeze
--- a/data/test_analytics_json_fields.schema.yaml
+++ b/data/test_analytics_json_fields.schema.yaml
@@ -1,0 +1,33 @@
+---
+type: object
+properties:
+  fields:
+    type: array
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+        required:
+          type: boolean
+        type:
+          type: string
+        enumerated_values:
+          type: array
+          items:
+            type: string
+        desc:
+          type: string
+        examples:
+          type: array
+          items:
+            type: string
+      required:
+        - name
+        - required
+        - type
+        - enumerated_values
+        - desc
+        - examples
+required:
+  - fields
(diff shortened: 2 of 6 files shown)


gchan:
Someone noticed that `expanded_reason` hasn't been documented yet.

I believe the intention of this key is for reporting more detailed failure reasons. Here is an example in our fixtures:

```
    "failure_expanded": [
      {
        "expanded": [
          "  expected: false",
          "       got: true",
          "",
          "  (compared using ==)",
          "",
          "  Diff:",
          "  @@ -1 +1 @@",
          "  -false","  +true"
        ],
        "backtrace": [
          "./spec/models/analytics/upload_spec.rb:25:in `block (3 levels) in \u003ctop (required)\u003e'","./spec/support/log.rb:17:in `run'",
          "./spec/support/log.rb:66:in `block (2 levels) in \u003ctop (required)\u003e'",
          "./spec/support/database.rb:19:in `block (2 levels) in \u003ctop (required)\u003e'",
          "/Users/abc/Documents/rspec-buildkite-analytics/lib/rspec/buildkite/analytics/uploader.rb:153:in `block (2 levels) in configure'",
          "-e:1:in `\u003cmain\u003e'"
        ]
      }
    ],
 ```
 
The current validation expects `failure_expanded` to be an Array of Hashes/Objects

Here is also a similar discussion for the JS collector https://github.com/buildkite/test-collector-javascript/issues/17

**Comment by plaindocs** [comment]
Thanks @gchan, would that be at the same level as ` failure_reason` and optional?

**Comment by gchan** [comment]
@plaindocs Yup, `failure_expanded` is at the same level as `failure_reason`.

Note that I incorrectly called the key `expanded_reason` in my previous post, I have updated it.

**Review by plaindocs — APPROVED** [review]
I agree that it's not perfect, but it's definitely better than going back to edit the markdown tables. 

I've tweaked the UUID description, and half-heartedly added Extended Failure. Although that too needs fixing up and splitting out into a section.

File: app/models/test_analytics_json.rb [hunk 1]
```diff
@@ -0,0 +1,5 @@
+TEST_ANALYTICS_JSON = {
```
plaindocs, line 1: Combining these does make it less obvious which table is generated from where when you're editing the source, so I'd be tempted to split these out.

gchan:
> @gchan are backtrace and expanded just strings from the CLI?

Yes, the keys `backtrace` and `expanded` are strings for the hash/dictionary. There is no restriction on what the strings/keys are.

PR #1590, opened 2022-06-14
Change 'team names' to 'team slugs'

jradtilbrook:
The docs aren't quite accurate in this part. The value that is present in the environment variables is the slugified version of the team name.

It is referred to as the slug in the APIs so changing that here as well.


plaindocs (approved):
Nice one @jradtilbrook

PR #1593, opened 2022-06-15
Add emojicom feedback widget to footer

lachlanhardy:
This resolves DOC-45:
https://linear.app/buildkite/issue/DOC-45/how-helpful-was-this-article

You can view it live on the footer of any page:
https://bk-docs-pr-1593.onrender.com/docs/tutorials/getting-started

image

image

image

We can also choose particular emoji (up to 5) or do our own custom images. @olyism, in particular, I'd love your thoughts here. Emoji is very Buildkite, but perhaps custom icons might be cool?

Recommended image dimensions are 40px × 40px (max). To support high resolution displays upload an 80px × 80px (max). Maximum image size is 20 KB.

I also feel the padding is too large here and I'm tempted to re-write the padding structure of the footer to solve that (since we can't control things inside the widget.

Feedback very welcome!

I'll leave this as draft while we work out what we want to do about whitespace and the images, but we can also launch before we have the image decision and put those in place when ready.

### DIFF
--- a/app/assets/stylesheets/public/_footer.scss
+++ b/app/assets/stylesheets/public/_footer.scss
@@ -55,4 +55,8 @@
     margin-top: 3px;
     width: 16px;
   }
+
+  .emoji-feedback {
+    grid-column: 1 / -1;
+  }
 }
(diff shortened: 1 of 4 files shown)


plaindocs:
Replaced by #1612

PR #1597, opened 2022-06-17
Agent name %random

sj26:
Add documentation for the new `%random` parameter available in agent names.

For the moment this is 6 alphanumeric characters, but that's not guaranteed. It will remain around there. I feel like the other items are vague enough that these words are sufficient.

An example would be: `agent-%random` becomes `agent-RXo43u`.

I've also updated the agent v2 example because it never supported `%spawn`, but it does support `%random` because the latter is done server-side.

Fixes DOC-397.


plaindocs (approved):
Looks good. I'd be tempted to add the `RXo43u` example anyway, even if it isn't guaranteed for perpetuity.

sj26:
Yeah I didn't see any other examples in there which is why I didn't include it. I'll merge this much. If you'd like to add examples afterward that'd be rad!

PR #1600, opened 2022-06-20
💄Improve table of contents styling

olyism:
Small styling tweak for table of contents (TOC) so it aligns better with the overall sidebar look and feel.

## Before

Screen Shot 2022-06-20 at 9 44 37 am

- Drop shadow 
- Body font
- No overflow scroll for exceedingly long TOC

## After

Visual tweaks as shown on my local dev:

- Consistent look and feel as per left sidebar nav
- System font
- Position fixed with overflow scrolling

Screen Shot 2022-06-20 at 12 09 06 pm

### Overflow scroll support

Screen Shot 2022-06-20 at 12 10 58 pm

### Does not affect existing small screen styles

Screen Shot 2022-06-20 at 12 09 18 pm

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -91,21 +91,6 @@ $color-green: rgb(3, 165, 98);
   margin: .25em 0 0 0;
 }
 
-.Docs__toc {
-  line-height: 1.3;
-
-  p { margin-bottom: .5em; }
-  .Docs__toc__list {
-    margin-top: .5em;
-
-    li {
-      color: $color-brand-green;
-      list-style-type: square;
-      margin-left: 1.2em;
-    }
-  }
-}
-
 @media (min-width: $screen-lg) { //TOC as side column on larger screens
   .Docs__article {
     display: grid;
@@ -117,28 +102,6 @@ $color-green: rgb(3, 165, 98);
 
     &--is-landing-page {
       display: block;
-    }
-  }
-
-  .Docs__toc {
-    font-size: 16px;
-    height: 95%;
-    position: absolute;
-    right: 0;
-    top: 0;
-
-    .Docs__toc__sticky {
-      background: #fefefe;
-      box-shadow: 2px 2px 5px 3px rgba(0, 0, 0, 0.3%);
-      padding: 1px 5px 8px 15px;
-      position: sticky;
-      top: $header-height + 30px;
-      width: $sidebar-width;
-    }
-
-    a {
-      color: #666;
-      text-decoration: none;
     }
   }
 }
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -128,8 +128,19 @@ $color-green: rgb(3, 165, 98);
 }
 
 .Docs__attribute__table {
-  colgroup col:nth-child(1) {
-    width: 5rem;
+  table-layout: fixed;
+  width: 100%;
+
+  th {
+    word-wrap: break-word;
+
+    code {
+      white-space: normal;
+    }
+  }
+
+  td {
+    word-wrap: break-word;
   }
 
   tr.importance th {
@@ -192,10 +203,6 @@ $color-green: rgb(3, 165, 98);
 
   .Docs__attribute__env-var {
     font-size: 0.8em;
-
-    code {
-      white-space: break-spaces;
-    }
   }
 
   .Docs__attribute__default {
--- a/app/assets/stylesheets/public/_nav.scss
+++ b/app/assets/stylesheets/public/_nav.scss
@@ -127,6 +127,7 @@
 
     &--level2 {
       @media (min-width: $screen-md) {
+        background-color: #fff;
         border-right: #ddd 1px solid;
         height: calc(100vh - #{$header-height});
         left: 0;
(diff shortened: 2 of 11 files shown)


lachlanhardy:
I can't see the lefthand border on the TOC column in Safari, Chrome, or Firefox, so not sure what happened there.
image
cf https://bk-docs-pr-1600.onrender.com/docs/pipelines/dependencies

Presumably related, but this no longer handles the overlarge table such as: https://bk-docs-pr-1600.onrender.com/docs/pipelines/environment-variables
image


Although it turns out, neither does the live site currently. Right hand cut off here is not cropping but the end of the browser window: 
image

lachlanhardy (commented):
Screenshot is lovely, but as per my comments, I don't see that live. It also fails to handle the overlarge tables on some pages.

app/assets/stylesheets/public/_nav.scss
```diff
@@ -127,6 +127,7 @@
 
     &--level2 {
       @media (min-width: $screen-md) {
+        background-color: #fff;
```
olyism, line 130: In case of content overlap, the white background will cover up what's behind the sidebar

app/assets/stylesheets/public/_search.scss
```diff
@@ -15,6 +15,7 @@
     font-family: inherit;
     font-size: 16px;
     outline: none;
+    margin: 0;
```
olyism, line 18: Safari form elements have native margins. Force this to 0.

app/assets/stylesheets/public/_toc.scss
```diff
@@ -2,6 +2,8 @@
   box-sizing: border-box;
 
   @media (min-width: $screen-lg) {
+    background-color: #fff;
+    border-left: #ddd 1px solid;
```
olyism, line 6: In case of content overlap, the white background will cover up what's behind the sidebar

Fix left border that got accidentally left out.

lachlanhardy:
New version looks great!

app/assets/stylesheets/public/_docs.scss
```diff
@@ -128,8 +128,19 @@ $color-green: rgb(3, 165, 98);
...
-  colgroup col:nth-child(1) {
-    width: 5rem;
+  table-layout: fixed;
+  width: 100%;
+
+  th {
+    word-wrap: break-word;
+
+    code {
+      white-space: normal;
```
olyism, line 138: Wrap and break words within `th` and `td`. Also wrapping inline `code` so elements like long variables break into a new line and avoids distorting the layout.

app/models/page/renderer.rb
```diff
@@ -181,11 +181,13 @@ def init_responsive_tables(doc)
     doc.css('table.responsive-table').each do |table|
       thead_ths = table.css('thead th')
 
-      table.search('./tbody/tr').each do |tr|
-        tr.search('./td').each_with_index do |td, i|
-          faux_th = "<th aria-hidden class=\"responsive-table__faux-th\">#{thead_ths[i].children}</th>"
-          
-          td.add_previous_sibling(faux_th)
+      unless thead_ths.empty?
```
olyism, line 184: Prevent error and skip prepending _faux `th`_ when no `th` are found in responsive tables.

karensawrey:
no-suggestions
Are the search bar suggestions supposed to work in the Render previews?.. If yes, it's a bug. If no, then I'm okay with the rest of the changes in this PR. @olyism

olyism:
hi @lachlanhardy thanks for the feedback. I've looked at this again and propose the following changes and commits to address the issue of table distorting page layout.

## Fix table layout and break word to second line

- Use `table-layout: fixed` so width of table isn't decided by the table cells
- Couple the above with word break in `td`, `th` and `code`

The above effect has been applied to all `.Docs__attribute__table` and `.responsive-table`. 

This does mean that long variable names would break into a second line. It's still legible so for now it shouldn't cause developers too much problem.

@karensawrey thoughts on breaking long variable names? ☝️ 

See commit https://github.com/buildkite/docs/pull/1600/commits/33e8907cf2f1e60e2cf14cd9592482cc45169f4e

### Before

Screen Shot 2022-06-20 at 9 12 47 pm

Table width pushing out the page layout.

### After

Screen Shot 2022-06-20 at 9 13 37 pm

Long words are broken down to second line to preserve the page layout.

## Defensive styling to prevent page layout distortion

I've added CSS tweaks to provide fixed sidebars with solid background color so that, in the event when a page layout gets distorted, it'll gracefully fallback to horizontal scrolling.

I've also added a few Safari tweaks to override some of its browser styles for better cross-browser pixel perfection.

See commit https://github.com/buildkite/docs/pull/1600/commits/1c0af0ad280a9b4461c733b6ed782911edd21a8f

## Future considerations

I think it's worth considering a more holistic approach to address the presentation of table data. There may be ways to better format the tables, use responsive/adaptive technique or even use interactive UI to help developers look up the data they need.

I'll create a backlog ticket to look into this later.

karensawrey (approved):

app/assets/stylesheets/public/_docs.scss
lachlanhardy: There were _reasons_ we didn't do this. Mostly around ease of cutting and pasting variable names or keys into scripts and config files. We never satisfactorily resolved it. Perhaps it is time to revisit those ideas, @plaindocs?

lachlanhardy:
Ah, boo. I wrote a comment above, but you've already addressed it more completely. Sorry about that!

Yes, long-standing issue that needs *some* kind of consistent workaround. Maybe this is it!

karensawrey:
No worries, @lachlanhardy! The point is - it's time we addressed the issue of long variable names breaking things... Not the first time, not the first iteration on tables (@olyism this is really an ongoing struggle that started waaaaay back!).

PR #1607, opened 2022-06-26
searching/downloading artifacts from parallel steps

nickofthyme:
Adds a note to explain the use of `--step` option when searching or downloading artifacts from parallel steps.

See https://forum.buildkite.community/t/download-artifacts-from-parallel-builds/1502/2

### DIFF
--- a/pages/agent/v3/cli_artifact.md.erb
+++ b/pages/agent/v3/cli_artifact.md.erb
@@ -160,7 +160,7 @@ Return a list of artifacts that match a query.
 
 ## Parallelized steps
 
-Currently, buildkite does not support collating artifacts from parallelized steps under a single key. Thus using the `--step` option with a parallelized step key will return only artifacts from the last completed step.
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Nice improvement! Thanks :-)

pages/agent/v3/cli_artifact.md.erb
```diff
@@ -158,6 +158,16 @@ Return a list of artifacts that match a query.
 <%= render 'agent/v3/help/artifact_search' %>
 
 
+## Parallelized steps
+
+Currently, buildkite does not support collating artifacts from parallelized steps under a single key. Thus using the `--step` option with a parallelized step key will return only artifacts from the last completed step.
```
plaindocs, line 163:
```suggestion
Currently, Buildkite does not support collating artifacts from parallelized steps under a single key. Thus using the `--step` option with a parallelized step key will return only artifacts from the last completed step.
```

plaindocs:
@ddbeck FYI

PR #1609, opened 2022-06-27
📈 Replace GA script with Segment Analytics.js

olyism:
This PR replaces the existing Google Analytics script with [Segment](https://www.segment.com). It also uses production profiles for the production site, and development profiles for dev and staging sites.


lachlanhardy:
@olyism I've made a bunch of changes here, that I can walk you through, but are hopefully readable in the commits. They don't work on Render yet, but I think I'm close and this is much more idiomatic Rails.

lachlanhardy:
@olyism Just leaving this here so folks can follow the conversation. The IDs now come into Render via their entry in the UI. I haven't found an effective way to do it in code. I might raise a ticket with their support because I've clearly done something wrong code-wise

render.yaml
```diff
@@ -26,4 +26,5 @@ services:
         value: enabled
       - key: ROBOTS_NO_INDEX # Imported from Heroku app
         value: true
-
+      - key: TEST_ENV_VAR
```
lachlanhardy, line 29: I tried this, but couldn't get it to work.

olyism: Hi @lachlanhardy thanks for trying this again. I've been struggling to get Render to reliably do its thing and no luck. Mainly because I'm struggling to cognitively follow a complicated process to replicate it.

I'm thinking to just do something really simple, like `IF hostname is buildkite.com THEN use production scripts ELSE dev scripts`. Not the most elegant solution but at least it serves the purpose of not contaminating analytics data.

elle:
@olyism This PR has a few moving parts. I'm going to break it apart into separate more thematic PRs, if that's alright with you...

olyism:
hi @elle yes breaking up into secret thematic PRs makes sense 👍

elle:
Closing this in favour of https://github.com/buildkite/docs/pull/1765

PR #1623, opened 2022-07-05
Adding BUILDKITE_PULL_REQUEST_DRAFT env var

karensawrey:
Adding information about BUILDKITE_PULL_REQUEST_DRAFT environment variable to the list of env variables. 

@sj26 I have doubts about the example I've provided here. Please take a look.

### DIFF
--- a/data/environment_variables.yaml
+++ b/data/environment_variables.yaml
@@ -319,7 +319,7 @@ variables:
   desc: |
     This environment variable only gets added for a build with a pull request when that pull request is a draft.
   modifiable: false
-  example: "BUILDKITE_PULL_REQUEST_DRAFT=true"
+  example: "true"
 - name: BUILDKITE_PULL_REQUEST_REPO
   desc: |
     The repository URL of the pull request.
--- a/data/environment_variables.yaml
+++ b/data/environment_variables.yaml
@@ -317,7 +317,7 @@ variables:
   example: "`\"main\"`, or `\"\"` if not a pull request"
 - name: BUILDKITE_PULL_REQUEST_DRAFT
   desc: |
-    This environment variable only gets added for a build with a pull request when that pull request is a draft.
+    Set to `true` when the pull request is a draft. 
   modifiable: false
   example: "true"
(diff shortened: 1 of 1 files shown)


karensawrey:
@pzeballos any chance you could look at the example in this one and verify if it's a correct one?

plaindocs (commented):
I think the format for the other examples doesn't include the environment variable name, so here it would be just `false`.

You can also specify possible values with

```
  values:
    - true
    - false
```

karensawrey:
> I think the format for the other examples doesn't include the environment variable name, so here it would be just `false`.
> 
> You can also specify possible values with
> 
> ```
>   values:
>     - true
>     - false
> ```

Thanks, @plaindocs

data/environment_variables.yaml
```diff
@@ -315,6 +315,11 @@ variables:
     The base branch that the pull request is targeting.
   modifiable: false
   example: "`\"main\"`, or `\"\"` if not a pull request"
+- name: BUILDKITE_PULL_REQUEST_DRAFT
+  desc: |
+    This environment variable only gets added for a build with a pull request when that pull request is a draft.
```
plaindocs, line 320: How about? 

```suggestion
    Set to `true` when the pull request is a draft. 
```

pzeballos: I would add that the variable will be missing otherwise

plaindocs: Why don't we set it to false in that scenario? :-p But yeah, makes sense.

pzeballos: It's on purpose :P the var will only exists when is true because there is a size limit that we may hit so we've being careful with new en vars

plaindocs: TIL. Thanks @pzeballos !

karensawrey: Updated. Plz take a look & let's merge.

sj26:
Sorry I didn't see this earlier, but it looks good now — thanks!

PR #1626, opened 2022-07-06
Adding info about explicit step dependencies

karensawrey:
With uploaded steps entering the frame.

### DIFF
--- a/pages/pipelines/defining_steps.md.erb
+++ b/pages/pipelines/defining_steps.md.erb
@@ -105,7 +105,7 @@ For more example steps and detailed configuration options, see the example `pipe
 
 If your pipeline has more than one step and you have multiple agents available to run them, they will automatically run at the same time. If your steps rely on running in sequence, you can separate them with [wait steps](/docs/pipelines/wait-step). This will ensure that any steps before the 'wait' are completed before steps after the 'wait' are run.
 
-<div class="Docs__note">
+<div class="Docs__troubleshooting-note">
   <h3>Explicit dependencies in uploaded steps </h3>
   <p>A pipeline upload step will force any uploaded steps to become explicit <a href="/docs/pipelines/dependencies">dependencies</a> of the steps that depend on that pipeline upload step. For example, if step B depends on step A, and step A uploads step C, then step B will also depend on step C.</p>
 </div>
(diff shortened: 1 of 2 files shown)


pages/pipelines/defining_steps.md.erb
```diff
@@ -105,6 +105,11 @@ For more example steps and detailed configuration options, see the example `pipe
 
 If your pipeline has more than one step and you have multiple agents available to run them, they will automatically run at the same time. If your steps rely on running in sequence, you can separate them with [wait steps](/docs/pipelines/wait-step). This will ensure that any steps before the 'wait' are completed before steps after the 'wait' are run.
 
+<div class="Docs__troubleshooting-note">
+  <h3>Explicit dependencies in uploaded steps </h3>
+  <p>A pipeline upload step will force any uploaded steps to become explicit <a href="/docs/pipelines/dependencies">dependencies</a> of the steps that depend on that pipeline upload step. For example, if step B depends on step A, and step A uploads step C, then step B will also depend on step C.</p>
```
plaindocs, line 110: Let's reword away from 'will force'? Something like

> All uploaded steps become explicit dependencies of the step that depends on the step that uploaded them

Nah, that's awful. 

> All uploaded steps automatically become explicit dependencies of any step that depends on the step that uploaded them.

> If an upload step is a dependency of another step, all uploaded steps have the same explicit dependency. 

Urgh, @karensawrey @ddbeck any suggestions here? I'm tying myself in knots.

karensawrey: How about:
When a step with a dependency for another step uploads a step, the uploaded step(s) also becomes an explicit dependency. 
? @plaindocs

karensawrey: I like the "If a step depends on an upload step, then all steps uploaded become dependencies of the original step." variant a lot! Thanks, @ddbeck! I'm keeping the "explicit" though.

plaindocs: Thanks for jumping in folks. I like the new version, but am gonna suggest one small change. 

> If a step <a href="/docs/pipelines/dependencies">depends</a> on an upload step, then all steps uploaded by that step become dependencies of the original step. For example

karensawrey: Thanks, @plaindocs, implemented. Plz +1 the PR.

plaindocs (approved):
QUICK merge it before we get any more feedback ;-)

PR #1628, opened 2022-07-06
Emphasize difference between agent and job lifecycles

ddbeck:
This tries to do a few things, so that readers don't attempt to go against the grain with hooks (e.g., trying to use the agent-shutdown hook to clean up after individual jobs).

Mainly, I tried to emphasize:

- The minimalist nature of the `pre-bootstrap` and `agent-shutdown` hooks
- The job details exposed to job hooks
- Applications for agent and job hooks

Things I did _not_ attempt to do despite being tempted: rewriting the whole thing or challenging the distinction between agent and job lifecycles.

### DIFF
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -104,6 +104,7 @@ a plugin can define.
 The Buildkite agent executes agent lifecycle hooks.
 These hooks can only be defined in the [agent `hooks-path`](#hook-locations-agent-hooks) directory.
 Agent lifecycle hooks can be executables written in any programming language.
+On Unix-like systems (such as Linux and macOS), hooks must be files that are executable by the user the agent is running as.
 
 Use agent lifecycle hooks to prepare for or clean up after all jobs that may run.
 For example, use `pre-bootstrap` to block unwanted jobs from running or use `agent-shutdown` to tear down a service after all jobs are finished.
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -136,7 +136,7 @@ they are run as part of each job:
 
 Job lifecycle hooks are sourced for every job an agent accepts.
 Use job lifecycle hooks to prepare for jobs, override default behavior, or clean up after jobs that have finished.
-For example, use `environment` to set a job's environment variables or `pre-exit` to delete temporary files and remove containers.
+For example, use the `environment` hook to set a job's environment variables or the `pre-exit` hook to delete temporary files and remove containers.
 If your hook is related to the startup or shutdown of the agent, consider [agent lifecycle hooks](#job-lifecycle-hooks) for those tasks instead.
 
 Job lifecycle hooks are shell scripts that you can use to run commands and export environment variables to jobs.
--- a/pages/agent/v3/hooks.md.erb
+++ b/pages/agent/v3/hooks.md.erb
@@ -139,8 +139,8 @@ Use job lifecycle hooks to prepare for jobs, override default behavior, or clean
 For example, use the `environment` hook to set a job's environment variables or the `pre-exit` hook to delete temporary files and remove containers.
 If your hook is related to the startup or shutdown of the agent, consider [agent lifecycle hooks](#job-lifecycle-hooks) for those tasks instead.
 
-Job lifecycle hooks are shell scripts that you can use to run commands and export environment variables to jobs.
-These hooks have access to all the standard [Buildkite environment variables](/docs/pipelines/environment-variables).
+Job lifecycle hooks have access to all the standard [Buildkite environment variables](/docs/pipelines/environment-variables).
+
 Job lifecycle hooks are copied to `$TMPDIR` directory and *sourced* by the agent's default shell.
 This has a few implications:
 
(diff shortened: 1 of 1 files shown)


pages/agent/v3/hooks.md.erb
```diff
@@ -101,12 +101,16 @@ a plugin can define.
 
 ### Creating agent lifecycle hooks
 
-Agent lifecycle hooks can be written in the programming language of your choice
-and are executed by the Buildkite agent. See the documentation for each agent
-lifecycle hook for details on the interface between them and the Buildkite
-agent.
-
+The Buildkite agent executes agent lifecycle hooks.
```
plaindocs, line 104: How about 

> The Buildkite agent can run agent lifecycle hooks before bootstrapping each agent, and after shutting down each agent. 

Although it runs them if they are there, not every if they aren't

pages/agent/v3/hooks.md.erb
```diff
@@ -101,12 +101,16 @@ a plugin can define.
...
 These hooks can only be defined in the [agent `hooks-path`](#hook-locations-agent-hooks) directory.
+Agent lifecycle hooks can be written in any programming language, but make sure the file is executable.
```
plaindocs, line 106: Slightly mixed verbs here? And we should probably say as long as the file is executable by the user the agent is running as or something for Linux like stuff?

pages/agent/v3/hooks.md.erb
plaindocs: And it looks like my understanding might not be correct here. Or the description might be out.

plaindocs: > Executed before any job is started. Useful for [adding strict checks](https://bk-docs-pr-1628.onrender.com/docs/agent/v3/securing#strict-checks-using-a-pre-bootstrap-hook) before jobs are permitted to run. 

I **think** this means run when the agent starts, before it runs any jobs. It doesn't mean runs before each job. Right?

pages/agent/v3/hooks.md.erb
plaindocs: And would it be useful to say  "unlike job lifecycle hooks, swhich are sourced" here?

pages/agent/v3/hooks.md.erb
```diff
@@ -129,12 +133,15 @@ they are run as part of each job:
 
 ### Creating job lifecycle hooks
 
-Job lifecycle hooks run for every job an agent accepts. They are shell
-scripts you can use to run commands and export environment variables. These
-hooks have access to all the standard [Buildkite environment variables](/docs/pipelines/environment-variables).
+Job lifecycle hooks are sourced for every job an agent accepts.
+Use job lifecycle hooks to prepare for jobs, override default behavior, or clean up after jobs that have finished.
+For example, use `environment` to set a job's environment variables or `pre-exit` to delete temporary files and remove containers.
```
plaindocs, line 138:
```suggestion
For example, use the `environment` hook to set a job's environment variables or the `pre-exit` hook to delete temporary files and remove containers.
```

pages/agent/v3/hooks.md.erb
```diff
@@ -129,12 +133,15 @@ they are run as part of each job:
...
+If your hook is related to the startup or shutdown of the agent, consider [agent lifecycle hooks](#job-lifecycle-hooks) for those tasks instead.
 
-Job lifecycle hooks are copied to `$TMPDIR` directory and *sourced* by the
-agent's default shell. This has a few implications:
+Job lifecycle hooks are shell scripts that you can use to run commands and export environment variables to jobs.
```
plaindocs, line 141: Should this go before some of the other description?

plaindocs (approved):
I like the reduced scope of this change. Let's fix and merge, and discuss larger changes separately, and also about agent / job lifecycles.

pages/agent/v3/hooks.md.erb
ddbeck: Verb mismatch fixed with 6b9ee458a1e9cfa1db77956cfdd282d8e731d431. Unix-like permissions noted with 01505466a7d8125ba15464df8535ea1c4c34573b.

I'm of two minds about mentioning job lifecycle hooks' source behavior here. I think executing a script as a script is fairly typical, while sourcing isn't. If you're reading this _before_ being exposed to the job hooks, it's a bit mysterious and weird. As a reference, it's worth mentioning (e.g., if you're landing on this section after already doing some work with job hooks), though I'm struggling with how to say so without implying that the other one is, in fact, weird.

For now I've left it without mention. Thoughts?

pages/agent/v3/hooks.md.erb
ddbeck: Fixed with 6a2f16a387ab29f9f263490d7db4c564d3a3dca2.

pages/agent/v3/hooks.md.erb
ddbeck: This whole thing felt a little half baked. Revised with 5be4f71153d0b07b7437f623e9454a73da507b83.

pages/agent/v3/hooks.md.erb
plaindocs: OK, that makes sense

pages/agent/v3/hooks.md.erb
plaindocs: I'll buy that.

plaindocs (approved):
:shipit:

PR #1630, opened 2022-07-07
Document how branch could differ from commit

jradtilbrook:
The value of `BUILDKITE_BRANCH` might not contain the commit in `BUILDKITE_COMMIT` as these values can be specified when
triggering a build (ie. user input). The value of `BUILDKITE_COMMIT` is what the agent uses to checkout a specific commit.
`BUILDKITE_BRANCH` is mostly for UI and grouping.

If users want to do some checks and elevated privileges based on build branch, they should confirm the branch that contains
the `BUILDKITE_COMMIT` themselves during the agent workflow.

### DIFF
--- a/data/environment_variables.yaml
+++ b/data/environment_variables.yaml
@@ -78,7 +78,7 @@ variables:
   example: "/usr/local/bin"
 - name: BUILDKITE_BRANCH
   desc: |
-    The branch being built.
+    The branch being built. The value of `BUILDKITE_COMMIT` might not exist on the branch specified here for manual builds where these values are input from a user.
   modifiable: false
   example: "main"
 - name: BUILDKITE_BUILD_CHECKOUT_PATH


jradtilbrook:
I'm not too sure how to word this. But the recommendation came from Block to update the docs around this as it isn't always guaranteed to match

data/environment_variables.yaml
```diff
@@ -78,7 +78,7 @@ variables:
   example: "/usr/local/bin"
 - name: BUILDKITE_BRANCH
   desc: |
-    The branch being built.
+    The branch being built. The value of `BUILDKITE_COMMIT` might not exist on the branch specified here for manual builds where these values are input from a user.
```
plaindocs, line 81:
```suggestion
    The branch being built. Note that for manually triggered builds, this branch is not guaranteed to contain the commit specified by `BUILDKITE_COMMIT`.
```

plaindocs (approved):
Nice one @jradtilbrook ! I've made a wording change and will merge.

PR #1634, opened 2022-07-12
Add note about permissions for github provider

jradtilbrook:
To help users when adding new GitHub repository providers, add a note here about what permissions are required on both
Buildkite and GitHub


karensawrey (approved):
Great stuff, let's add it!

PR #1636, opened 2022-07-12
Require sentence case for H3 through H6

ddbeck:
This PR creates a new Vale style that requires sentence case headings for H3, H4, H5, and H6. H2s will be addressed in future PRs. Specifically, this PR does a few little things to get us from here to there:

1. It adds a new style, `Buildkite.h2-h6_sentence_case`, to require sentence case (temporarily excluding H2).
2. It fixes the failures caused by the new style.
3. It adds another new style, `Buildkite.existence-case-sensitive`, to catch inconsistencies created by the fixes above. I did this to avoid creating new inconsistencies between body text and the headings.

Enforcement on H2 causes many, many errors, some of which require discussion to resolve. I'll follow up on that, if this gets merged.


karensawrey (approved):
Great! Let's ship it!

PR #1646, opened 2022-07-18
Default priority is shipped

plaindocs:
Nice one @jradtilbrook and @sj26  !

### DIFF
--- a/pages/pipelines/managing_priorities.md.erb
+++ b/pages/pipelines/managing_priorities.md.erb
@@ -17,11 +17,7 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
-Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority are assigned before jobs with lower priority, regardless of which has been longest in the queue.
-
-<!-- date -->
-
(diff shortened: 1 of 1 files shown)


jradtilbrook (approved):

pages/pipelines/managing_priorities.md.erb
```diff
@@ -17,11 +17,7 @@ steps:
 ```
 
-Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority are assigned before jobs with lower priority, regardless of which has been longest in the queue.
-
-<!-- date -->
-
-Prioritization can only be set on jobs, and only applies to command jobs, including plugin commands. Setting a default command job priority across a whole build is coming at the start of 2022.
+Job priority is considered before jobs are dispatched to [agent queues](/docs/agent/v3/queues), so jobs with higher priority are assigned before jobs with lower priority, regardless of which has been longest in the queue. Priority one applies to command jobs, including plugin commands.
```
sj26, line 20: > Priority **one** applies to command jobs, including plugin commands.

Is this a typo? Was it meant to be "only"?

plaindocs: Correct! This isn't airforce! https://github.com/buildkite/docs/pull/1653#pullrequestreview-1045242680

PR #1649, opened 2022-07-19
add note about M1 chips and homebrew path changes

jeremybumsted:
This could definitely use some better wording, but the main idea here is to make sure we note about the path changes present in M1 installations of Homebrew.

### DIFF
--- a/pages/agent/v3/macos.md.erb
+++ b/pages/agent/v3/macos.md.erb
@@ -37,7 +37,7 @@ For the various installation methods.
 
 ### Homebrew installation on macOS on M1 chips
 
-* Configuration: `/opt/homebrew//buildkite-agent/buildkite-agent.cfg`
+* Configuration: `/opt/homebrew/etc/buildkite-agent/buildkite-agent.cfg`
 * Agent Hooks: `/opt/homebrew//buildkite-agent/hooks`
 * Builds: `/opt/homebrew/buildkite-agent/builds`
 * Log: `/usr/local/var/log/buildkite-agent.log`
--- a/pages/agent/v3/macos.md.erb
+++ b/pages/agent/v3/macos.md.erb
@@ -33,16 +33,20 @@ See the [Agent SSH Keys](/docs/agent/v3/ssh-keys) documentation for more details
 
 ## File locations
 
-For the various installation methods.
+File locations depend on your installation method and Mac hardware.
 
-### Homebrew installation on macOS on M1 chips
+### Homebrew installation
+
+To see the paths to the agent's configuration, hooks, builds, and logs on your system, run `brew info buildkite-agent`.
+
+The typical paths for [Mac computers with Apple silicon](https://support.apple.com/en-gb/HT211814) (such as M1 chips) are:
 
 * Configuration: `/opt/homebrew/etc/buildkite-agent/buildkite-agent.cfg`
 * Agent Hooks: `/opt/homebrew//buildkite-agent/hooks`
 * Builds: `/opt/homebrew/buildkite-agent/builds`
 * Log: `/usr/local/var/log/buildkite-agent.log`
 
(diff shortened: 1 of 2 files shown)


plaindocs (changes requested):
Ello mate, thanks for the update. 

In this section we're telling folks where things are installed, not telling them where to install stuff. If you drop me a comment with the default location on M1, basically the output of ``$(brew --prefix)` I'll split the section into two, for M1 and not M1.

Also, you probs need to `git submodule update` to get rid of the emojis in the commit (and or revert those changes)

karensawrey:
@plaindocs for reverting emoji commits, we're now recommending use `git reset --soft HEAD~1` to undo the last commit, un-stage the erroneous submodule change, and commit again.

jeremybumsted:
@plaindocs 👋  It should be `/opt/homebrew/` as the default on M1 macs - if it's easier we can just close this PR.

plaindocs:
@jeremybumsted How about this? I assume I'm right in leaving the log file untouched?

ddbeck:
@plaindocs I hope it's not too far along for me to make a couple suggestions here, one big and one narrow:

## The bigger, optional suggestion

Instead of naming the specific paths, how about using Homebrew's own machinery to make this section generalized for _any_ Homebrew installation, even ones to non-standard prefixes? We could tell Homebrew users something like:

> To see the paths to the agent's configuration, hooks, builds, and logs, run `brew info buildkite-agent`.

Or, if we wanted to break it down, we could use `$(brew --prefix)` itself:

> * Configuration: `$(brew --prefix)/etc/buildkite-agent/buildkite-agent.cfg`
> * Agent Hooks: `$(brew --prefix)/etc/buildkite-agent/hooks`
> * Builds: `$(brew --prefix)/var/buildkite-agent/builds`
> * Log: `$(brew --prefix)/var/log/buildkite-agent.log`

(This is less desirable, only because it won't work in all shells. I'm a fish user and this would have annoyed me, historically.)

## Narrower, less optional suggestion

If the above suggestion is unwelcome, then we ought to use [Apple's terminology](https://support.apple.com/en-gb/HT211814) to distinguish between Intel and ARM-based Macs. The headings would then be something like:

> ### Homebrew installation on Mac computers with Apple silicon
> ### Homebrew installation on Mac computers with Intel processors

Or to condense it a bit:

> ### Homebrew installation on Apple silicon-based Macs
> ### Homebrew installation on Intel-based Macs

This helps the headings stay fresher (e.g., you won't have to add M2 to the heading soon) and more closely follows the UI you would use to find out what hardware you're actually running (I'm skeptical that this hardware distinction is already known to most developers).

jeremybumsted (commented):
Some minor adjustments to the paths, and corrections to the root of the path 😄

plaindocs:
@ddbeck do you want to implement your second suggestion, and as much of the first as you see fit? 
We're definitely hitting my inexperience in macOS here/

ddbeck:
@plaindocs On further consideration, I decided we could do an all-of-the-above approach: provide the command (which will be authoritative), the typical paths, and use the Apple-ish language. Let me know what you think!

plaindocs (approved):
Much better. 

If the paths are authoritative shall we delete `typical`? Or if you're reluctant, how about replacing with `default`?

ddbeck:
@plaindocs I chose "typical" because we're really representing a double default which I'm trying to avoid having to explain in-text: if you have the default Homebrew installation path (i.e., your `brew --prefix` is the default) _and_ you're not overriding any of the agent paths (I'm not sure all of them are configurable, either, so it's possible some will be correct and others won't), only then will the listed paths will be correct.

plaindocs:
Yeah, it's one of those where if folks have changed the defaults you hope they know enough to know they changed the defaults...

:shipit:

PR #1645, opened 2022-07-15
Pipelines Overview screenshots rehaul

karensawrey:
Updating all the screenshots in https://buildkite.com/docs/pipelines.

PR #1654, opened 2022-07-21
🍱 Revise pipelines screenshots

olyism:
This is a follow up PR to #1645 tweak [Pipelines page](https://buildkite.com/docs/pipelines)'s screenshots. Changes as follows:

- [x] 🍏 **Recreate screenshots from macOS Safari** so they have consistent look and typography with all the other screenshots
- [x] 🐛 Fixed screenshots showing experimental/featured-flagged UI
- [x] 📐 Consistent image dimensions
- [x] 📱 Use @2x retina image sizing (not the perfect `srcset` implementation yet, that's for a future scope)
- [x] 🔍 Change focal point styling
- [x] 🍱 Add a new screenshot for pipeline filtering
- [x] 🔥 Remove unused image
- [x] 📄 Updated screenshot styleguide documentation

## Screenshot

Screen Shot 2022-07-21 at 11 44 29 pm


plaindocs (approved):
Speedy work @olyism !

plaindocs:
One non-blocking addition to this for the future is that I'd love to create a Demo organization that we always use to take screenshots, partly as an added layer of consistency, and party as a guarantee that we're not leaking feature flagged screenshots.

We can use the same organization when we get around to automating REST requests as well. Maybe a daily cron job on a public repo, such as the Docs itself would be sufficient. Suggestions for the name of the organization on a postcard please. Must include an emoji.

PR #1658, opened 2022-07-22
Revise secrets management using hooks

ddbeck:
We received some feedback on this section of the docs, reporting that the location of hooks is unclear and that it was a barrier to setting up Test Analytics. I had a similar experience setting up Test Analytics for the first time, so this section is ripe for revision.

What I tried to do:

- Use Test Analytics as a less complex example for exporting a token to every step of a pipeline.
- Generalize the existing example to illustrate conditionally exporting secrets to a named step (as written, it was rather specific coverage of creating a GitHub deployment using secrets in a hook, with the general point—using conditions to minimize exposure—buried in a note).
- Favor concrete shell "export" terminology over the abstract "expose".


pages/pipelines/secrets.md.erb
```diff
@@ -11,57 +11,56 @@ A best practice for secret storage is to use your own secrets storage service, s
...
-  - command: scripts/trigger-github-deploy
-    key: trigger-github-deploy
-```
-{: codeblock-file="pipeline.yml"}
+By default, the `environment` hook file is stored in the agent's `hooks` directory.
+The path to this directory varies by platform; read the [installation instructions](/docs/agent/v3/installation) for the path on your platform.
+The path can also be overridden by the [`hooks-path`](/docs/agent/v3/hooks#hook-locations-agent-hooks) setting.
 
-This step runs the `scripts/trigger-github-deploy` script from the repository, and creates a GitHub Deployment using a [GitHub personal access token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line). The script has the following source code:
+For example, suppose you want to expose a Test Analytics API token to a specific pipeline.
```
plaindocs, line 25: I'd write around suppose, and in this context we can probably delete entirely? 

```suggestion
For example, to expose a Test Analytics API token to a specific pipeline.
```

pages/pipelines/secrets.md.erb
```diff
@@ -11,57 +11,56 @@ A best practice for secret storage is to use your own secrets storage service, s
 There are also various [Buildkite Plugins](/docs/plugins) that integrate reading
 and exposing secrets to your build steps using secrets storage services.
 
-## Storing secrets in environment hooks
+## Exporting secrets with environment hooks
```
plaindocs, line 14: Exporting is a definitely better than Storing here, but I wonder if there is a third option? 

> Injecting secrets from an environment hook?

> Securely accessing secrets from ...

pages/pipelines/secrets.md.erb
```diff
@@ -11,57 +11,56 @@ A best practice for secret storage is to use your own secrets storage service, s
...
+```
+{: codeblock-file="pipeline.yml"}
+
+In your `environment` hook, you can export the deployment token only when when the job is the deployment step in a specific pipeline:
 
 ```bash
 #!/bin/bash
 set -euo pipefail
 
 if [[ "$BUILDKITE_PIPELINE_SLUG" == "my-app" ]]; then
```
plaindocs, line 54: Would it make more sense to chain conditions using && instead of nesting using if? It fits my brain better, but I don't feel strongly about this.

pages/pipelines/secrets.md.erb
```diff
@@ -11,57 +11,56 @@ A best practice for secret storage is to use your own secrets storage service, s
...
+    export SECRET_DEPLOYMENT_ACCESS_TOKEN="bd0fa963610b..."
   fi
 fi
 ```
 {: codeblock-file="hooks/environment"}
 
-<section class="Docs__note">
-  <p>Adding conditional checks, such as the pipeline slug and the step identifier, helps to limit accidental use and disclosure of secrets.</p>
-</section>
+The script exports `SECRET_DEPLOYMENT_ACCESS_TOKEN` only when the specific pipeline and step is reached.
```
plaindocs, line 62: 'reached' is a little odd here I think. It exports it to that job only, right?

and there is a mismatch between `trigger-deploy` and `trigger-github-deply` if we're matching the previous example I think

plaindocs (approved):
I've got a few nits and suggestions, but this is definitely an improvement.

ddbeck:
Hi @MelissaKaulfuss, this PR was prompted by your feedback. You're welcome to review (or not). Either way, thank you!

pages/pipelines/secrets.md.erb
ddbeck: I tried to avoid a longer sentence and I guess it came out weird? Revised with b2e7ab36.

pages/pipelines/secrets.md.erb
ddbeck: The original was nested and I thought it might've been a house style thing. Switched to logical AND with 016a43cb.

pages/pipelines/secrets.md.erb
ddbeck: Dropped "reached" with 02c2c453. And fixed the mismatch with eac0edea.

pages/pipelines/secrets.md.erb
ddbeck: I am not strongly in favor of any one of them. I just really disliked "storing" for this particular case. If you've got a preference though, I'm happy to follow that.

plaindocs: Nothing strong. I think the direction of hook -> step implied with Exporting is better than variable -> hook, ie it's closer to what folks are trying to do. :man_shrugging:

karensawrey (approved):

PR #1659, opened 2022-07-22
Addressing the feedback

karensawrey:
According to the docs' user feedback, the docs were unclear on where the Buildkite agent is actually run. Adding info and a link to the page in question + fixing a faulty capital letter on the agent installation page.

### DIFF
--- a/pages/agent/v3/ssh_keys.md.erb
+++ b/pages/agent/v3/ssh_keys.md.erb
@@ -6,7 +6,7 @@ If your agent needs to clone your repositories using git and SSH, you'll need to
 
 ## Finding your SSH key directory
 
-The Buildkite agent runs [on your own machine](/docs/agent/v3/installation), whether it's a VPS, server, desktop computer, embedded device. When the Buildkite agent runs any git operations, SSH will look for keys in `~/.ssh` under whichever user the agent runs as. Each platform's [agent installation documentation](/docs/agent/v3/installation) specifies the exact location of this directory.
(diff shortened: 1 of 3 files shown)


pages/agent/v3/ssh_keys.md.erb
```diff
@@ -6,7 +6,7 @@ If your agent needs to clone your repositories using git and SSH, you'll need to
 
 ## Finding your SSH key directory
 
-When the Buildkite Agent runs any git operations, SSH will look for keys in `~/.ssh` under whichever user the agent runs as. Each platform's [agent installation documentation](/docs/agent/v3/installation) specifies the exact location of this directory.
+The Buildkite agent runs [on your own machine](/docs/agent/v3/installation), whether it's a VPS, server, desktop computer, embedded device. When the Buildkite agent runs any git operations, SSH will look for keys in `~/.ssh` under whichever user the agent runs as. Each platform's [agent installation documentation](/docs/agent/v3/installation) specifies the exact location of this directory.
```
plaindocs, line 9: I think this is actually less helpful here, even though it's a direct copy of the content in `installation.md.erb`. How about something more like 

> When the Buildkite agent runs any git operations, it looks for SSH keys in `~/.ssh` under the user the agent runs as. Each platform's [agent installation documentation](/docs/agent/v3/installation) specifies which user the agent runs as and which directory the SSH keys are in. For example, on Debian the agent runs as `buildkite-agent` and the SSH keys are in `/var/lib/buildkite-agent/.ssh/` but on macOS the agent runs as the user who started the `lauchd` service, and the SSH keys are in that users `.ssh` directory

Do we think the example is helpful here?

plaindocs (changes requested):

plaindocs:
Bonus related typo if you want to fix it in this PR, in docs/agent/v3/macos we have "On macOS, the Buildkite agent runs as the user who started the launched service." which should probably be "On macOS, the Buildkite agent runs as the user who started the `launchd` service."

pages/agent/v3/ssh_keys.md.erb
karensawrey: Implemented. Not sure about adding an extra example.

plaindocs (approved):

PR #1660, opened 2022-07-25
Update Algolia config

lachlanhardy:
This resolves DOC-426:
https://linear.app/buildkite/issue/DOC-426/update-algolia-config-to-reflect-new-html-structure

This updates the config settings for Algolia and also makes sure that the search now works in dev and in the Render environments for PRs.

Testing to be done between Render and the live site to confirm this works as expected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,3 +67,5 @@ group :test do
   # Browser testing stuff
   gem "capybara"
 end
+
+gem "dotenv", "~> 2.7"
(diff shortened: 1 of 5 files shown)


plaindocs:
@lachlanhardy we could probably add an RSpec test that checks for all of the Algolia selectors, right?

lachlanhardy:
@plaindocs Cool idea!

lachlanhardy:
@plaindocs I'll add that test. Meanwhile, I can't find any situations where the results don't line up the same across the two. I can't decide if that's great or not :) At least it means I've not broken it!

Still I'd appreciate testing from the rest of the team, as you'll all know the content better.

lachlanhardy:
@plaindocs and I have been struggling to prove this is working - or NOT working - so we're ditching it in favour of the upgrade to Algolia Crawler.

PR #1671, opened 2022-07-29
Be more helpful when talking about the agent

plaindocs:
I'm trying to address this feedback a little

> This is a whole lot of "how it works that is written under the assumption that I know what to do with it. I'm just confused."

### DIFF
--- a/pages/agent/v3.md.erb
+++ b/pages/agent/v3.md.erb
@@ -4,6 +4,11 @@
 
 The Buildkite agent is a small, reliable and cross-platform build runner that makes it easy to run automated builds on your own infrastructure. Its main responsibilities are polling buildkite.com for work, running build jobs, reporting back the status code and output log of the job, and uploading the job's artifacts.
 
+This page contains reference information for Buildkite organization administrators.
+It covers agent installation and configuration details and how agents communicate with Buildkite.
+If you're working with a team that already uses Buildkite and you want to write code that agents will run, read [Pipelines](/docs/pipelines).
+If you're setting up a Buildkite organization and you don't already have agents running, read [Getting started](/docs/tutorials/getting-started).
+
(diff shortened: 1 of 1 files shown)


pages/agent/v3.md.erb
```diff
@@ -4,6 +4,8 @@
 
 The Buildkite agent is a small, reliable and cross-platform build runner that makes it easy to run automated builds on your own infrastructure. Its main responsibilities are polling buildkite.com for work, running build jobs, reporting back the status code and output log of the job, and uploading the job's artifacts.
 
+You (or your organization) need one or more running agents to run builds, but once you've installed the agent and got it running on your own infrastructure, you don't need to interact with it directly. Whether you're starting builds automatically with every commit, or runnning them manually by clicking a button, Buildkite handles everything from telling the agent what vesion control references to use, where to get the changes from, and what code to run; as well as reporting the outcome back to Buildkite.com.
```
ddbeck, line 7: This is good information (and worth adding) but I think what might be missing is something that helps the reader understand three things:

- Whether they're the intended audience for this content (i.e., are you a Buildkite infra maintainer?)
- What they're actually going to find in this page (i.e., does this page contain a topic you actually care to learn about?)
- Where to look elsewhere, if this is not what they're looking for (because it probably isn't)

So maybe you could preface this new text with something along the lines of:

```suggestion
This page contains reference information for Buildkite organization administrators.
It covers agent installation and configuration details and how agents communicate with Buildkite.
If you're working with a team that already uses Buildkite and you want to write code that agents will run, read [Pipelines](/docs/pipelines).
If you're setting up a Buildkite organization and you don't already have agents running, read [Getting started](/docs/tutorials/getting-started).

You (or your organization) need one or more running agents to run builds, but once you've installed the agent and got it running on your own infrastructure, you don't need to interact with it directly. Whether you're starting builds automatically with every commit, or runnning them manually by clicking a button, Buildkite handles everything from telling the agent what vesion control references to use, where to get the changes from, and what code to run; as well as reporting the outcome back to Buildkite.com.
```

(This page also has a structural problem of it being a sort of grab bag of random agent info. For instance, the experimental features could be its own page, or part of the configuration page. The stuff about redaction could be part of the "Securing Your Buildkite Agent" page. And so on. But I figure that's a little out of scope.)

plaindocs: Agreed, love that addition on lines 9 and 10.

> Buildkite organization administrators

this is something to dig into another day.

ddbeck (approved):
👍

PR #1673, opened 2022-07-29
🚸 Use comparison table to visualize API parity

olyism:
Proposing some usability and styling improvements to make the API parity page. I used a common design pattern similar to comparison tables.

Also uses a yaml file to manage the comparison data.

### Before

Listing REST and GraphQL differences in two separate sections using list.

Screen Shot 2022-07-30 at 12 12 48 am

### After

Using a comparison table.

Screen Shot 2022-07-30 at 12 13 53 am

A sticky table header persists so the REST and GraphQL table headers are always visible for reference.

Screen Shot 2022-07-30 at 12 15 20 am

## Tested in popular browsers

Manually tested on my macOS.

### Chrome

Screen Shot 2022-07-30 at 12 16 35 am

### Safari

Screen Shot 2022-07-30 at 12 17 07 am

Screen Shot 2022-07-30 at 12 17 27 am

### Firefox

Screen Shot 2022-07-30 at 12 18 18 am

Screen Shot 2022-07-30 at 12 18 50 am

### DIFF
--- a/app/assets/stylesheets/public/_comparison-table.scss
+++ b/app/assets/stylesheets/public/_comparison-table.scss
@@ -1,0 +1,61 @@
+.comparison-table {
+  th,
+  td {
+    background-color: white;
+    padding: 15px;
+
+    &:first-child {
+      padding-left: 0;
+    }
+
+    &:nth-child(even) {
+      background-color: #f8f8f8;
+    }
+  }
+
+  thead {
+    position: sticky;
+    top: 0;
+
+    @media (min-width: $screen-md) {
+      top: $header-height;
+    }
+
+    th {
+      @include blur-background;
+      vertical-align: bottom;
+
+      &:nth-child(even) {
+        @include blur-background(#f8f8f8);
+      }
+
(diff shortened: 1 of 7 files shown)


karensawrey:
Oh wow, I LOVE it.

plaindocs:
So this looks great, but it's suuuper misleading, because the list of differences isn't a complete or nearly complete list of features, it's restricted to the places we know there are differences. So this makes it look like the feature set is almost complimentary, when actually it's more like 90% overlap.

karensawrey:
@plaindocs I remember talking about organizing a full comparative table - so, I guess, we've got it now. Maybe this can be put on ice until we've dragged the GraphQL docs into the docs? Visually, I very much like what I'm seeing here. I also agree with your comment on this being misleading.

olyism:
Note to self: for simplicity and maintainability we could list high-level shared API capabilities in the table, and add more details for feature differences.

In the mean time, I'm closing this PR. Will keep an eye out for feedback on the current feature parity page to validate what improvements are needed.

PR #1675, opened 2022-08-01
Adding info on granular access permissions in REST

karensawrey:
As described in the topic.


plaindocs (changes requested):
I think this doesn't highlight that difference enough. I'd like to see it as it's own bullet point in features only REST API has.

pages/apis/api_differences.md.erb
```diff
@@ -10,6 +10,7 @@ On this page, we've collected the known limitation where some API features are o
 
 ## Features only available in the REST API
 
+* <%= pill "ACCESS TOKEN", "access-token" %> [limit the token scope using granular access permissions](/docs/apis/managing-api-tokens#token-scopes).
```
plaindocs, line 13: How about this? 

```suggestion
* <%= pill "ACCESS TOKEN", "access-token" %> [granular access permissions](/docs/apis/managing-api-tokens#token-scopes).
```

plaindocs (approved):

PR #1677, opened 2022-08-02
Keep render previews for 2 days after the last commit instead of 5

plaindocs:
To bring the cost of inactive open PRs down. If it's annoying we can revert.


karensawrey:
What about the commits deployed on a Friday? Will they expire by Monday?

karensawrey (approved):

PR #1678, opened 2022-08-02
Revert "Merge pull request #1641 from buildkite/doc-438-graphql-data-…

plaindocs:
Reverting due to Nav oddness

image 

image

This reverts commit 047614f00ca707ec188e89f50ee025f1bfd798e4, reversing
changes made to 0805db772dbf8348dbaf2af1e3a343b387b12e69.

PR #1641, opened 2022-07-14
✨ GraphQL reference docs

olyism:
New feature that generates markdown pages from a GraphQL API schema JSON export.

Here's a 5 min Loom video of me talking about the high-level UX and changes https://www.loom.com/share/d07d7374b4d2451e83c08750ded9fded

## Overview of changes

- [x] Set up script to generate all markdown pages
- [x] Render schema information onto pages
- [x] Navigation: dedicated nav for API, Integrations and GraphQL API docs
- [x] Styling 💄 💅 ✨
- [x] Update notification message
- [x] Linting fixes
- [x] Algolia to not index the schema pages
- [x] ~Better surface mutation~ Not needed as this is in query
- [x] Sort alphabetically
- [x] Add instructions

## Screenshots

GraphQL API's landing page with left nav divided by schema types.

Screen Shot 2022-07-18 at 9 01 20 pm

A schema page generated by the script.

Screen Shot 2022-07-18 at 9 03 15 pm

PR #1613, opened 2022-06-29
 API disparity between GraphQL and REST

karensawrey:

PR #1679, opened 2022-08-03
🐛 Fix API Differences link in the new nav structure

olyism:
We've reverted the PR https://github.com/buildkite/docs/pull/1641, which lifted APIs and Integrations in the nav, because the [API Differences page](https://buildkite.com/docs/apis/api-differences) is missing.

I've amended this by adding the API Differences page in the nav under APIs > REST.

Screen Shot 2022-08-03 at 11 30 59 am

How I implemented the fix in this PR (you can also [watch the Loom 🎦 ](https://www.loom.com/share/3192534d18f9497c91ad3095e7a55365))

1. **"Revert the revert"** https://github.com/buildkite/docs/pull/1678
2. **Add API Differences in the Nav** under APIs > REST in `data/nav.yml`
3. **Add nav icon** as an improvement to minimize confusion by helping customers know that they're leaving the Pipelines section

## Shedding light on recent UX confusion 🤔❓

- Oversight of not adding API Differences under the new top level APIs nav
- Duplication of API links
- Resolved by having dupe links
- Dupe links help 

The UX confusion arises from API Differences being added under Pipelines > APIs > API Difference in https://github.com/buildkite/docs/pull/1613 but not added under the newly introduced top level APIs nav.

Screen Shot 2022-08-03 at 11 39 25 am

[PR 1641 that introduced top level APIs and Integrations to the nav](https://github.com/buildkite/docs/pull/1641) was merged after API Differences page, which by oversight didn't add API Differences under the new nav.

Screen Shot 2022-08-03 at 11 44 00 am

I've decided to dupe the links to help existing customers with finding the API information.

As a UX improvement, I've added an icon and title tag as an affordance to let customers know that they're leaving the Pipeline section.

Screen Shot 2022-08-03 at 1 29 34 pm


olyism:
Here's a Loom walking through the UX changes

https://www.loom.com/share/3192534d18f9497c91ad3095e7a55365

olyism:
Pulling back to draft as Render preview is returning a 500 error. 😬

olyism:
Ready for review again. 500 error caused by `image_url` in view partial which, strangely, works on local dev but not Render preview

plaindocs:
@olyism love the Loom, love the new link affordances.

I'm not sure why the Nav differences didn't show up in the original PR merge though.

From an editing point of view, are both the API nav in Pipelines and the API nav in the top bar, in `nav.yaml`? When I was looking yesterday they didn't seem to be. 

My ideal I think is that we only have to edit the Nav in one place, and it should match. (with optional flattening as you've done). As soon as you have to remember to edit the same Nav in two places, human error gets involved real quick.

olyism:
thanks @plaindocs 

> I'm not sure why the Nav differences didn't show up in the original PR merge though.

Yeah so this was my oversight. I forgot to add API Difference under APIs in `nav.yml`

```yaml
  - name: 'Pipelines'
    path: 'tutorials/getting-started'
    children:
      ……
      - name: 'APIs'
        children:
          ……
          # Karen's PR added the nav link here 👇 
          - name: 'API Differences'
            path: 'apis/api-differences'
          ……
  ……
  - name: 'APIs'
    path: 'apis'
    children:
      - name: 'Overview'
        path: 'apis'
      - name: 'Managing API Tokens'
        path: 'apis/managing-api-tokens'
      # But I forgot to add API Differences nav here before I merged 👇 
      - name: 'API Differences'
        path: 'apis/api-differences'
      ……
```

> From an editing point of view, are both the API nav in Pipelines and the API nav in the top bar, in nav.yaml? When I was looking yesterday they didn't seem to be.

Yes, all navs are managed within `nav.yml`. Pipelines start at [line 2](https://github.com/buildkite/docs/blob/doc-477-graphql-and-api-diff-fix/data/nav.yml#L2), and APIs start at line [362](https://github.com/buildkite/docs/blob/doc-477-graphql-and-api-diff-fix/data/nav.yml#L362).

> My ideal I think is that we only have to edit the Nav in one place, and it should match. (with optional flattening as you've done). As soon as you have to remember to edit the same Nav in two places, human error gets involved real quick.

Yes, I think it's best to only edit one place as well. I thought about removing API completely from under Pipelines but just wary about the change being too drastic. I would like to gradually remove the shortcut links under Pipelines in a month's time.

**OR**, we could have no children API and Integration links under Pipelines, so we won't have to double-handle these links.
Screen Shot 2022-08-03 at 9 09 30 pm

Let me know if the approach in the screenshot is better and I can push this commit up.

plaindocs (approved):
Thanks for jumping on this @olyism 

I think I do prefer the screenshot version, what about you?

olyism:
No worries :) yes, I prefer the simplified version too. Have pushed the commit.

karensawrey (approved):

PR #1680, opened 2022-08-03
Fix link in REST API - Builds page

plaindocs:
Tiny link fix.


kalopilato (approved):
Lovely! Seems unnecessary but I gave it a quick double check on the PR preview anyhow and confirmed the anchor's now working as expected. ✅  :shipit:

PR #1691, opened 2022-08-09
🚸 Integrations overview page

olyism:
Adding an overview/landing page for the Integrations section.

Currently when you visit `buildkite.com/docs/integrations` you're redirected to the GitHub page, which is a UX bug.

In light of this, I've made the following changes

- Added an Integrations index page/landing page with tiles to link to child pages
- Remove redirect rule
- Styling improvements for Tiles

Screen Shot 2022-08-16 at 9 55 15 pm

## Nav changes to consider for later iterations

Whilst this PR specifically addresses the re-direct UX bug, we also notice that other integrations, such as Slack, is uncategorized. I'm catching up with other teams to get insight on which integrations are most useful to the customer and will raise more design discussions for future improvement PRs.


data/tiles.yml
```diff
@@ -41,3 +41,75 @@ test_analytics_guides:
         url: "/docs/test-analytics/your-own-collectors"
       - text: "CI environment variables"
         url: "/docs/test-analytics/ci-environments"
+  - ""
+integrations_row1:
+  - title: "Source Control"
+    desc: "Buildkite integrates with several popular code repository management systems. These integrations allow you to trigger builds based on version control events."
+    url: "/docs/integrations"
+    links:
+      - text: "GitHub"
```
plaindocs, line 50: Might be nice to add a oneliner description to these, what do you think? (doesn't have to block this PR)

plaindocs (commented):
I love it. I still have the caveat that these are Pipelines integrations though, and if we're moving things around it might make sense to make that right as well.

Problem there is that that we don't yet have equivalent Analytics content.

plaindocs:
I think the page redesign could still stand without the Horizontal nav change.

config/routes.rb
```diff
@@ -72,7 +72,6 @@
 
   # Doc sections that don't have overview/index pages, so need redirecting
   get "/docs/tutorials",    to: redirect("/docs/tutorials/getting-started"), status: 302
```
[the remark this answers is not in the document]
olyism: Definitely, have added a ticket in the backlog 👍

data/tiles.yml
olyism: Yeah sounds good, happy to raise a separate PR to add the oneliners. This PR is very bare, mainly trying to resolve the weird redirect behavior 🐛 💥 

What do you have in mind for the oneliner? Something like the expressing the simplicity of integrating with GitHub.

As GitHub is most likely the top use case, I think it might be beneficial to have an entire row dedicated to Source Control and the VCS' as tiles?

olyism:
> I think the page redesign could still stand without the Horizontal nav change.

Cool, fixed up this commit by reverting the horizontal nav change and force pushed.

> Problem there is that that we don't yet have equivalent Analytics content.

What Analytics content could we use right now? Should Collectors be under Integrations, or something else?

I've manually tested in Chrome, Safari and Firefox. Ready for review.

olyism:
As a side note: the tiles look very rudimentary. I plan to address the tiles look-and-feel in a later ticket (I think I'll tackle it with the Tiles WX improvement).

plaindocs:
I think the question isn't "What can we mash into that shape for TA" but more "How do we find a design that works for unevenly shaped content"

olyism:
> I think the question isn't "What can we mash into that shape for TA" but more "How do we find a design that works for unevenly shaped content"

Ah right, I see. I've unresolved the previous comment. This is a tricky one 🤔 . Let's have a chat tomorrow on the best possible design for this.

olyism:
thanks @plaindocs for the chat, so I've made the changes, force pushed and ready for review again. I've summarized this PR's changes in the description at the top, by TL;DR:

- Added Integrations > Overviews landing page
- Added 3 tiles to the landing page: SSO, Plugins and Other Integrations
- Removed redirect rule from `integrations` to `integrations/sso`
- Source Control stayed under Pipelines, untouched

Tested in Safari, Chrome and Firefox on Render.

app/models/landing_pages.rb
```diff
@@ -1,5 +1,8 @@
 class LandingPages
   def self.all
-    [ 'test-analytics' ]
+    [
+      "test-analytics",
```
plaindocs, line 4: These are keys in `tiles.yaml`?

plaindocs (approved):
If we integrate some or all of the changes in https://github.com/buildkite/docs/pull/1707 I'm happy with this.

app/models/landing_pages.rb
olyism: Ah, these are identifier for pages, so Test Analytics and Integrations are Landing Pages as specified in the array.

By doing so it'll apply landing page behavior, such as removing the right column place to have a wider page layout to fit in the tiles.

As a side note, probably a bit confusing and not the best WX. I'll take a look at some WX improvements this Thursday (such as the ticket to supplant this behavior with `{:notoc}` or to not have YAML data everywhere).

PR #1693, opened 2022-08-11
🔥 No need to specify image sizes

olyism:
To improve the technical writing experience (WX 😋) for adding screenshots, I propose to simply not include the `width` and `height` arguments in the image render methods.

In this PR, I've made the following changes:

1. Removed screenshot instructions relating to size attributes in `STYLE.md`
2. Removed code relating to `.responsive-image-container`

I don't see the a lot of UX value in `.responsive-image-container`. After looking at the stylesheets and the method in `page.rb`, all it's doing is apply a `max-width` to contain the image.

All `img` tags already have the standard responsive styling applied (i.e. `max-width: 100%;`) so it won't break mobile view.

For upcoming work I'd like to do the following:

- Fixed page container, so images (as well as other page elements) will fit nicely into the page for better legibility
- Proper retina image handling: so we can upload a @2x size image and then use something like `srcset` to render the image file for the right image density
 
We also don't need to remove width and height for existing images.

## Screenshots 

Screen Shot 2022-08-11 at 3 18 26 pm

Screen Shot 2022-08-11 at 3 19 05 pm

### DIFF
--- a/styleguide/STYLE.md
+++ b/styleguide/STYLE.md
@@ -455,8 +455,8 @@ This information was aggregated by going over the existing screenshots in the do
 
 ### Taking and processing screenshots
 * **Format:** PNG
-* **Ratio:** arbitrary, but **strictly even number of pixels** for both height and width. Recommended size `width: 1024px, height: 880px` when you're taking a full-width screen
-* **Size:** the largest possible resolution that makes sense. It's preferable that you take the screenshots on a Mac laptop with a Retina screen using Safari. Images should be exported at double (`@2x`) the original screen. Recommended dimension is `width: 2048/2, height: 880/2` to get the best possible view across different screen sizes.
+* **Ratio:** arbitrary, Recommended size `width: 1024px, height: 880px` when you're taking a full-width screen
+* **Size:** the largest possible resolution that makes sense
 * **No feature flag:** please remember to turn off all experimental features when taking screenshots
 * **Border:** no border
 * **Drop shadow:** no
@@ -468,14 +468,11 @@ This information was aggregated by going over the existing screenshots in the do
 * **Naming screenshots:** lowercase, words separated by hyphens; number after the title, for example, "installation-1"
 
 ### Adding screenshots or other images
-> Before you proceed, make sure that both the width and the height of the image are an even number of pixels!
-
 Steps for adding add an image to a documentation page:
 1. Name the image file (lowercase, separate words using hyphens; add a number to the filename, for example, 'installation-1' if you are adding several images to the same page)
 2. Put the file into the corresponding `images` folder (a folder with the same name as the page you are adding this image to; create such folder if it doesn't exist yet)
 3. Compose relevant alt text for the image file using Title case
-4. Add your image file to the documentation page using the following code example `<%= image "your-image.png", width: 1110, height: 1110, alt: "Screenshot of Important Feature" %>`.
-For large images/screenshots taken on a retina screen, use `<%= image "your-image.png", width: 1110/2, height: 1110/2, alt: "Screenshot of Important Feature" %>`.
+4. Add your image file to the documentation page using the following code example `<%= image "your-image.png", alt: "Screenshot of Important Feature" %>`
 
 ## GraphQL API schemas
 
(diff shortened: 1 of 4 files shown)


karensawrey:
Will the rule that the images need to have an even number of pixels still hold after this update?

styleguide/STYLE.md
```diff
@@ -455,8 +455,8 @@ This information was aggregated by going over the existing screenshots in the do
 
 ### Taking and processing screenshots
 * **Format:** PNG
-* **Ratio:** arbitrary, but **strictly even number of pixels** for both height and width. Recommended size `width: 1024px, height: 880px` when you're taking a full-width screen
-* **Size:** the largest possible resolution that makes sense. It's preferable that you take the screenshots on a Mac laptop with a Retina screen using Safari. Images should be exported at double (`@2x`) the original screen. Recommended dimension is `width: 2048/2, height: 880/2` to get the best possible view across different screen sizes.
+* **Ratio:** arbitrary, Recommended size `width: 1024px, height: 880px` when you're taking a full-width screen
```
karensawrey, line 458: What about the retina screens? 1024 x 880 seems tiny for a full-width screen screenshot, especially for the retina and 2K, 4K screens.

karensawrey:
Another question - how a retina/super-hi-res screenshot needs to be marked in the code now? Could you please provide an example of what a tag for such an image would look now?

olyism:
> Will the rule that the images need to have an even number of pixels still hold after this update?

Even pixels prevents sub-pixels so it looks nicer. But for better WX, I think the trade-off isn't worth the trouble.

styleguide/STYLE.md
olyism: Ah good point. I'll tweak this a bit. We should be taking screenshots on an @2x retina screen. So whilst the image is 1024 * 880, the native image size is 2048 * 1760

olyism:
> Another question - how a retina/super-hi-res screenshot needs to be marked in the code now? Could you please provide an example of what a tag for such an image would look now?

Right, so if we want to support retina we should be taking all screenshots from a retina screen.

I'll try to automate this workflow. Maybe there's a way for the backend to somehow get the image size and then set the 1/2 width. The idea is writers don't need to manually set the image size, the app should do it for you.

Ideally I'd like to do `srcset`, but we can look at that for a later iteration to stay with the scope.

I'll put this PR back to draft and have a look.

karensawrey:
Thanks for looking into that, @olyism! I believe, all of the screenshots taken within the last year were taken via retina or 4K devices (x/2 worked for 4K screenshots, too).

olyism:
Ok, the more I experiment with this WX improvement the more I think we should do retina properly. Because currently, everyone loads the @2x size image, regardless if. you have a retina screen or not.

This is the ideal workflow that I'm thinking:

1. The write uploads the image, say called _image.png_, and adds the image to the page
2. The app automatically generates _image.png_ and _image@2x.png_
3. The view outputs the responsive image markup

Like this:
```html
<img
  srcset="test.png, test.png 2x"
  src="test.png"
  alt="Test image"
>
```

A quick way is to use JavaScript to half the width and height. I think this option is hack-y in my opinion, although it removed the need to specify image size, it doesn't solve the retina image loading problem.

The last option is to ask the writers to process and upload two images — the @2x one and the resized one. This solves for retina but not workflow.

The approach used in this PR so far removes the need for setting image dimensions, whilst you lose the control of the image dimension, the template gracefully contain the image.

So~ I think I should do a spike to figure out the effort and feasibility of doing retina properly — that optimizes both the client-size experience and workflow.

Will update soon.

karensawrey:
Awesome! @olyism please take into consideration the fact that a good percentage of older screenshots were _not_ taken on Retina screens, so maybe we need a rule for "images bigger than X by Y to be treated as Retina and /2-ed". No idea if that's possible, though.

olyism:
> Awesome! @olyism please take into consideration the fact that a good percentage of older screenshots were not taken on Retina screens,

Totally! Should be possible by reading the file. But this is tricky though because some images would be small, like the AWS buttons, for instance.

I think we could do a batch search and replace and apply retina for all images doing the halving width & height thing (e.g. remove all `width: 1200/2, height: 600/2`, remove the dimensions and add a `retina => false` for all other bitmap image.

This would change the scope of this PR, so by going down the path of doing this properly, I think I'll close this PR and raise a separate one focused on addressing retina image.

PR #1696, opened 2022-08-12
Add note about not supporting gitlab commit status

jradtilbrook:
We've had a few support questions about gitlab commit statuses but unfortunately, we don't support them at the moment. So this is to clarify that.

I'm not sure if we want to put any reason or other information in here though

### DIFF
--- a/pages/integrations/gitlab.md.erb
+++ b/pages/integrations/gitlab.md.erb
@@ -39,4 +39,4 @@ If you host your repositories on [gitlab.com](https://gitlab.com/) enter your gi
 
 ## Commit statuses
 
-We don't support updating commit statuses on GitLab at the moment. Please get in touch with us at support@buildkite.com if you like this feature.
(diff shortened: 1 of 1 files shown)


karensawrey:
Thanks for the contribution, @jradtilbrook! I'm not sure I want to suggest people keep asking for the feature. Let me see if I can find a workaround :)

pages/integrations/gitlab.md.erb
```diff
@@ -36,3 +36,7 @@ If you host your repositories on [gitlab.com](https://gitlab.com/) enter your gi
 ## Build skipping
 
 <%= render_markdown partial: 'integrations/build_skipping' %>
+
+## Commit statuses
+
+We don't support updating commit statuses on GitLab at the moment. Please get in touch with us at support@buildkite.com if you like this feature.
```
karensawrey, line 42:
```suggestion
Updating commit statuses on GitLab is not supported at the moment. If you would like this feature to be added at some point, please let us know at support@buildkite.com.
```

karensawrey (approved):
I think it's now reworded it in the vague way I was after.

PR #1698, opened 2022-08-14
Add missing documentation on span objects 'detail' attribute

gchan:
PIE-1010

The docs are missing attributes for span objects (for HTTP and SQL).

A customer reached out about this as they were trying to figure out how
to POST the URL, method, and library details to the span object.

Customer message
https://buildkite-community.slack.com/archives/C03HXN4ARB5/p1660262430972669?thread_ts=1660256140.870549&cid=C03HXN4ARB5

Relevant code
https://github.com/buildkite/buildkite/blob/1e04bd5774c1c54927d15668b35c9bdfeae479f9/app/models/analytics/execution/ingestion_result.rb#L123

Docs page
https://buildkite.com/docs/test-analytics/importing-json#json-test-results-data-reference-test-result-objects

PR #1702, opened 2022-08-16
💄 Provide up to 6 levels of nesting for lists

olyism:
Following up with https://github.com/buildkite/docs/pull/1698 I think occasionally we may need deeper nested lists. In light of this, I've added up to 6 levels of nested lists to the CSS.

Unfortunately, CSS doesn't do something like `nth-level` selector, I think JavaScript is an overkill, and good content design shouldn't need to go beyond 6 levels of list nesting.

Screen Shot 2022-08-16 at 4 00 52 pm

### DIFF
--- a/app/assets/stylesheets/public/utils/_text_content.scss
+++ b/app/assets/stylesheets/public/utils/_text_content.scss
@@ -3,32 +3,62 @@
 
   h2, h3, h4, blockquote, pre, p, table { margin: 1rem 0; }
   h2 { font-size: #{(24/18)}rem; margin-top: 1.5em; }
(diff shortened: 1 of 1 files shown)


plaindocs (approved):
Looks good to me @olyism . Fully agree that we shouldn't need more than six.

I just had a quick look to see if we could set the max depth with a linter, but I couldn't find anything.

olyism:
> I just had a quick look to see if we could set the max depth with a linter, but I couldn't find anything.

Should be alright. I think we're all sensible to not do 6 levels of bullet lists 😛

PR #1703, opened 2022-08-16
Add GraphQL example of setting multiple env vars

jayco:
I had a question from a customer:

`AS the document shows PipelineScheduleUpdateInput has env: String not [String] . How can I update env with a list of variables such as env: [“bar=foo”, “var1"=“val1”, “var12"=“good”] ?`

Here's the mutation https://buildkite.com/user/graphql/console/2ad2ddd5-dd14-485c-b87d-af5751b3ec5b if you would like to check it out.

This probably applies to anything we take in as a String and separate with a new line - so there may be more opportunities.

Here is what the change looks like:

Screen Shot 2022-08-16 at 4 17 11 pm

You may want to consider linking to this recipe or adding something in https://buildkite.com/docs/apis/graphql/schemas/pipelinescheduleupdateinput to let folks know.

Another thing - this is kinda hacky; a better alternative would be to have its value that accepts an Array. I am not sure if that is on the Product roadmap, but just a FYI.

### DIFF
--- a/pages/apis/graphql/graphql_cookbook.md
+++ b/pages/apis/graphql/graphql_cookbook.md
@@ -1276,3 +1276,40 @@ mutation UpdateTeamPipelineReadonly {
   }
 }
 ```
+
+## Update multiple environment variables for a schedule
+
+This updates a schedules environment variables by using the new-line value "\n" as a delimiter.
+
+First, find the schedule's ID:
+
+```graphql
+query GetSchedules {
+  pipeline(slug: "organization/pipeline") {
+    schedules(first: 500) {
+      edges {
+        node {
+          id
+          label
+        }
+      }
+    }
+  }
+}
+```
+
+Then, use the ID to delete the user:
+
+```graphql
+mutation UpdateSchedule {
+  pipelineScheduleUpdate(input:{
+    id: "schedule-id"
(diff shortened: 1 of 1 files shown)


plaindocs (commented):
Thanks @jayco 

what I suggest here is switching this to the more generic

## Updating multiple environment variables

and use the Schedule as an example. Off the top of your head, do you know if there are more Mutations that use `env` in the same way? 

If you're happy with that approach I'll implement it.

jayco:
> Thanks @jayco
> 
> what I suggest here is switching this to the more generic
> 
> ## Updating multiple environment variables
> and use the Schedule as an example. Off the top of your head, do you know if there are more Mutations that use `env` in the same way?
> 
> If you're happy with that approach I'll implement it.

Oof - sorry for the late reply @plaindocs 

Yeah I did a quick check and this is used as a generic delimiter so I think that is the right call

lachlanhardy:
@jayco Heyo! Do you reckon this is mergeable or should the Onboardin team take another look at it?

dannymidnight:
I've brought this one back to life and moved it under pipelines. cc @mbelton-buildkite

mbelton-buildkite (approved):

PR #1704, opened 2022-08-16
Install test Analytics

plaindocs:


sj26 (approved):
Nice, let's give it a whirl

PR #1710, opened 2022-08-18
♻️ Edit tiles content in page, not separate yml

olyism:
Improve writer experience by editing tiles in the same place.

Previously tiles content are managed in `tiles.yml`. The original idea was to manage tiles like 'reusable content blocks' that can be inserted on any place. However, this use case has not emerged and it's making it more difficult for writers to manage content.

In the PR I've changed the tiles helper's parameter to take the array data, rather than the key in the yml file. The data structure has not been changed at all.


plaindocs (changes requested):
Let's take a moment to step back and think about how this should work. I know I said making changes one page is a good thing, but bringing the data into the page isn't necessarily the anwser to that.

I'll elaborate in Basecamp.

olyism:
Cool, I've drafted this PR. Let's discuss on Basecamp.

PR #1716, opened 2022-08-29
Default and maximum timeout for command steps info

karensawrey:
Info on specifying the default and maximum timeouts for the command step to prevent jobs from running for an undesirably long time or forever.


pages/pipelines/command_step.md.erb
```diff
@@ -201,7 +201,7 @@ _Optional attributes:_
...
+      The maximum number of minutes a job created from this step is allowed to run. If the job exceeds this time limit, or if it finishes with a non-zero exit status, the job is automatically canceled and the build fails. Jobs that time out with an exit status of 0 are marked as "passed".<br> To prevent jobs from consuming too many job minutes or running forever, you can also specify default and maximum timeouts from your organization's <a href="https://buildkite.com/organizations/~/pipeline-settings">Pipeline Settings page</a>, or on a pipeline's Builds settings page. Default timeouts apply in order of specificity — step's, pipeline's, and finally the organization's. Maximum timeout will be applied to any command step without a timeout or with a timeout greater than the maximum timeout.<br> Note that command steps on the Buildkite <a href="https://buildkite.com/pricing">Free plan</a> have a maximum job timeout of 240 minutes.<br>
```
plaindocs, line 204: This makes it sound like default and maximum are different things, but they aren't.

plaindocs: Regarding specificity, I'm not quite clear on the intention.

plaindocs (commented):
image

Could easily add a `Build timeout` section to the bottom of this file, where we explain in more detail, and disambiguate from the agent time out?

pages/pipelines/command_step.md.erb
karensawrey: But they are different things.

plaindocs: Ahah, yes, in *that* case we need to be more clear about this. :-p

pages/pipelines/command_step.md.erb
```diff
@@ -201,12 +201,24 @@ _Optional attributes:_
...
 
+### Build timeouts
+
+Timeouts for jobs can be specified as [command steps attributes](/docs/pipelines/command-step#timeout_in_minutes), but it's possible to avoid having to set them manually every time.
+
+To prevent jobs from consuming too many job minutes or running forever, you can specify default as well as maximum timeouts from your organization's [Pipeline Settings page](https://buildkite.com/organizations/~/pipeline-settings), or on a pipeline's Builds settings page.
+
+<%= image "default-timeout.png", width: 1724/2, height: 736/2, alt: "Set default and maximum timeout period for your jobs" %>
+
+Defaults apply in order of specificity — the step's timeout will be used if supplied, otherwise it defaults to the pipeline's default if supplied, or finally to the organization. Maximum timeout will be applied to any command step without a timeout or with a timeout greater than the maximum timeout.
```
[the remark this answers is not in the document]
karensawrey: Step, of course. It felt like it was clearly explained in the text. Looks like it wasn't. Any suggestions on how to make it clearer?..

plaindocs, line 218: How about this?

```suggestion
Default timeouts are _applied_ in order from most  to least specific — so an organization default takes precedence over a pipeline timeout, which in turn takes precedence over a step level timeout. 
Maximum timeouts are applied to any command step without a timeout or with a timeout greater than the maximum timeout.
```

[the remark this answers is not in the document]
plaindocs: WAIT, maybe I actually got that wrong.

You original "applied in order" makes me the think that the generic overrides the specific. Is that correct?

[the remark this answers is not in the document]
karensawrey: Yeah, you were correct about the most to the least specific, but the example had to go in reverse order. I've just corrected that.

PR #1719, opened 2022-09-05
Temporarily removing info about monitors

karensawrey:
- Leaves a mention
- Removes the pages & backlinks
- The initial commit might break the table structure for tiles.yml


karensawrey:
A less invasive approach is required

PR #1723, opened 2022-09-07
Fix webhooks overview overflow bug

preflightcookie:
Linear: https://linear.app/buildkite/issue/ONB-16/

Docs [Webhooks Overview page](https://buildkite.com/docs/apis/webhooks) content overflows to the right. This introduces a horizontal scrollbar where one is not expected or desired. Not identified on any other pages.

Impacting all viewport widths. Present with and without new nav banner.

Screen Shot 2022-09-06 at 11 36 00 am

Tracked root cause to the second HTTP headers table which has a long string of example text in it: `timestamp=1619071700,signature=30222eb518dc3fb61ec9e64dd78d163f62cb134a6ldb768f1d40e0edbn6e43f0` This doesn't have any whitespace in it and therefore isn't wrapping nicely

Tried a few options to fix, including adding spaces to the string (fixed the issue but traded off accuracy of the documentation) and trying unsuccessfully to get the text to wrap through various means. Landed on adding the option to make a table "fixed width". This forces the text to wrap.

This won't fix overflow in the future but gives us a tool to use when we encounter it without changing overall table design across docs, clipping content or reducing accuracy of documentation.

Fixed version:
Screen Shot 2022-09-07 at 4 44 48 pm

Also in this PR:
* Minor refactor to group media queries for `Docs__article` together. 
* Removed redundant CSS in `@media (min-width: $screen-lg)`
* Added Webhooks to API section of homepage
* Added `fixed-width` to styleguide

### DIFF
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -55,6 +55,7 @@ $color-green: rgb(3, 165, 98);
 .Docs__article {
   flex: 7;
   font-weight: 400;
+  overflow: hidden;
 
   @media (min-width: $screen-md) {
     margin-left: $sidebar-width + 30px;
--- a/app/assets/stylesheets/public/_docs.scss
+++ b/app/assets/stylesheets/public/_docs.scss
@@ -94,6 +94,7 @@ $color-green: rgb(3, 165, 98);
     grid-template-columns: minmax(10px, 1fr) 250px;
     position: relative;
     gap: 30px;
+    grid-template-columns: 1fr $sidebar-width;
 
     &--is-landing-page {
       display: block;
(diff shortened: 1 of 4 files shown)


plaindocs (approved):
Nice! 

Would be nice to mention `fixed-width` here https://github.com/buildkite/docs/blob/main/styleguide/STYLE.md#two-column-tables

And if you fancy cheating and adding Webhooks to this list on the home page that would probably make sense

image

lachlanhardy (approved):
Nice!

karensawrey (approved):
LGTM

preflightcookie:
Thanks all. Updated @plaindocs with both of your suggestions 👍 re-requested review

plaindocs (approved):
:shipit: :sun_with_face:

PR #1726, opened 2022-09-07
add note about environment variable precedence

jeremybumsted:
A user reached out to share some feedback around some seemingly contradicting spots in the docs around environment variable precedence:

---

The Step Defaults documentation states:

> If you’re using YAML steps, you can set defaults which will be applied to every command step in a pipeline unless they are overridden by the step itself. You can set default agent properties and default environment variables:
agents - A map of agent characteristics such as os or queue that restrict what agents the command will run on
env - A map of environment variables to apply to all steps

However the Environment variable precedence documentation then states:

>There are two places in a pipeline.yml file that you can set environment variables:
In the env attribute of command and trigger steps.
At the top of the yaml file, before you define your pipeline’s steps.
Defining an environment variable at the top of your yaml file will set that variable on each of the command steps in the pipeline, and is equivalent to setting the env attribute on every step. This includes further pipeline uploads through buildkite-agent pipeline upload.
| Top level pipeline environment variables will override what is set in the env attribute of an individual step.

Basically the first doc seems to imply you can define default environment variables which you can override at the step level (which would be nice), however the second doc - which is the truth - states that top-level (default) values override the env values declared in steps.
Strangely it seems the first doc is correct in terms of the agents configuration i.e. you can define defaults which can be overridden by an individual step.

---

Hopefully adding this small note will help direct folks to read about the environment variable precedence and keep things consistent 😄. If there is a better way to word this please do change it as needed!


pages/pipelines/defining_steps.md.erb
```diff
@@ -33,6 +33,11 @@ If you're using [YAML steps](/docs/tutorials/pipeline-upgrade), you can set defa
 * `agents` - A map of agent characteristics such as `os` or `queue` that restrict what agents the command will run on
 * `env` - A map of <a href="/docs/pipelines/environment-variables">environment variables</a> to apply to all steps
 
+<div class="Docs__note">
+  <p>Because environment variables can be set in many different places, it is recommended to familiarize yourself
+  with <a href="/docs/pipelines/environment_variables#environment-variable-precedence">environment variable precedence</a> to ensure your values don't get overwritten at runtime.</p>
```
plaindocs, line 38:
```suggestion
  <p>Because environment variables can be set in many different places, check <a href="/docs/pipelines/environment_variables#environment-variable-precedence">environment variable precedence</a> to make sure that your  environment variables work as you expect them to.</p>
```

plaindocs (approved):
Nice one @jeremybumsted 
I'll round to a major overhaul of those docs one day.

preflightcookie:
Hey @jeremybumsted how'd you go with this change? Feel free to formalise the PR and merge it whenever suits and thanks for making this improvement :)

preflightcookie (approved):

lachlanhardy:
@jeremybumsted Do you want to wrap this one or would you like me to take it?

mbelton-buildkite (approved):
I've pushed changes to update the format and made some tweaks to the language. I shouldn't have changed the meaning, but please double check. :)

PR #1736, opened 2022-09-26
Hint that the webhook signature be generated off the request body

timriley:
Change the variable name for the example webhook signature validation code from `payload` to `webhook_request_body`.

The variable name `payload` here could be misconstrued to be some kind of automatically deserialized request body data, when we actually want webhook recipients to use the raw body string itself.

### DIFF
--- a/pages/apis/webhooks.md.erb
+++ b/pages/apis/webhooks.md.erb
@@ -83,7 +83,7 @@ Buildkite can optionally send an HMAC signature in place of a webhook token.
 
 The `X-Buildkite-Signature` header contains a timestamp and an HMAC signature. The timestamp is prefixed by `timestamp=` and the signature is prefixed by `signature=`.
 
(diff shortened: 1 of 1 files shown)


sj26:
The explanation above doesn't mention the request body at all. I reckon some words up there might be good.

Under _Webhook signature_ there could be an extra sentence:

```markdown
Buildkite generates the signature using HMAC-SHA256; a hash-based message authentication code [HMAC](https://en.wikipedia.org/wiki/HMAC) used with the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash function and a secret key.
The webhook token value is used as the secret key.
The timestamp is an integer representation of a UTC timestamp.
+The raw request body is the signed message.+
```

And same in _Verifying HMAC signatures_:

```
Using the token as the secret along with the timestamp from the webhook, compute the expected signature based on the request body. There should be a library available in the programming language you are using that can perform this operation.
```

timriley:
Great point, @sj26; that single code example is doing heavier lifting than it should. I'll bring more words in like you suggest now.

timriley:
@sj26 Added :)

sj26 (approved):
Yeah, very nice 👍

preflightcookie (approved):

preflightcookie:
Looks lovely @timriley thanks! Feel free to merge when you're ready 🚀

PR #1746, opened 2022-10-19
update build.state available states table

jeremybumsted:
We are missing the `failing` state in a table on conditionals for `build.state`, this PR will fix that :)


lachlanhardy (approved):
LGTM.

:shipit:

PR #1747, opened 2022-10-20
Set an env var on prod to flag this is not a review app

elle:
The goal is to be able to decide whether Segment script is included in the views.
So that Segment only captures activity in production only.

env var settings in the `Dockerfile` (or more accurately `docker-compose.yml`) override the env vars that are set in Render UI.
Therefore, we set `RENDER_PR_APP` in the Render UI to be true, and this value will propagate to all the review apps. We then set `RENDER_PR_APP` to be false in `docker-compose.yml`

This can be a bit confusing if someone only looks at the Render UI. So hopefully this commit message will help explain the idea. More details and explorations are available in the Linear issue.

Linear: https://linear.app/buildkite/issue/ONB-29/🔗-integrate-docs-with-segment

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -5,6 +5,7 @@ services:
       context: .
       args:
         RAILS_ENV: production
+        RENDER_PR_APP: false
     ports:
       - '3000:3000'
     environment:


olyism (approved):

olyism:
Hi @elle, PR looks good, and thank you for elaborating on this.

A few minor comments:

> The goal is to be able to decide whether Segment script is included in the views.

Further to this point, we want to conditionally have a different Segment API ID in the script depending on the environment — i.e. production or not-production. This allows us to not contaminate production destinations' metrics (e.g. Google Analytics) as well as test tracking in non-production environments.

> We then set RENDER_PR_APP to be false in `docker-compose.yml`

I think you mean set RENDER_PR_APP to be false in `docker-compose.production.yml`?

lachlanhardy:
This gels with my understanding of the constraints. Nice!

PR #1749, opened 2022-10-20
Add specs to confirm Emojicom rendering

elle:
i.e. Emojicom should not be rendered on landing pages

Taken from https://github.com/buildkite/docs/pull/1609

### DIFF
--- a/spec/features/emojicom_rendering_spec.rb
+++ b/spec/features/emojicom_rendering_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+RSpec.feature "Emojicom rendering" do
+  context "home page" do
+    scenario "does not display emojicom widget" do
+      visit "/docs"
+
+      expect(page).not_to have_css "#emojicom-widget-inline"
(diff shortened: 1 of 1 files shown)


spec/features/emojicom_rendering_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.feature "Emojicom rendering" do
+  context "home page" do
+    scenario "does not display emojicom widget" do
+      visit "/docs"
+
+      expect(page).not_to have_text "#emojicom-widget-inline"
```
lachlanhardy, line 8: Is there a reason this one uses `have_text` while the others use `have_css`?

lachlanhardy (approved):
One minor consistency question, otherwise this is great!

spec/features/emojicom_rendering_spec.rb
elle: Nope, good catch. And updated

PR #1751, opened 2022-10-24
Test docker-compose set variables

elle:
Same as https://github.com/buildkite/docs/pull/1637 Just confirming that this works.
Will clean up afterwords


lachlanhardy (approved):
👍🏼

PR #1752, opened 2022-10-24
Test docker-compose set variables

lachlanhardy:
Same as https://github.com/buildkite/docs/pull/1637 Just confirming that this works.
Will clean up afterwords

Trying again


elle (approved):
🖌️

PR #1754, opened 2022-10-25
[PDP-234] Fix table for plugin hooks to show vendored plugins don't run until after checkout

triarius:
We did some investigating and it turns out the vendored plugins are not processed until after checkout. This is done in the `VendoredPluginPhase` in https://github.com/buildkite/agent/blob/9897c0a8178262a0331b01a23c389c992c711696/bootstrap/bootstrap.go#L711-L762. This is only called after checkout in bootstrap: https://github.com/buildkite/agent/blob/9897c0a8178262a0331b01a23c389c992c711696/bootstrap/bootstrap.go#L135-L146.

However, the last thing `VendoredPluginPhase` does is call the environment hooks on vendored plugins. Thus, the vendored plugins enviroment hooks actually run *after* the post-checkout hooks for other plugins or the agent.

## Before:
https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks
2022-10-25T20:56:28,650996755+11:00
## After:
https://bk-docs-pr-1754.onrender.com/docs/agent/v3/hooks#job-lifecycle-hooks
2022-10-25T20:58:30,980467813+11:00


moskyb (approved):

pda:
Aside, and out of scope for this PR: I would much prefer this:

```diff
-Non-vendored plugin
-Vendored plugin
+Plugin (non-vendored)
+Plugin (vendored)
```

Or perhaps even just “Plugin” and “Plugin (vendored)”, where “non-vendored” is implicit.

PR #1755, opened 2022-10-25
Set ENV VAR for Render PR Apps to `false`

lachlanhardy:
Again with ONB-29.

Setting `RENDER_PR_APP` to `false`

    This sets the ENV VAR to `false` in every environment. In Render,
    the ENV VAR has been set to `true` via the Render UI. This inherits
    to all PR environments inside Render, but can be overridden on
    individual PRs via the PR-specific environment settings.

@elle I think this means you can set the tracking VARs appropriately for Render or not. It doesn't solve setting them for other environments like TEST or DEVELOPMENT etc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ EXPOSE 3000
 # Let puma serve the static files
 ENV RAILS_SERVE_STATIC_FILES=true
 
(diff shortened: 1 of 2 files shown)


elle (approved):
Thanks for persevering with this @lachlanhardy

PR #1756, opened 2022-10-26
agent hooks: use “Plugin” and “Plugin (vendored)”

pda:
Turning https://github.com/buildkite/docs/pull/1754#issuecomment-1291521985 into a pull request…

Currently the agent hook table refers to plugins as “Vendored plugin” and “Non-vendored plugin” (see below).
I think the key info is “Plugin”, ~and that “non-vendored” is implicit unless specified otherwise~, so I propose we call them ~“Plugin”~ “Plugin (non-vendored)” and “Plugin (vendored)”.

~It could be argued that this leaves “Plugin” ambiguous in isolation, and could lead people to believe it includes vendored plugins e.g. in the `environment` hook. But I think the context makes it clear.~ If this is a concern, though, we could go with “Plugin (non-vendored)” and “Plugin (vendored)” to at least bring “Plugin” to the forefront. _(edit: we did that)_

Before:

197744031-943f7de1-78dd-4eff-86dc-9eb3804085f1

After: _(this hasn't been updated to reflect the “Plugin (non-vendored)” change)_

image


DrJosh9000 (approved):

triarius:
Totally agree with bringing "Plugin" to the fore. However, I vote for "Plugin (non-vendored)" and "Plugin (vendored)".

My reasoning is that if you only read the first 4 rows of the "After" screenshot, you may not realise that there is a distinction between them. If you arrive at this page with the impression that all plugins are vendored, you could leave with the impression that a vendored plugin could run at checkout, which is incorrect. One may argue that it does not make sense for a vendored plugin to run at checkout, which I agree with. However, it is quite reasonable to expect a vendored plugin to run at post-checkout, so if all you read from this table is the 4th row, you won't be informed otherwise.

If the bare "Plugin" were "Plugin (non-vendored)", this would be much clearer.

pzeballos:
My 2 cents here 😊, most users don't know what it's a vendor/non-vendored plugin, so I think it would be clearer to only make the clarification when it's a vendored plugin, and just call Plugin for the rest

triarius:
> My 2 cents here blush, most users don't know what it's a vendor/non-vendored plugin, so I think it would be clearer to only make the clarification when it's a vendored plugin, and just call Plugin for the rest

My concern is that it is not clear that "Plugin" is not a superset of "Plugin (vendored)" even if you read both.

pda:
I've gone with “Plugin (non-vendored)” and “Plugin (vendored)” for clarity, and it's still an improvement over “Non-vendored plugin” and “Vendored plugin”.

PR #1758, opened 2022-10-27
DRY layout files

elle:
... and other small cleanups

What has been done:

- Make `layout_by_path` private
  - This is a helper method that should not be accessed externally
- Disable auto generated files when using Rails generators
- Remove temp checks for `ENV` variables
- Define index action for homepage
  - To remove the confusion of where the home page is defined,
    making it easier to understand what is happening
  - Also removing one branch in the layout conditional
- Switch from `provide` to `content_for`
  - since we are not using or calling streaming anywhere in the app
  - Ref: https://api.rubyonrails.org/classes/ActionController/Streaming.html
- Match title in both layout files
  - in prep to clean up the layout views even more
- Pull out the head section in the three layout templates
  - to dry the duplication and to make it easier to see what is on the page
- Pull Algolia js tag into its own partial
  - Again, dry-ing the layouts templates

The principles behind the layout files cleanups are:

- Removing duplication
- Removing conditionals as much as possible
- Removing (or at least decreasing) the differences between the files - so that it is easier to see the differences and guide additional refactoring

Linear issue: https://linear.app/buildkite/issue/ONB-68/clean-up-application-layout-around-home-page

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -2,13 +2,15 @@ class PagesController < ApplicationController
   append_view_path "pages"
   layout :layout_by_path
 
+  def index
+    render :index, layout: "homepage"
+  end
+
   def show
     @page = Page.new(view_context, params[:path])
 
     # If the page doesn't exist, throw a 404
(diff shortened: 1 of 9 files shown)


preflightcookie (approved):
Lovely stuff, much cleaner 🎉

PR #1759, opened 2022-10-28
Introduce a layout for landing pages

elle:
which allows us to remove various conditional checks from the layout templates.

And on the way here, I also:

- Move application level partials to `app/views/application`
  - which is the Rails convention
  - this dir is within the lookup paths for template inheritance (see ref)
  - and means we don't need to explicitly reference the `layouts/` path
  - Ref: https://guides.rubyonrails.org/layouts_and_rendering.html#template-inheritance
- Pull home page header into its own partial
  - Just like is used in the default application layout template
- Pull footer for the home page back into the layout file
  - for more consistency
- Rename is_landing_page? to the Ruby convention
  - Also added specs
  - And removed references for it in the graphql layout,
    since it can just be defined as false
- Add a layout for a landing page
  - which means we added a branch for the `layout_by_path` again
    but we removed multiple checks from the layout views

These are small steps, in the right direction...
Linear: https://linear.app/buildkite/issue/ONB-69/create-a-layout-template-file-for-landing-page


lachlanhardy (approved):
Super clean!

PR #1763, opened 2022-11-03
Add Android collector to docs

meiqimichelle:
We've had an Android test collector for ages, but have not updated our public docs. We should fix that.

This PR:
- [x] Creates a new page to house information about our Android collector
- [x] Copy-pasta docs from the [collector readme](https://github.com/buildkite/test-collector-android/blob/main/README.md)
- [x] Deletes Android from the 'coming soon' page
- [x] Adds Android to the left nav
- [x] Adds Android collector to tiles on TA docs home
- [x] Adds 'beta' tag to Android collector

PR #1764, opened 2022-11-04
Add .NET collector to docs

meiqimichelle:
We've had a .NET test collector for ages, but have not updated our public docs. We should fix that. This is a companion PR to https://github.com/buildkite/docs/pull/1763 that does the same for Android.

This PR:
- [x] Creates a new page to house information about our .NET collector
- [x] Copy-pasta docs from the [collector readme](https://github.com/buildkite/test-collector-dotnet/edit/main/README.md)
- [x] Deletes .NET from the 'coming soon' page
- [x] Adds .NET to the left nav
- [x] Adds .NET collector to tiles on TA docs home
- [x] Adds 'beta' tag to .NET collector ~Should we remove the [beta] tag from all collectors entirely? It doesn't seem to be displaying correctly in the left nav anyways (will only show up for the collector at the bottom of the component list -- probs not worth the time to fix?) What do you think, @olyism ?~ See comment here for context: https://github.com/buildkite/docs/pull/1764#pullrequestreview-1175110673


data/nav.yml
```diff
@@ -299,6 +299,8 @@
             path: 'test-analytics/python-collectors'
           - name: 'Golang'
             path: 'test-analytics/golang-collectors'
+          - name: '.NET'
+            path: 'test-analytics/dotnet-collectors'
```
olyism, line 303: @meiqimichelle if we want to add the _beta pill_ in the left nav we can use the `pill` key.

```suggestion
            path: 'test-analytics/dotnet-collectors'
            pill: 'beta'
```

olyism (commented):
Hi @meiqimichelle thanks for uploading the .NET Test Collectors page. All the content looks good to me.

> Should we remove the [beta] tag from all collectors entirely? It doesn't seem to be displaying correctly in the left nav anyways

The beta feature isn't actually broken. You need to specify a 'pill' in `nav.yml` as well as in the [`beta_pages` model](https://github.com/buildkite/docs/blob/main/app/models/beta_pages.rb). My thoughts on the 'beta' tag is it's better to set consistent expectation across the [repo](https://github.com/buildkite/test-collector-dotnet) (which mentions beta in the page title) and here in Docs.

On a side note, I must admit the current writing experience (WX) of the beta pill is cumbersome. There's a backlog ticket for the team to address this but unfortunately for now we'll need to edit it in two different places.

data/nav.yml
meiqimichelle: Ohhhhhh icic. Thanks for the explanation! I was moving too quickly in-browser 😅

joshprice:
Was just about to raise an issue for this, glad to see this is close to publication!

PR #1765, opened 2022-11-04
Change from GA to Segment

elle:
- Add dotenv-rails gem
  - to be able to define development specific env variables
  for example for Segment
- 🔒️ Configure CSP
- Change from Google Analytics to Segment.io

Use a different value for production and development environments.

= Add SEGMENT_TRACKING_ID to .env.developemt
- On Render SEGMENT_TRACKING_ID is et to be the same as in dev env
   because Render uses the same env variables
  across prod and PR review apps
- Override the value of the SEGMENT_TRACKING_ID variable in Dockerfile
  so that production env will use a different key

Linear: https://linear.app/buildkite/issue/ONB-29/🔗-integrate-docs-with-segment

Also related: https://github.com/buildkite/docs/pull/1609/commits/4d8d794cc21cc478601ba93bd24973379f724d69

In local dev env:
local-dev-env

On Render:
render-pr-app


app/models/segment.rb
```diff
@@ -0,0 +1,9 @@
+module Segment
+  def self.key
```
lachlanhardy, line 2: Neat!

lachlanhardy (approved):
This looks great!

It's made me wonder though (sorry!) if we can avoid the model and conditional if we set the ENV_VAR to be just plain `SEGMENT_KEY` and change the value of it depending on environment. Would that work?

elle:
Good thinking. PR updated

PR #1768, opened 2022-11-07
Test commit for Render. We won't merge it

lachlanhardy:


lachlanhardy:
@olyism @elle 

Test worked. Segment is confirmed to be receiving both Development and Production data correctly. Closing this now.

PR #1771, opened 2022-11-09
🛎 reword API token comment

jfhinchcliffe:
After this change: https://github.com/buildkite/buildkite/pull/10259 this comment is no longer correct.
We will prohibit an org being added back to an API access token in the following situations:
- An organization revokes the token
- The User is removed from an organization

basecamp context: https://3.basecamp.com/3453178/buckets/3258372/messages/4855872984#__recording_5456583703


olyism (approved):
thanks @jfhinchcliffe. LGTM 🚀 please merge when you can.

PR #1546, opened 2022-05-31
Trialling `muffet` for link checking

lachlanhardy:
This resolves DOC-313:
https://linear.app/buildkite/issue/DOC-313/better-link-checker

PR #1778, opened 2022-11-21
Replace link following test with `muffet`

triarius:
This makes https://github.com/buildkite/docs/pull/1546 work.

Muffet is much faster. The builds used to take ~14-16 min. They are now around 2.5 min. (If the docker images are cached)

However, because rails is so slow to start up, I had to make the `muffet` container wait for a bit before running. Docker-Compose does not have a good way to say that a container should depend on the “readiness” of another container before it starts. So I simulated that with a loop polling the endpoint until it responds 2xx. Unfortunately, just polling the `/` route is not sufficient, as other routes seem to take a while to become “warm” (see some failed builds on this branch). So I chose a route that failed frequently and used it as the test to see if the server is “ready”.

Closes https://github.com/buildkite/docs/pull/1546

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -16,7 +16,18 @@ steps:
             - BUILDKITE_ANALYTICS_TOKEN
 
   - label: ":spider: muffet"
-    command: http://app:3000/docs --include=/docs/ --exclude=https://github.com/buildkite/docs/ --max-connections=10 --color=always --verbose
+    # We need to wait until rails has started before running muffet as otherwise it will error out
+    # and the test will appear to have failed without having run. The time to wait is hard to
+    # predict, and furthermore, some paths take longer to be be ready than others. The path in this
+    # loop was chosen after some non-systematic observations. So it does not guarantee that the
+    # server will be ready. But it seems to work well in practice.
+    command: >-
+      while ! wget --spider -Sq http://app:3000/docs/agent/v3/hooks;
+        do echo 💎🛤️🦥 Rails is still starting;
+        sleep 0.5;
+      done &&
+      echo 💎🛤️🚆 Rails has started running &&
+      /muffet http://app:3000/docs --include=/docs/ --exclude=https://github.com/buildkite/docs/ --max-connections=10 --color=always
     env:
       RAILS_ENV: test
     plugins:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -7,13 +7,27 @@ steps:
     env:
       RAILS_ENV: test
     plugins:
-      - docker-compose#v3.0.0:
+      - docker-compose#v3.9.0:
           run: app
           env:
             - BUILDKITE_JOB_ID
             - CI
             - RAILS_ENV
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -7,14 +7,34 @@ steps:
...
             - RAILS_ENV
             - BUILDKITE_ANALYTICS_TOKEN
 
+  - label: ":spider: muffet"
+    # We need to wait until rails has started before running muffet as otherwise it will error out
+    # and the test will appear to have failed without having run. The time to wait is hard to
+    # predict, and furthermore, some paths take longer to be be ready than others. The path in this
+    # loop was chosen after some non-systematic observations. So it does not guarantee that the
+    # server will be ready. But it seems to work well in practice.
+    command: 'while ! wget --spider -Sq http://app:3000/docs/agent/v3/hooks; do echo 💎🦥 Rails is still starting; sleep 0.5; done && echo 💎🏃 Rails has started running && /muffet http://app:3000/docs --include=/docs/ --exclude=https://github.com/buildkite/docs/ --max-connections=10 --color=always'
```
dabarrell, line 24: This was how I've done something similar in the past before (waiting for postgres to be ready for integration specs in docker compose). Nice. It'd be worth moving it into a separate file for readability though (`.buildkite/scripts/ci-muffet` or similar - not sure of our convention yet)

.buildkite/pipeline.yml
```diff
@@ -7,14 +7,34 @@ steps:
...
+      - docker-compose#v3.9.0:
           run: app
           env:
             - BUILDKITE_JOB_ID
             - CI
             - RAILS_ENV
             - BUILDKITE_ANALYTICS_TOKEN
 
+  - label: ":spider: muffet"
+    # We need to wait until rails has started before running muffet as otherwise it will error out
```
elle, line 19: May I suggest squashing all the commits into one and moving this comment to the commit message?

Reasoning behind this is: comments can become stale, while git log stays accurate with the current change

elle (commented):
Really like that it does what the test was doing just quicker. Absolutely great

.buildkite/pipeline.yml
triarius: I'm going to do both because if someone changes these lines, I think they are less likely to propagate an updated version of this message in their commit message than if the message were in a comment, them remembering to update the comment. This is especially so if they only make a cosmetic change to the lines. A comment would not be stale, but a commit message would be hidden behind the new commit.

.buildkite/pipeline.yml
triarius: @dabarrell It's a bit tricky as `comand` here isn't run directly on the ec2 instance that the buildkite-agent is on, but rather inside a docker container that it launches through the plugin. It's possible to get a script in a docker container, but it's very cumbersome if you are not building the image yourself, as is the case here.

However, I think the readability you want is achieved by using YAML multilane strings, so I have done that now.

sj26 (approved):
Looks great, runs fast, nice work 👍

plaindocs:
:tada:

PR #1781, opened 2022-11-24
Improve the Provisioning existing users section for SSO Okta docs

JuanitoFatas:
We recently learned more about how okta works and this section confuses our customers.

They are two options to provision existing users:

1. Remove and reassign users on okta; or
1. Use "Provision users" function (requires okta Lifecycle Management SKU feature)

CleanShot 2022-11-24 at 13 51 32@2x

More context at [this ticket](https://linear.app/buildkite/issue/FDN-1334/display-scim-panel-on-sso-page-for-all-orgs-with-an-appropriate-plan) (internal only)

### DIFF
--- a/pages/integrations/sso/okta.md.erb
+++ b/pages/integrations/sso/okta.md.erb
@@ -48,8 +48,10 @@ Existing Okta users aren't automatically provisioned in Buildkite; you'll need t
 
 This can be done one of two ways:
 
-1. Using the _Provision User_ function on the _Assignments_ tab of the Okta Buildkite app (if it's available), or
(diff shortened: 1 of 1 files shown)


olyism (commented):
Thanks @JuanitoFatas @yob LGTM 👍 

@jradtilbrook @pzeballos this relates to the OKTA issue we discussed during our last catch up. Would be keen to get your views on if this PR addresses the confusion customer has around provisioning users.

jradtilbrook (approved):
Yeah I think that should make it clearer

pzeballos (approved):

PR #1783, opened 2022-11-25
Remove `soft_failed` from available states

jradtilbrook:
`soft_failed` is actually a job state and does not apply at the build level so trying to compare to `soft_failed` results in an error


olyism (approved):
Thanks @jradtilbrook. LGTM 🚀 Please merge when you're ready.

PR #1786, opened 2022-12-05
Fix path to update agent help script

sj26:
I noticed the path to this script was incorrect while I was attempting to run it.


olyism (approved):
Thanks @sj26 for picking this up. LGTM 🚀

PR #1788, opened 2022-12-05
Remove empty env vars

sj26:
We're annotating the agent docs with environment variables that don't exist in some cases:

Screen Shot 2022-12-05 at 11 59 32 am

So skip that bit when there is no variable:

Screen Shot 2022-12-05 at 12 00 24 pm


olyism (approved):
Thanks @sj26 for putting together this improvement. LGTM 🚀

PR #1793, opened 2022-12-06
Add GitHub Action for labelling PRs

dannymidnight:
After recently browsing through PRs in this repository I thought it might be nice to add some basic categorisation in the form of GitHub labels. 

This PR adds a new GitHub Action Workflow to automatically apply labels (ie. `pipelines`, `test-analytics`, `agent`, etc). 
 I'm sure there's plenty more labels we could add but these seem like the most useful.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,14 @@
+agent:
+  - pages/agent/**/*
+test-analytics:
+  - pages/test_analytics/**/*
+pipelines:
+  - pages/pipelines/**/*
+plugins:
+  - pages/plugins/**/*
+integrations:
+  - pages/integrations/**/*
+apis:
+  - pages/apis/**/*
+tutorials:
+  - pages/tutorials/**/*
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+pipelines:
+  - pages/pipelines/**/*
+plugins:
+  - pages/plugins/**/*
+integrations:
+  - pages/integrations/**/*
+apis:
+  - pages/apis/**/*
+tutorials:
+  - pages/tutorials/**/*
```
mbelton-buildkite, line 14: I don't think this one is super useful. It looks like a generic label for tutorial-type content, but that whole directory is specific to Pipelines. I think it would be better to have the _Pipelines_ label attached instead.

mbelton-buildkite (commented):
I think one label could change, otherwise looks good.

mbelton-buildkite (approved):
LGTM

PR #1798, opened 2022-12-08
Update CSP for Emojicom

lachlanhardy:
This resolves ONB-108:
https://linear.app/buildkite/issue/ONB-108/csp-issues-with-emojicom-in-docs


lachlanhardy:
This PR has two extra commits so I've screwed something up. Going to twiddle with my machine's understanding of `main` and fix it.

PR #1799, opened 2022-12-09
Update CSP for Emojicom

lachlanhardy:
This resolves ONB-108:
https://linear.app/buildkite/issue/ONB-108/csp-issues-with-emojicom-in-docs

This PR replaces https://github.com/buildkite/docs/pull/1798 because git

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -18,15 +18,16 @@
 
 Rails.application.config.content_security_policy do |policy|
   policy.default_src :self
-  policy.font_src    :self, 'https://www2.buildkiteassets.com/'
-  policy.img_src     :self, 'https://buildkiteassets.com/', 'https://buildkite.com/', ENV.fetch('BADGE_DOMAIN', 'https://badge.buildkite.com')
+  policy.font_src    :self, "https://www2.buildkiteassets.com/"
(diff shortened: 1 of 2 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -25,8 +25,9 @@
 
   policy.script_src(
     :self,
-    "https://www.googletagmanager.com/",
-    "https://cdn.segment.com/"
+    'https://www.googletagmanager.com/',
+    'https://cdn.segment.com/',
+    'https://cdn.emojicom.io/'
```
elle, line 30: Should we standardise on linter config?

elle (approved):
Other than one question about linter config, LGTM

config/initializers/content_security_policy.rb
lachlanhardy: I prefer double quotes so that they're consistent everywhere. I assumed it was a project-based linter setting, but perhaps it was something in my local setup?

Do you want me to change them back here (or rather remove that commit) or we'll fix it moving forward?

dannymidnight (approved):
Changes LGTM.

config/initializers/content_security_policy.rb
dannymidnight: Looks like there's nothing specific to quotations in the our RuboCop/linter config. If that's the convention you've seen kicking around then I'd stick with what you have.

PR #1800, opened 2022-12-09
Update agent tokens docs for job tokens beta

sj26:
We have a beta feature built into the agent and supported from the agent api allowing access tokens to be specified per job. The agent token management form links to this place, so we should cover the option available in docs and link back to token management for folks coming back in the opposite direction.

The new agent page, including agent token management, and this checkbox within it are not yet released, but this is an early PR with some words to support the release.

Screen Shot 2022-12-09 at 12 28 53 pm


tessereth:
Superseded by https://github.com/buildkite/docs/pull/2609

PR #1802, opened 2022-12-12
Remove notification about recently updated navigation

elle:
Closes #ONB-84

Linear: https://linear.app/buildkite/issue/ONB-84/remove-recently-updated-nav-message-at-top-of-docs-pages


lachlanhardy (approved):
Just ripping the whole thing out, huh?

I'm wary of doing that without a replacement, but we can always grab the code from the history if we need it and I'd rather clean it out than leave empty infrastructure we may not use again.

elle:
Thanks @lachlanhardy. Merging shortly

PR #1808, opened 2022-12-19
Fix link to Getting started page

mbelton-buildkite:
Given the location and link text, this appears to be a mistake.


dannymidnight (approved):
👍

PR #1809, opened 2022-12-21
Adds if and skip attributes

karensawrey:
This PR adds `if` and `skip` optional attributes to the group step.
Updated after a Slack inquiry.


karensawrey:
Works as it should:
https://bk-docs-pr-1809.onrender.com/docs/pipelines/group-step
attr-added

pzeballos (approved):
LGTM

PR #1810, opened 2022-12-22
Adds BUILDKITE_PLUGIN_CONFIGURATION env var + more

karensawrey:
This PR adds BUILDKITE_PLUGIN_CONFIGURATION environment variable according to https://github.com/buildkite/agent/pull/1382. 

It also adds the information about BUILDKITE_PLUGIN_NAME from the same PR. This part is a separate commit and can be removed.

### DIFF
--- a/data/content/environment_variables.yaml
+++ b/data/content/environment_variables.yaml
@@ -298,7 +298,7 @@ variables:
 - name: BUILDKITE_PLUGIN_CONFIGURATION
   desc: |
     A JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
-  example: "{\"array-key\":[{\"subkey\":[1,2,\"llamas\"]}]}"
+  example: "{\"image\":\"node:lts-alpine3.14\"}"
 - name: BUILDKITE_PLUGIN_NAME
   desc: |
     The "env key-compatible" version of the current plugin name. Used for constructing the `BUILDKITE_PLUGIN_<name>_<param>` variables.
--- a/data/content/environment_variables.yaml
+++ b/data/content/environment_variables.yaml
@@ -297,7 +297,7 @@ variables:
   example: "deploy:ops:production"
 - name: BUILDKITE_PLUGIN_CONFIGURATION
   desc: |
-    JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
+    A JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
   example: "{\"array-key\":[{\"subkey\":[1,2,\"llamas\"]}]}"
 - name: BUILDKITE_PLUGIN_NAME
   desc: |
--- a/data/content/environment_variables.yaml
+++ b/data/content/environment_variables.yaml
@@ -295,6 +295,10 @@ variables:
     A colon separated list of the pipeline's non-private team slugs.
   modifiable: false
   example: "deploy:ops:production"
+- name: BUILDKITE_PLUGIN_CONFIGURATION
+  desc: |
+    JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the BUILDKITE_PLUGINS environment variable).
+  example: "{\"array-key\":[{\"subkey\":[1,2,\"llamas\"]}]}"
 - name: BUILDKITE_PLUGINS
   desc: |
     A JSON object containing a list plugins used in the step, and their configuration.
--- a/data/content/environment_variables.yaml
+++ b/data/content/environment_variables.yaml
@@ -301,7 +301,7 @@ variables:
   example: "{\"image\":\"node:lts-alpine3.14\"}"
 - name: BUILDKITE_PLUGIN_NAME
   desc: |
-    The "env key-compatible" version of the current plugin name. Used for constructing the `BUILDKITE_PLUGIN_<name>_<param>` variables.
+    The current plugin's name, with all letters in uppercase and any spaces replaced with underscores.
   example: "DOCKER"
 - name: BUILDKITE_PLUGINS
   desc: |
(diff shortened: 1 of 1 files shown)


karensawrey:
Preview:
added_env_vars

data/content/environment_variables.yaml
```diff
@@ -295,6 +295,14 @@ variables:
     A colon separated list of the pipeline's non-private team slugs.
   modifiable: false
   example: "deploy:ops:production"
+- name: BUILDKITE_PLUGIN_CONFIGURATION
+  desc: |
+    JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
```
mbelton-buildkite, line 300: Based on the other table entries, it looks like we include the article at the start of the description:
```suggestion
    A JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
```

mbelton-buildkite: Looking at the entry for `BUILDKITE_PLUGINS`, I notice it's described as a _JSON object_ rather than a _JSON string_. Should they both be the same, or is this one different?

data/content/environment_variables.yaml
```diff
@@ -295,6 +295,14 @@ variables:
...
+  example: "{\"array-key\":[{\"subkey\":[1,2,\"llamas\"]}]}"
```
mbelton-buildkite, line 301: The other examples on the page look to use realistic values rather than descriptive text like _array-key_ and _subkey_. Are there other example values that might work better here?

data/content/environment_variables.yaml
```diff
@@ -295,6 +295,14 @@ variables:
...
+- name: BUILDKITE_PLUGIN_NAME
+  desc: |
+    The "env key-compatible" version of the current plugin name. Used for constructing the `BUILDKITE_PLUGIN_<name>_<param>` variables.
```
mbelton-buildkite, line 304: I find this a bit confusing because I can't see "env key-compatible" naming or `BUILDKITE_PLUGIN_<name>_<param>` described on this page. Looking at [this page](https://github.com/buildkite/docs/blob/main/pages/plugins/writing.md.erb#L49), maybe something like:

> The current plugin's name (as opposed to the names of all plugins). The value matches the GitHub repository slug, not the name defined in `plugin.yml`. You can use the value to access environment variables specific to the plugin using the format: `BUILDKITE_PLUGIN_<PLUGIN_NAME>_<CONFIGURATION_PROPERTY>`.

What do you think?

mbelton-buildkite (commented):
Thanks @karensawrey! I've left a few comments for you to consider. 😊

data/content/environment_variables.yaml
karensawrey: BUILDKITE_PLUGINS _is_ a JSON object (that contains a list of plugins and their configs) :) A string is just a string.

data/content/environment_variables.yaml
karensawrey: I'm very open to a more descriptive example, @mbelton-buildkite! :) At the moment, I cannot suggest a better one - maybe you have something cool in mind?

data/content/environment_variables.yaml
karensawrey: This honestly sounds more confusing to me.

pzeballos: My 2 cents here... I also think it's more confusing. Maybe not so much explanation, like for example:

> The current plugin's name. The value matches the GitHub repository slug.

I don't think it is necessary to add more information. Keep it simple like we do with the others variables

data/content/environment_variables.yaml
mbelton-buildkite: Nothing cool, but could we use the info from the `BUILDKITE_PLUGINS` entry?

```suggestion
  example: "{\"github.com/buildkite-plugins/docker-buildkite-plugin#v3.7.0\":{\"image\":\"node:lts-alpine3.14\"}}"
```

Is that correct? Maybe not, since it doesn't have the config in arrays. If there's no simple example you can put in, it's okay to leave it as it is. :)

data/content/environment_variables.yaml
mbelton-buildkite: Yeah, from what I can see on the other page I mentioned, we don't have a good way to talk about these values for plugins. The `name` field in a `plugin.yml` can be a combination of upper and lower case letters, and can contain spaces. It _looks_ like this value is an uppercase version of the name, with spaces replaced with underscores. It's different from the GitHub slug because it's encouraged for that to include the suffix `-buildkite-plugin`. Is that right? If so, perhaps include that in the description. For example:

> The current plugin's name, with all letters in uppercase and any spaces replaced with underscores.

What do you think?

karensawrey:
@pzeballos can you please take a look at the remaining suggestions? They LGTM, but I'd love to have a more knowledgeable pair of eyes on the new example. Thanks a million!

data/content/environment_variables.yaml
pzeballos: Actually it would only have the inner dictionary:

```suggestion
  example: "{\"image\":\"node:lts-alpine3.14\"}"
```

data/content/environment_variables.yaml
pzeballos: Looks good!

data/content/environment_variables.yaml
```diff
@@ -295,6 +295,14 @@ variables:
...
   modifiable: false
   example: "deploy:ops:production"
+- name: BUILDKITE_PLUGIN_CONFIGURATION
+  desc: |
+    A JSON string holding the current plugin's configuration (as opposed to all the plugin configurations in the `BUILDKITE_PLUGINS` environment variable).
+  example: "{\"array-key\":[{\"subkey\":[1,2,\"llamas\"]}]}"
+- name: BUILDKITE_PLUGIN_NAME
+  desc: |
+    The "env key-compatible" version of the current plugin name. Used for constructing the `BUILDKITE_PLUGIN_<name>_<param>` variables.
+  example: "DOCKER_COMPOSE"
```
pzeballos, line 305: To be consistent with the other examples you should use:
```suggestion
  example: "DOCKER"
```

karensawrey:
@mbelton-buildkite please take a look at this one & let's finally merge :)

mbelton-buildkite (approved):
Looks good to me! Thanks 😊

PR #1820, opened 2023-01-09
Fix typo in tags-from-host default value

nick-f:
Tiny fix to update a typo in https://buildkite.com/docs/agent/v3/configuration#tags-from-host

image


dannymidnight (approved):
Thanks!

PR #1822, opened 2023-01-09
Add copy button on code examples

dannymidnight:
It's annoying having to manually copy code snippets from the docs into a shell / editor. 

This PR adds the most basic version of a "Copy to clipboard" button that becomes visible on hover and focus for all code blocks. 

This change still needs a round of design but you get the gist. There's currently no success state (i.e. "Copied 👍🏻) and I've totally guessed my way around icons! 

https://user-images.githubusercontent.com/656826/211252617-a90917cf-2424-4fd5-a973-ecabdefb80b9.mov

### DIFF
--- a/app/assets/stylesheets/pygments.scss
+++ b/app/assets/stylesheets/pygments.scss
@@ -66,30 +66,4 @@
 
 .highlight {
   position: relative;
-
-  &:focus, &:hover {
-    .highlight__button {
-      visibility: visible;
-    }
-  }
-}
-
-.highlight__button {
-  visibility: hidden;
-  position: absolute;
-  right: .8em;
-  top: 1em;
-  padding: 5px;
-  cursor: pointer;
-
-  > svg {
-    height: 100%;
-    width: 100%;
-    padding: 0;
-    border: 0;
-  }
-
-  &:hover {
-    visibility: visible;
-  }
 }
--- a/app/assets/stylesheets/pygments.scss
+++ b/app/assets/stylesheets/pygments.scss
@@ -67,7 +67,7 @@
 .highlight {
   position: relative;
 
-  &:hover {
+  &:focus, &:hover {
     .highlight__button {
       visibility: visible;
     }
--- a/app/assets/stylesheets/pygments.scss
+++ b/app/assets/stylesheets/pygments.scss
@@ -63,3 +63,33 @@
 .highlight .ld, /* Literal.Date */
 .highlight .m   /* Literal.Number */
 { color: #ca1655; }
+
+.highlight {
+  position: relative;
+
+  &:hover {
+    .highlight__button {
+      visibility: visible;
+    }
+  }
+}
+
+.highlight__button {
+  visibility: hidden;
+  position: absolute;
+  right: .8em;
+  top: 1em;
+  padding: 5px;
+  cursor: pointer;
+
+  > svg {
+    height: 100%;
+    width: 100%;
+    padding: 0;
+    border: 0;
+  }
+
+  &:hover {
+    visibility: visible;
+  }
+}
--- a/app/models/page/renderer.rb
+++ b/app/models/page/renderer.rb
@@ -57,9 +57,11 @@ class Page::Renderer
   end
 
   def add_copy_to_clipboard_button(doc)
-    doc.css("div.highlight").map do |block|
+    doc.css("div.highlight").map do |node|
+      node.set_attribute("tabindex", 0)
+
       # Render clipboard icon from https://www.figma.com/file/zJ2OfBZPY0bYGkv0TZ2FNR/Heroicons?node-id=2%3A943&t=igXNWnw6MjFltc3X-4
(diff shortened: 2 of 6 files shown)


app/models/page/renderer.rb
```diff
@@ -22,6 +22,8 @@ def render(text, options = {})
     doc = add_code_filenames(doc)
     doc = add_callout(doc)
     doc = init_responsive_tables(doc)
+    doc = add_copy_to_clipboard_button(doc)
```
olyism, line 25: Here's an idea: what are your thoughts around adding the copy button with JavaScript rather than using the renderer?

I think this is worth a try because:

1. Separation of concern: renderer's focus is to render the actual content, whilst JavaScript provides UI functionality
2. Less work for the renderer, let the client-side do the work

app/assets/stylesheets/pygments.scss
```diff
@@ -63,3 +63,33 @@
...
+
+.highlight__button {
+  visibility: hidden;
+  position: absolute;
+  right: .8em;
+  top: 1em;
+  padding: 5px;
+  cursor: pointer;
+
+  > svg {
```
olyism, line 85: Maybe better to stick with class selector to avoid nesting? e.g. `.highlight__copy-icon` or `highlight__button-copy-icon`?

app/assets/stylesheets/pygments.scss
```diff
@@ -63,3 +63,33 @@
...
+    .highlight__button {
+      visibility: visible;
+    }
+  }
+}
+
+.highlight__button {
+  visibility: hidden;
+  position: absolute;
+  right: .8em;
```
olyism, line 80: Let's use `right: .4em;` and `top: .4em;` here, as there are rare occasions one-line code blocks.

Screen Shot 2023-01-09 at 11 22 54 pm

app/assets/stylesheets/pygments.scss
```diff
@@ -63,3 +63,33 @@
 .highlight .ld, /* Literal.Date */
 .highlight .m   /* Literal.Number */
 { color: #ca1655; }
+
+.highlight {
+  position: relative;
+
+  &:focus, &:hover {
+    .highlight__button {
+      visibility: visible;
```
olyism, line 72: I suggest we can use `opacity` instead of `visibility` to toggle the copy button because you can tab to a 0 opacity button (whereas you can't do so with `visibility: hidden;` or `display: none;`.

If we take the `opacity` approach, we can remove the `tabindex="1"` in the `div.highlight` element.

Screen Shot 2023-01-09 at 11 29 21 pm

☝️ Screenshot on my local dev, using the `opacity` approach, and I can tab to focus on the copy button without mouseover.

olyism:
Hi @dannymidnight this looks really awesome and works really well. Few things from me:

1. Would you like a hand with the 'Copied' UI in this PR? If so, I can add a commit (either via another PR or push directly to this branch, whichever you might prefer)
2. Show/hide copy button: a suggestion from me would be to use `opacity`, instead of `visibility`, to show/hide the button. This allows the button to be focused (see [comment above](https://github.com/buildkite/docs/pull/1822#pullrequestreview-1240403006)).
3. Maybe we could consider using JavaScript to render the button? This way, we use the renderer for content, as well as reduce renderer's workload.
4. You've used the correct icon for copy ✅ FYI, the design team's moving towards using [Heroicons](https://heroicons.com/) (by Tailwind) as our icon set

Lastly, I've added some styling nitpicking suggestions. Thoughts?

mbelton-buildkite (commented):
Yesss! Keen to see this live.

A couple of thoughts:
- It needs a success state to show that the action happened.
- I'd prefer it to stay visible rather than on hover so that it's clear the functionality exists to people. I'm not too worried about this point, though.

olyism:
As a thought, to keep the scope small we can just copy the app's copy-pasta feature and the success state so we can ship this sooner.

Screen Shot 2023-01-10 at 11 50 17 am

I can do a more delightful success state design in a separate PR, and gives me an opportunity to establish a common pattern with other teams.

Thoughts?

app/models/page/renderer.rb
dannymidnight: Good call out. I think we probably still want to attach some kind of data attribute to make the javascript not dependant on CSS classes since this often leads to unintentional breakages but I think it's a good suggestion based on where it landed 👍🏻

app/assets/stylesheets/pygments.scss
dannymidnight: SVGs are slightly awkward apply classes so I've kept this as is.

dannymidnight:
> As a thought, to keep the scope small we can just copy the app's copy-pasta feature and the success state so we can ship this sooner.

I had a play around pulling that component out but alas it's using React and we're not currently using that in docs atm. I've had another play around that I _think_ is slightly nicer. Keen for another round of feedback @olyism @mbelton-buildkite!

Here's another quick demo with hover / focus:

https://user-images.githubusercontent.com/656826/211464767-9b7a4ab3-602d-4494-b5c2-4240cd640e2a.mov

mbelton-buildkite (approved):
The behaviour looks good to me! 🥳

olyism (approved):

PR #1826, opened 2023-01-16
Adding a missing /etc/ in the file location for hooks.

mcncl:
File location for hooks is ` /opt/homebrew/etc/buildkite-agent/hooks`, not ` /opt/homebrew//buildkite-agent/hooks`


pages/pipelines/scheduled_builds.md.erb
```diff
@@ -91,3 +91,6 @@ For more information on how modulo works, see the official documentation of [Fug
   <tr><th><code>0 16 L * *</code></th><td>The last day of the month at 4pm UTC</td></tr>
   <tr><th><code>0 0 * * 2%2+1</code></th><td>The start of every odd Tuesday</td></tr>
 </table>
+
+>🚧 Every X minutes use
+> <code>*/N * * * *</code> only supports factors of 60, so intervals of 10, 20, 30 minutes will work as expected. However, due to limitations of Cron, schedules cannot, for example, be ran every 45 minutes using <code>*/45 * * * *</code>. This will result in a build being run on the hour, for example 12:00, and 45 minutes past, 12:45.
```
mbelton-buildkite, line 96: This doesn't appear related to the title and description of the PR. Is it meant to be included in this change?

I'm also not sure about the location. Does this relate to the slash operator? If so, I think it would be better to include this information where the slash operator is defined. Perhaps adding a new paragraph there along the lines of:

```
When defining a period of minutes, the slash operator only supports factors of 60 due to a limitation of Cron. For example, using `/10`, `/20`, or `/30` results in a build running at each period, as expected. However, using`/45` results in a build running on the hour and at 45 minutes past the hour since 45 is not a factor of 60.
```

Are there similar constraints for the other positions, like hours or months?

mbelton-buildkite (commented):
The update to the file location looks good to me, but I'm not sure about the other change.

mcncl:
Closing PR in favour of new one as commits getting messy

PR #1829, opened 2023-01-18
Add warning around modulo caveats

mcncl:
Modulo cannot be used with values that are not factors of 60. This causes builds to run at the top of the hour and N past the hour, where N is the value in minutes for the modulo run.

Example:
`*/45 * * * *` will not only run builds at 45 minutes past the hour, but also at the top or the hour; 12:00, 12:45, 13:00, 13:45, and so on.


mbelton-buildkite (commented):
Thanks for putting this change together. I'd like to clarify a few things first:

- Is this a caveat to the modulo? It looks to be about the slash command.
- What's the reason for placing the warning below the table? It looks like it should sit alongside the definition of the concept in one of the sections above.
- Is the caveat only related to minutes? Or, are there similar constraints for the other positions, like hours or months?

PR #1833, opened 2023-01-19
Block step and input step docs: always render dropdown on blockfieldmodal if required: false

l-suzuki:
### Description ✏️
Recently Pipelines Edge had an [escalation](https://3.basecamp.com/3453178/buckets/23112918/card_tables/cards/5744680321) where a customer was confused as to why the BlockFieldsModal was rendering a dropdown when they had less than 6 options provided. Turns out it's because the `required` field was set to `false`. It's expected behaviour in that it's purposely build into the code, but it also took us a minute to figure out whether this was a bug or not, and why it was implemented this way, as this behaviour is not currently mentioned in the docs! 

The reason is this:
If required == false and radio buttons are rendered, the user cannot "unselect" once an option has been checked. 

This PR adds a bit of extra context to the docs to avoid this confusion in the future.

### Context 🥟
[Linear Ticket](https://linear.app/buildkite/issue/EDGE-507/samsara-possible-ui-bug-on-select-input)

### DIFF
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -253,6 +253,7 @@ _Optional attributes:_
     <td><code>required</code></td>
     <td>
       A boolean value that defines whether the field is required for form submission.<br>
+      When this value is set to <code>false</code>, non-multiple options will always displayed in a dropdown box, regardless of how many are provided.<br>
       <em>Default:</em> <code>true</code>
     </td>
   </tr>
--- a/pages/pipelines/block_step.md.erb
+++ b/pages/pipelines/block_step.md.erb
@@ -18,7 +18,7 @@ steps:
 
 <%= image "block-step.png", width: 944/2, height: 364/2, alt: "Screenshot of a basic block step" %>
 
-You can add form fields to block steps by adding a `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items. Block steps with input fields can only be defined using a `pipeline.yml`.
+You can add form `fields` to block steps by adding a fields attribute. Block steps with input fields can only be defined using a `pipeline.yml`. There are two field types available: text or select. The select input type displays differently depending on how you configure the options. If you allow people to select multiple options, they display as checkboxes. If you are required to select only one option from six or fewer, they display as radio buttons. Otherwise, the options display in a dropdown menu.
 
 <%= image "block-step-with-fields.png", width: 944/2, height: 1178/2, alt: "Screenshot of a block step with input fields" %>
 
@@ -253,7 +253,7 @@ _Optional attributes:_
     <td><code>required</code></td>
     <td>
       A boolean value that defines whether the field is required for form submission.<br>
-      When this value is set to <code>false</code>, non-multiple options will always displayed in a dropdown box, regardless of how many are provided.<br>
+      When this value is set to <code>false</code> and users can only select one option, the options display in a dropdown menu, regardless of how many options there are.<br>
       <em>Default:</em> <code>true</code>
     </td>
   </tr>
--- a/pages/pipelines/input_step.md.erb
+++ b/pages/pipelines/input_step.md.erb
@@ -19,7 +19,7 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
-You can add form fields to input steps by adding the `fields` attribute. There are two field types available: text or select. The select input type is displayed as radio buttons when it contains less than six items, and as a drop down list when it contains more than six items.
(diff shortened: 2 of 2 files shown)


pages/pipelines/block_step.md.erb
```diff
@@ -253,6 +253,7 @@ _Optional attributes:_
     <td><code>required</code></td>
     <td>
       A boolean value that defines whether the field is required for form submission.<br>
+      When this value is set to <code>false</code>, non-multiple options will always displayed in a dropdown box, regardless of how many are provided.<br>
```
l-suzuki, line 256: I found this a bit difficult to articulate in a clear manner, and also was unsure as to how much detail to add, so keen for some advice on wording here :)

mbelton-buildkite: Okay, let me check the logic first:

- if (can select multiple options) -> Checkboxes
- else if (options > 6 || !required) -> Dropdown
- else -> radio buttons

So, it only displays as radio buttons when you can only select one option AND you're required to select an option AND there are 6 or fewer options.

Is that right? If so, perhaps:

```suggestion
      When this value is set to <code>false</code> and users can only select one option, the options display in a dropdown menu, regardless of how many options there are.
<br>
```

pages/pipelines/block_step.md.erb
```diff
@@ -253,6 +253,7 @@ _Optional attributes:_
...
       <em>Default:</em> <code>true</code>
     </td>
   </tr>
```
mbelton-buildkite, line 259: Rather than overloading the tables, I think adding the explanation further up the page would be better. You could update the paragraph at line 21 with something like:

> You can add form fields to block steps by adding a `fields` attribute. Block steps with input fields can only be defined using a `pipeline.yml.` There are two field types available: text or select. The select input type displays differently depending on how you configure the options. If you allow people to select multiple options, they display as checkboxes. If you are required to select only one option from six or fewer, they display as radio buttons. Otherwise, the options display in a dropdown menu.

What do you think?

mbelton-buildkite (commented):
This is a bit of an awkward one, but I've suggested updating the description in a more general spot and removing this extra information from the tables.

pages/pipelines/block_step.md.erb
l-suzuki: @mbelton-buildkite just to clarify your meaning here - do you mean we would add this to line 21 and not add anything on line 256? 

(If so, I'm in two minds about it - 
I get what you mean about not wanting to tables to be too bulked out with text. But, on the other hand,
when I'm looking up something in the docs, if I'm already somewhat familiar with the thing I'm looking up, but wanting to check a detail about it (like in this case), I will usually jump straight to the attributes section as I assume that's where the nitty-gritty mechanics are listed out fully... so if something like this was mentioned above, but not in the table, I might be a bit confused by that.) 

Keen to hear your thoughts here!

pages/pipelines/block_step.md.erb
l-suzuki: Yes that wording is much clearer, good call.

l-suzuki: And yes that logic is correct!

mbelton-buildkite: Yay! I'll add this suggestion for the other page, too.

pages/pipelines/input_step.md.erb
```diff
@@ -238,6 +238,7 @@ _Optional attributes:_
     <td><code>required</code></td>
     <td>
       A boolean value that defines whether the field is required for form submission.<br>
+      When this value is set to <code>false</code>, non-multiple options will always displayed in a dropdown box, regardless of how many are provided.<br>
```
mbelton-buildkite, line 241: As suggested on the other page:

```suggestion
      When this value is set to <code>false</code> and users can only select one option, the options display in a dropdown menu, regardless of how many options there are.<br>
```

pages/pipelines/block_step.md.erb
mbelton-buildkite: Yeah, I get why you're torn. What you've described is a common way for experienced users to behave. The question is: Is this information important for experienced users to know when they look up this field? I'm on the fence, so I'm happy to go with your gut and include it in the table. Looking at the rest of the table, it's not out of place.

I do think we should also replace line 21 with what I've got since it describes how the options display. Note, there's the same part on the Input step page to replace.

So, I suggest:

- Updating the text in the table as you're doing.
- Replace line 21 in this file.
- Replace the corresponding text in the Input step file.

l-suzuki: "Is this information important for experienced users to know when they look up this field?" 
->  this PR came about because a buildkite customer raised an escalation with us because they thought the dropdown menu rendering for 2 options with `required: false` was a bug, since the behaviour wasn't captured in the docs. I can't know the answer for sure, but I think it would have probably helped to have this info in the table in this particular case.

So, I think your final suggestions are good! I'll action those today. Thanks very much for your feedback on this

mbelton-buildkite (approved):
Nice! Feel free to merge when you're ready 😊

PR #1852, opened 2023-02-06
Link to a working Codepen example for the Emojis API

dannymidnight:
This change updates a link in the REST API docs to use a revised codepen with a valid access token + organisation. 

_Note:_ The access token used has been scoped to a single organisation (buidkite-docs-examples) with no read/write permissions.


mbelton-buildkite (approved):
Thanks for fixing this!

PR #1857, opened 2023-02-10
[ONB-136] Add pipeline for keeping GraphQL docs up-to-date

dannymidnight:
This PR adds a new pipeline for auto-generating a pull request with content generated from the latest Buildkite GraphQL schema.

The [pipeline](https://buildkite.com/buildkite/docs-graphql) is currently setup to fetch the latest/active introspected GraphQL schema from buildkite.com which to start with will be triggered on a schedule. This is only temporary as I'm keen to fast follow to instead trigger the pipeline when schema changes are being deployed.

Here's an [example build](https://buildkite.com/buildkite/docs-graphql/builds/44#) and [PR](https://github.com/buildkite/docs/pull/1856) (ignore all the commits from this branch).

Keen to see how this goes. I decided to lean into the PR for visibility in changes but once we're more comfortable then perhaps merging directly into main may also be the way to go..

### DIFF
--- a/.buildkite/pipeline.graphql.yml
+++ b/.buildkite/pipeline.graphql.yml
@@ -1,0 +1,13 @@
+steps:
+  - label: Update GraphQL docs
+    command: .buildkite/update_graphql_docs
+    plugins:
+      - docker-compose#v4.9.0:
+          run: app
+          mount-ssh-agent: true
+          mount-buildkite-agent: true
+          env:
+            - API_ACCESS_TOKEN
+            - GIT_NAME
+            - GIT_EMAIL
+            - GH_TOKEN
--- a/.buildkite/update_graphql_docs
+++ b/.buildkite/update_graphql_docs
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "+++ Fetch latest schema"
+rake graphql:fetch_schema >| data/graphql_data_schema.json
+
+echo "+++ Check for schema changes"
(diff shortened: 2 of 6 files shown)


.buildkite/pipeline.graphql.yml
```diff
@@ -0,0 +1,13 @@
...
+    plugins:
+      - docker-compose#v4.9.0:
+          run: app
+          mount-ssh-agent: true
+          mount-buildkite-agent: true
+          env:
+            - API_ACCESS_TOKEN
+            - GIT_NAME
+            - GIT_EMAIL
+            - GH_TOKEN
```
mbelton-buildkite, line 13: Looks like you're not using `GH_TOKEN` anywhere?

mbelton-buildkite (approved):
I don't have access to view the build but it looks sensible :)

dannymidnight:
> I don't have access to view the build but it looks sensible :)

I'll get right on that!

.buildkite/pipeline.graphql.yml
dannymidnight: Good spot! `GH_TOKEN` is a [special env variable](https://cli.github.com/manual/gh_help_environment) `gh-cli` uses for authentication without needing to call the auth APIs.

> GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com API requests. Setting this avoids being prompted to authenticate and takes precedence over previously stored credentials.

PR #1858, opened 2023-02-10
Fix button background on "Copied" state

dannymidnight:
Minor fix for button style when displayed over content.

https://user-images.githubusercontent.com/656826/217972334-d0a512e7-28e8-431f-8dee-b671f9279830.mov


mbelton-buildkite (approved):
LGTM

PR #1873, opened 2023-02-16
Update states of cancel builds via API

pzeballos:
Now is also possible to send a cancellation signal when the build is in a `failing` state

### DIFF
--- a/pages/apis/rest_api/builds.md.erb
+++ b/pages/apis/rest_api/builds.md.erb
@@ -459,7 +459,7 @@ Error responses:
 
 ## Cancel a build
 
-Cancels the build if its state is either `scheduled` or `running`.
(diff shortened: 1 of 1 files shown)


pages/apis/rest_api/builds.md.erb
```diff
@@ -459,7 +459,7 @@ Error responses:
 
 ## Cancel a build
 
-Cancels the build if its state is either `scheduled` or `running`.
+Cancels the build if its state is either `scheduled`, `running` or `failing`.
```
mbelton-buildkite, line 462: Use a serial comma:

```suggestion
Cancels the build if its state is either `scheduled`, `running`, or `failing`.
```

mbelton-buildkite (approved):
One style nitpick to address, otherwise looks good. 😊

PR #1878, opened 2023-02-20
Drop `.erb` extension from markdown files

dannymidnight:
GitHub doesn't currently render rich diffs for `.md.erb` files since they're not treated as Markdown.

There was a previous attempt at resolving this issue by using `linguist-language` (https://github.com/buildkite/docs/pull/1511) however that setting only changes the attribution of files in the repository stats. 

This PR attempts to improve the content review process by leaning into the `.md` file extension (even though we're optionally using ruby to enrich our markdown files) which causes the magical button to appear.

Here's an example of what a "rich diff" looks like on GitHub:

https://user-images.githubusercontent.com/656826/220015016-ec85801d-b128-43d1-bffc-16147850cdc7.mov


mbelton-buildkite (approved):
🙌🏻 Nice!

It looks like this Just Works™ but I assume I'm missing something. How does this continue supporting the Ruby elements in pages?

dannymidnight:
> It looks like this Just Works™ but I assume I'm missing something. How does this continue supporting the Ruby elements in pages?

Yep, it should all continue to work as expected since we're explicitly controlling the ruby rendering as opposed to letting rails handle it.
https://github.com/buildkite/docs/blob/da6e937c3ddb03fb2525ed04cc072ffba732d50a/app/models/page.rb#L111-L116

The main thing here is that we're breaking a common convention for using the ".erb" extension for files that include ERB. I'm OK with that since it will improve our content review process.

PR #1881, opened 2023-02-21
Custom status caveat

mcncl:
Custom status updates can only be used with builds from a webhook, when the _Update commit statuses_ box is unchecked.

### DIFF
--- a/pages/integrations/github.md
+++ b/pages/integrations/github.md
@@ -121,7 +121,7 @@ You can customize the commit statuses, for example to reuse the same pipeline fo
     <%= image "github-custom-status.png", alt: "Screenshot of GitHub build settings and the resulting GitHub pull request statuses" %>
 
 >🚧 Build source limitations when using custom statuses
-> In order for Buildkite to deliver custom status updates the build source must be "Webhook", builds from other sources will not allow Buildkite to identify the destination for the custom status updates.
+> In order for Buildkite to deliver custom status updates, the build source must be "Webhook"; builds from other sources will not allow Buildkite to identify the destination for the custom status updates, so deselecting _Update commit statuses_ will result in the failure to update custom commit statuses.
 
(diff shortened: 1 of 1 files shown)


mbelton-buildkite:
@mcncl, I rebased your branch to pick up changes yesterday that caused conflicts with existing branches.

pages/integrations/github.md
```diff
@@ -120,6 +120,9 @@ You can customize the commit statuses, for example to reuse the same pipeline fo
 3. When you make a new commit or pull request, you should see _my-custom-status_ as the commit status:
     <%= image "github-custom-status.png", alt: "Screenshot of GitHub build settings and the resulting GitHub pull request statuses" %>
 
+>🚧 Build source limitations when using custom statuses
```
mbelton-buildkite, line 123: Is this specific to custom statuses defined at the build level? Is it also relevant to step level definitions?

pages/integrations/github.md
```diff
@@ -120,6 +120,9 @@ You can customize the commit statuses, for example to reuse the same pipeline fo
...
+> In order for Buildkite to deliver custom status updates, the build source must be "Webhook"; builds from other sources will not allow Buildkite to identify the destination for the custom status updates, so deselecting _Update commit statuses_ will result in the failure to update custom commit statuses.
```
mbelton-buildkite, line 124: > so deselecting _Update commit statuses_ will result in the failure to update custom commit statuses.

Is this already covered above:

> In Pipeline Settings > GitHub, make sure Update commit statuses is not selected. Note that this prevents Buildkite from automatically creating and sending statuses for this pipeline, meaning you will have to handle all commit statuses through the pipeline.yml.

Or is that different?

mbelton-buildkite, line 124: How does someone verify that the build source is `webhook`?

mbelton-buildkite (commented):
Clarifying a few things to start :)

pzeballos:
There's no need to uncheck the "Update commit statuses"  box, it will create duplicate status but it will work. And it is correct that it only works for builds triggered by a webhook

pages/integrations/github.md
pzeballos: it's the same for build and step-level definitions

PR #1866, opened 2023-02-13
PIE-1421: Add Jasmine & Mocha instructions to JS collector

KatieWright26:
Our [`test-collector-javascript`](https://github.com/buildkite/test-collector-javascript) has had support for Jasmine and Mocha test suites since August 2022, however the [docs](https://buildkite.com/docs/test-analytics/javascript-collectors) hadn't been updated with the relevant set up instructions. 
I have updated the docs to match the current set up instructions in the [javascript collector](https://github.com/buildkite/test-collector-javascript) repo in the hope of attracting users to take advantage of our collector. 

I did need to tweak the doc a bit to get the information flowing, so feedback very, very welcome.

PR #1882, opened 2023-02-21
Document support for Jasmine and Mocha in JS Collector

mbelton-buildkite:
Continued from #1866


KatieWright26 (approved):
Looks great; also checked out the render.io preview and it's all looking as expected :) Thank you for finishing this off!

PR #1883, opened 2023-02-21
Add signal and signal reason in retry rules

sj26:
Add reference documentation for the signal and signal reason attributes which will soon be available in automatic retry rules.

Fixes PDP-211.

### DIFF
--- a/pages/pipelines/command_step.md
+++ b/pages/pipelines/command_step.md
@@ -300,7 +300,7 @@ _Optional Attributes_
   <tr>
     <td><code>exit_status</code></td>
     <td>
-      The exit status number that will cause this job to retry ('*' does not include 0) <br>
+      The exit status number that causes this job to retry ('*' does not include 0) <br>
       <p><em>Examples:</em></p>
       <ul>
         <li><code>"*"</code></li>
@@ -311,7 +311,7 @@ _Optional Attributes_
   <tr>
     <td><code>signal</code></td>
     <td>
-      The signal that will cause this job to retry. This signal will only be present if the agent sends a signal to the job and an interior process does not handle the signal. `SIGKILL` propagates reliably because it cannot be handled, and is a useful to differentiate graceful cancelation and timeouts.
+      The signal that causes this job to retry. This signal will only be present if the agent sends a signal to the job and an interior process does not handle the signal. `SIGKILL` propagates reliably because it cannot be handled, and is a useful to differentiate graceful cancelation and timeouts.
       <p><em>Examples:</em></p>
       <ul>
         <li><code>"*"</code></li>
--- a/pages/pipelines/command_step.md
+++ b/pages/pipelines/command_step.md
@@ -301,26 +301,35 @@ _Optional Attributes_
     <td><code>exit_status</code></td>
     <td>
       The exit status number that will cause this job to retry ('*' does not include 0) <br>
-      <em>Example:</em> <code>"*"</code><br>
-      <em>Example:</em> <code>2</code>
+      <p><em>Examples:</em></p>
+      <ul>
+        <li><code>"*"</code></li>
+        <li><code>2</code></li>
+      </ul>
     </td>
   </tr>
   <tr>
     <td><code>signal</code></td>
     <td>
       The signal causing the job to exit, if a signal was sent, and the currently running process did not intercept and handle the signal supplying a non-default exit status.<br>
-      <em>Example:</em> <code>"*"</code><br>
-      <em>Example:</em> <code>kill</code><br>
-      <em>Example:</em> <code>SIGINT</code>
+      <p><em>Examples:</em></p>
+      <ul>
+        <li><code>"*"</code></li>
+        <li><code>kill</code></li>
+        <li><code>SIGINT</code></li>
+      </ul>
     </td>
   </tr>
   <tr>
     <td><code>signal_reason</code></td>
     <td>
       The reason a process was signaled.<br>
-      <em>Example:</em> <code>"*"</code><br>
-      <em>Example:</em> <code>none</code><br>
-      <em>Example:</em> <code>cancel</code>
+      <p><em>Examples:</em></p>
+      <ul>
+        <li><code>"*"</code></li>
+        <li><code>none</code></li>
+        <li><code>cancel</code></li>
+      </ul>
     </td>
   </tr>
   <tr>
--- a/pages/pipelines/command_step.md
+++ b/pages/pipelines/command_step.md
@@ -311,7 +311,7 @@ _Optional Attributes_
   <tr>
     <td><code>signal</code></td>
     <td>
-      The signal causing the job to exit, if a signal was sent, and the currently running process did not intercept and handle the signal supplying a non-default exit status.<br>
+      The signal that will cause this job to retry. This signal will only be present if the agent sends a signal to the job and an interior process does not handle the signal. `SIGKILL` propagates reliably because it cannot be handled, and is a useful to differentiate graceful cancelation and timeouts.
       <p><em>Examples:</em></p>
       <ul>
         <li><code>"*"</code></li>
--- a/pages/pipelines/command_step.md
+++ b/pages/pipelines/command_step.md
@@ -308,7 +308,7 @@ _Optional Attributes_
   <tr>
     <td><code>signal</code></td>
     <td>
-      The signal sent to a process, if the process did not intercept and handle the signal<br>
+      The signal sent to a process, if the process did not intercept and handle the signal.<br>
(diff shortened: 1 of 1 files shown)


pages/pipelines/command_step.md
```diff
@@ -306,6 +306,24 @@ _Optional Attributes_
...
+      The signal sent to a process, if the process did not intercept and handle the signal<br>
+      <em>Example:</em> <code>"*"</code><br>
+      <em>Example:</em> <code>kill</code><br>
+      <em>Example:</em> <code>SIGINT</code>
+    </td>
+  </tr>
+  <tr>
+    <td><code>signal_reason</code></td>
+    <td>
+      The reason a process was signalled<br>
```
mbelton-buildkite, line 321: Style changes:

- End the sentence with a period.
- Use US English

```suggestion
      The reason a process was signaled.<br>
```

mbelton-buildkite: Would you mind adding a period to line 304 in this change, too?

pages/pipelines/command_step.md
```diff
@@ -306,6 +306,24 @@ _Optional Attributes_
       <em>Example:</em> <code>2</code>
     </td>
   </tr>
+  <tr>
+    <td><code>signal</code></td>
+    <td>
+      The signal sent to a process, if the process did not intercept and handle the signal<br>
```
mbelton-buildkite, line 312: This sentence is a bit unclear to me. Could you please unpack it a bit more?

> The signal sent to a process

Looking at [Signal handing in the agent](https://buildkite.com/docs/agent/v3#signal-handling), this value looks like it's the name of the signal? Is "a process" here also the build job process?

> if the process did not intercept and handle the signal

This was the confusing bit for me. The previous part of the sentence sounds like the process did receive the signal. This part sounds like a caveat to say that the previous part won't happen if the process intercepts and handles the signal. But, the caveat still involves the process receiving the signal. That seems like a contradiction.

I'm probably missing something simple 🤔

pages/pipelines/command_step.md
```diff
@@ -306,6 +306,24 @@ _Optional Attributes_
...
+      <em>Example:</em> <code>"*"</code><br>
+      <em>Example:</em> <code>kill</code><br>
+      <em>Example:</em> <code>SIGINT</code>
```
mbelton-buildkite, line 315: What do you think of listing out the example values rather than repeating? Thinking:

```suggestion
      <p><em>Examples:</em>
      <ul>
        <li><code>"*"</code></li>
        <li><code>kill</code></li>
        <li><code>SIGINT</code></li>
      </ul></p>
```

mbelton-buildkite (commented):
Some style nitpicks and clarifications :)

pages/pipelines/command_step.md
sj26: It's .. complex 😅

It's the signal sent by the agent to the bootstrap, and then likely forwarded from the bootstrap to the command process, and only appears here iff the signal was not handled by one of those interior processes and so was handled by the default linux exit handler and turned into an exit code based on the signal. I think.

The most reliable usage is for SIGKILL because there cannot be a handler for that signal, and so it propagates reliably. It's also arguably the most useful signal for rules in here, because it allows differentiating graceful cancellation from cancellation which times out, for example.

I would love better word suggestions :pray:

pages/pipelines/command_step.md
sj26: Sure! I was following the pattern on the page.

pages/pipelines/command_step.md
sj26: Done.

pages/pipelines/command_step.md
sj26: I've tried another iteration on the words.

pages/pipelines/command_step.md
mbelton-buildkite: Yeah, that would be awesome 🙌🏻

pages/pipelines/command_step.md
```diff
@@ -306,6 +306,24 @@ _Optional Attributes_
       <em>Example:</em> <code>2</code>
     </td>
   </tr>
+  <tr>
+    <td><code>signal</code></td>
+    <td>
+      The signal causing the job to exit, if a signal was sent, and the currently running process did not intercept and handle the signal supplying a non-default exit status.<br>
```
mbelton-buildkite, line 312: Okay, here's an attempt where I'm trying to strike a balance between everything we can tell them and what's helpful:

>The signal that causes the job to retry. The command process only receives a signal when it propagates from the agent and is not handled by another internal process like the bootstrap. We recommend using this for `SIGKILL` because it propagates reliably and differentiates between graceful cancelation and timing out.

What do you think? Does it make sense? The first sentence is the most important part. The rest is additional context that might help someone.

mbelton-buildkite (commented):
Maybe getting closer?

pages/pipelines/command_step.md
sj26: I'd riffed on your words in the theme of front-loading more useful information. What do you think?

pages/pipelines/command_step.md
```diff
@@ -302,8 +302,35 @@ _Optional Attributes_
...
+      <ul>
+        <li><code>"*"</code></li>
+        <li><code>2</code></li>
+      </ul>
+    </td>
+  </tr>
+  <tr>
+    <td><code>signal</code></td>
+    <td>
+      The signal that will cause this job to retry. This signal will only be present if the agent sends a signal to the job and an interior process does not handle the signal. `SIGKILL` propagates reliably because it cannot be handled, and is a useful to differentiate graceful cancelation and timeouts.
```
mbelton-buildkite, line 315: Typo + formatting:

```suggestion
      The signal that will cause this job to retry. This signal will only be present if the agent sends a signal to the job and an interior process does not handle the signal. <code>SIGKILL</code> propagates reliably because it cannot be handled, and is useful to differentiate graceful cancelation and timeouts.
```

Or, with my nitpicks:

- Avoid "will"
- Being "present" sounds a bit unclear to me, but I'm not sure if what I've got is right.

```suggestion
      The signal that causes this job to retry. This signal only occurs when the agent sends a signal to the job and an interior process does not handle the signal. <code>SIGKILL</code> propagates reliably because it cannot be handled, and is useful to differentiate graceful cancelation and timeouts.
```

mbelton-buildkite (commented):
I like what you've got! One small typo and a formatting error, and of course a couple more nitpicks you can consider.

mbelton-buildkite:
FYI, you'll need to pick up changes on `main` to fix the build :)

pages/pipelines/command_step.md
sj26: Nice. I'm used to ["present" in the rails sense](https://api.rubyonrails.org/classes/Object.html#method-i-present-3F), but that's probably not widely accessible here. "occurs" sounds like the signal might only be _sent_ in limited cases, which isn't quite right. I've had a go with "appears".

sj26:
Rebased on main, too

pages/pipelines/command_step.md
```diff
@@ -300,9 +300,36 @@ _Optional Attributes_
...
+      <ul>
+        <li><code>"*"</code></li>
+        <li><code>2</code></li>
+      </ul>
+    </td>
+  </tr>
+  <tr>
+    <td><code>signal</code></td>
+    <td>
+      The signal that causes this job to retry. This signal only appears if the agent sends a signal to the job and an interior process does not handle the signal. <code>SIGKILL</code> propagates reliably because it cannot be handled, and is a useful to differentiate graceful cancelation and timeouts.
```
mbelton-buildkite, line 314: Oops, I think this was missed in the last update. I assume it's either meant to be:

```suggestion
      The signal that causes this job to retry. This signal only appears if the agent sends a signal to the job and an interior process does not handle the signal. <code>SIGKILL</code> propagates reliably because it cannot be handled, and is a useful way to differentiate graceful cancelation and timeouts.
```

Or:

```suggestion
      The signal that causes this job to retry. This signal only appears if the agent sends a signal to the job and an interior process does not handle the signal. <code>SIGKILL</code> propagates reliably because it cannot be handled, and is useful to differentiate graceful cancelation and timeouts.
```

mbelton-buildkite (commented):
One last thing! It's so close

sj26:
One day I'll learn to proof read 😅

Thanks @mbelton-buildkite

mbelton-buildkite (approved):
Looks good to me! Feel free to merge when the builds pass and you're ready :)

sj26:
Thanks @dabarrell :pray:

PR #1885, opened 2023-02-22
Storybook to get a stocktake of all the Docs styles

olyism:
Spinning up [Storybook](https://storybook.js.org/) to do a stocktake of all the Docs components.

At the moment it's hard to see what Docs styles and components are in one view. Storybook makes this easy to explore what we have and preview different states. This also makes it easy for me to update Docs' look and feel when the design library changes.

Although this can be done in Figma, design can still appear different to code and it introduces double-handling to keep both environments up to date. It's also a lot faster to explore different states and interactions in Storybook.

To start, I've set up some rudimentary content:

- Basic README
- Colors
- Typography (headings and paragraph only)
- Buttons
- Tables
- Figma embed

The main UI problem I want to start looking at is tables. After doing the stocktake I can see we have five types and the writing experience is cumbersome and confusing using a combination of markdown, HTML and css classes.

Screen Shot 2023-02-22 at 12 01 08 PM


dannymidnight:
> Currently to use it you'd have to spin it up locally yourself. @dannymidnight do you think there might be an easy and good place to host this publics? 🤔

I reckon local is totally fine since it's hard to say if this storybook would be short-lived if for example we eventually started using a shared design system.

stories/Colors.stories.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+# Colors
+
+## Global
+
+<ColorPalette>
+  <ColorItem
+    title="Base"
+    colors={{
+      'base-0': '#ffffff',
```
dannymidnight, line 13: Should these colours be pulled from common file? It seems likely they'll drift out from reality which would make this page not so useful. 

Maybe there's a way to import JSON to scss?

stories/Typography.stories.mdx
```diff
@@ -0,0 +1,39 @@
+import { Meta, Typeset } from '@storybook/addon-docs';
+
+<Meta title="Base/Typography" />
+
+export const fontSizeBase = 16;
+export const fontFamilySystem = '-apple-system, BlinkMacSystemFont, Segoe UI, Roboto, "Helvetica Neue", Helvetica, sans-serif';
+export const fontFamilyDisplay = `"Aeonik", ${fontFamilySystem}`;
+export const fontFamilyBody = `"Maison Neue", ${fontFamilySystem}`;
```
dannymidnight, line 8: Same goes for all of this – ideally storybook is rendering using shared configuration.

mbelton-buildkite:
@olyism, is this something you're still pursuing? Or can we close it?

olyism:
hi @mbelton-buildkite let's close this for now. I'll add it for the next Docs week.

PR #1888, opened 2023-02-23
🐛 Reduce filter saturation to fix Safari color

olyism:
Fix a subtle visual bug that only occurs in Safari. When the nav's chev icon rotates during a css transition its color turns pink briefly.

After some investigation I think it's due to the `saturation` value in the css filter.

I've fixed this for now by reducing the `saturation` value and adjusted other values to mimic the purple.

Note: this is a patch as using css filters to change color still has discrepancy. In a future ticket it would be good to use a more elegant solution like inline SVG.

### DIFF
--- a/app/assets/stylesheets/components/_nav.scss
+++ b/app/assets/stylesheets/components/_nav.scss
@@ -1,5 +1,5 @@
 @mixin _make-nav-icon-purple {
-  filter: invert(24%) sepia(92%) saturate(7443%) hue-rotate(258deg) brightness(81%) contrast(109%);
+  filter: invert(15%) sepia(80%) saturate(550%) hue-rotate(210deg) brightness(80%) contrast(110%);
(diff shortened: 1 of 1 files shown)


olyism:
# Screenshots

## Before (Safari)

Screen Shot 2023-02-23 at 12 46 19 PM (2)

## After 

### Safari

Screen Shot 2023-02-23 at 12 47 49 PM

### Chrome

Screen Shot 2023-02-23 at 12 48 18 PM

clbarrell:
CHAMPION

olyism:
Several small fixes have surfaced, so I think I'll close this PR in a sec and open a new one with that clusters all the small fixes.

olyism:
Superseded by https://github.com/buildkite/docs/pull/1889 because I'm grouping all the small fixes together.

PR #1889, opened 2023-02-23
💄 Style makeover post release bug fixes

olyism:
- [x] Load the correct font file
- [x] Fix issue of webfont not loading
  - [x] Use `url` not `asset-url` for externally hosted fonts
  - [x] `preload` fonts to prevent FOUT
  - [x] Removed Circular from `_head.html.erb`
- [x] SVG diagram missing the new web font
- [x] Content fix: beta flags for test collector
- [x] Strange pink Safari toggle icon
- [x] TOC highlight fix: click event no longer overwritten by the scroll event
- [x] Consistent, slightly reduced table font sizes


olyism:
# Screenshots for the pink Safari icon fix

## Before (Safari) 🐽

Screen Shot 2023-02-23 at 12 46 19 PM (2)

## After ✅

### Safari 🟣

Screen Shot 2023-02-23 at 12 47 49 PM

### Chrome 🟣

Screen Shot 2023-02-23 at 12 48 18 PM

### Firefox 🟣

Screen Shot 2023-02-23 at 12 51 20 PM

olyism:
Web fonts for the SVG diagram has been fixed. 

Screen Shot 2023-02-23 at 2 45 44 PM

It's the only instance in the entire Docs site where an SVG diagram uses text.

olyism:
Added beta tags for beta collectors.

Screen Shot 2023-02-23 at 2 47 10 PM

Screen Shot 2023-02-23 at 2 47 26 PM

olyism:
Webfonts now loading correctly (as seen on Render).

Screen Shot 2023-02-24 at 7 46 32 AM

olyism:
hi @dannymidnight you're right, the font file isn't correct. I've downloaded the files from the Marketing Site and will need some help uploading them again. Thanks 🙏

BrettJay:
This might be a silly question, but I know with Rails and Next.js projects it’s possible to serve assets from a `/public` directory which ensures a stable URL without asset hashing for cache busting...

e.g. `buildkite.com/_site/fonts/aeonik.woff2`

(That doesn’t exist, an actual example of a stable public url asset hosted from site is `https://buildkite.com/_site/email/buildkite.png`)

Would that be suitable for webfonts in docs? My suspicion is that the URL we’re sourcing the web fonts here isn’t intended to be stable and so Docs would need to know when the URL changes.

app/views/application/_head.html.erb
```diff
@@ -1,9 +1,11 @@
 <meta charset='utf-8'>
 <%= tag :link, rel: "dns-prefetch", href: ENV["ASSET_HOST"] if ENV["ASSET_HOST"] %>
 <link rel="dns-prefetch" href="//www2.buildkiteassets.com" />
-<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/lineto-circular/lineto-circular-book-39c5d1ed54e49102939d0280aeb20f01ef021bf5ffa74dc25fcafb43fce62ff3.woff2" crossorigin="anonymous" />
-<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/lineto-circular/lineto-circular-bold-23e71bdfef13622d0d52b2b4b7ed3c1edb9e81f210692130dee9a521e97d062f.woff2" crossorigin="anonymous" />
-<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/lineto-circular/lineto-circular-black-a8002849ea799a4dcf1be0b1abe0f010991cbae508f24f62e9ee0262590197eb.woff2" crossorigin="anonymous" >
+<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/aeonik/Aeonik-Bold-78fff47500f27f2f0d8d777c9d431a87036e566af7577a58ccc0776b4eb4b994.woff2" crossorigin="anonymous" />
```
dannymidnight, line 4: Nice 👍

app/views/application/_head.html.erb
```diff
@@ -1,9 +1,11 @@
...
+<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/maison-neue/MaisonNeue-Book-5e1404c0c8bc64daaae66d3a37fa7b0c7ffe65340d181e9068741deff8669cda.woff2" crossorigin="anonymous" />
+<link rel="preload" as="font" type="font/woff2" href="https://www2.buildkiteassets.com/assets/maison-neue/MaisonNeue-BookItalic-2a5e47b91af0050b44c21f9038ef63a37e228fd53b0289dfccbcdd3aa44ba19c.woff2" crossorigin="anonymous" />
```
dannymidnight, line 6: I think we should drop preloading these until we've landed the revised fonts. https://github.com/buildkite/buildkite/pull/11319

olyism:
hi @BrettJay not a silly question at all. I struggled with serving assets from Docs' `public` folder and, to the best of my knowledge, it may have something to do with sprockets. I could load fonts locally on my local dev environment, but it was throwing 500 errors on Render 🤷‍♂️ 

In light of this, we went with the existing font approach — i.e. hosting on our buckets. This isn't ideal, it works for now, but we'd be keen to explore better solutions in the future.

I'm thinking the ideal state would be Docs loading assets from the same source as the Marketing Site in the future.

app/assets/javascripts/toc.js
```diff
@@ -42,6 +38,14 @@
           setCurrentLink(`#${heading.id}`);
         }
       });
+    };
+
+    window.addEventListener("scroll", handleScroll);
+
+    tocLinkNodes.forEach((link) => {
+      link.addEventListener("click", (e) => {
+        setTimeout(() => setCurrentLink(e.target.hash), 25);
```
dannymidnight, line 47: What's the thinking behind adding the timeout here?

app/assets/stylesheets/base/_text_content.scss
```diff
@@ -9,8 +9,8 @@
 
 @mixin body {
   color: map-get($color-aliases, 'text-body');
-  font-size: 1.1875rem;
-  line-height: 1.618;
+  font-size: $font-size-body;
+  line-height: map-get($line-heights, 'golden');
```
dannymidnight, line 13: ✨

app/assets/stylesheets/components/_nav.scss
```diff
@@ -228,12 +218,12 @@
 
   &__toggle {
     cursor: pointer;
-    font-weight: 600;
     width: 100%;
 
     &::before {
       background: transparent asset-url("icons/angle-down.svg") 50% 50% no-repeat;
       transform: rotate(-90deg);
+      will-change: transform;
```
dannymidnight, line 226: 👍

app/views/application/_head.html.erb
dannymidnight: You should be able to drop the font changes after a rebase.

dannymidnight (approved):
Nice one, looks like some good adjustments.

app/assets/javascripts/toc.js
olyism: So that it highlights link that you've clicked as the active TOC item, rather than the heading in the viewport.

What's happening is both the scroll position of heading and the clicked TOC item are competing, and the scroll event always trump.

I tried a bunch of different ways and this was the best I could come up with to ensure the click comes after the scroll position with setTimeout.

app/views/application/_head.html.erb
olyism: reverting back to Maison Bold, rather than Medium, because for Docs' use case the Medium font weight is too light. It makes sense for the visual design for Docs to diverge from Marketing Site here.

olyism:
hi @dannymidnight I've rebased the latest changes from `main`. Would be keen to get your eyes on it one more time to ensure before merging:

1. **Font:** loading Maison Bold, instead of Medium, because the proper bold contrasts better on Docs. We need bolding more than Marketing so it makes sense to diverge here
2. SVG diagram now loading Maison, and not Circular
3. Beta flags for beta test collectors
4. Fixed TOC highlight with setTimeout (be keen to hear if there's a more elegant solution)
5. Consistent table font size

Screen Shot 2023-03-06 at 5 11 12 PM

PR #1890, opened 2023-02-23
Add section describing changing the agent service user on Windows

HugeIRL:
This PR adds a subsection to the docs under "Running as a service" describing how to use nssm via the command-line to change the agent service user the agent is running as.


pages/agent/v3/windows.md
```diff
@@ -89,6 +89,15 @@ nssm start buildkite-agent
 nssm status buildkite-agent
 ```
 
+If you'd like to change the user the buildkite-agent service runs as, you can use the same third-party tool [nssm](https://nssm.cc/) using the command-line:
+
+```
+# Change ComputerName to the name of the System under "Settings"
+# Change AccountName to the name of the Account you'd like to use
+# Change password to the password of the account you provided in the above comment
```
mbelton-buildkite, line 97: "The above comment" is a bit ambiguous. Perhaps:

```suggestion
# Change password to the password of the account you'd like to use
```

What do you think?

pages/agent/v3/windows.md
```diff
@@ -89,6 +89,15 @@ nssm start buildkite-agent
...
+nssm set buildkite-agent ObjectName "ComputerName\AccountName" "password"
```
mbelton-buildkite, line 98: What's the `ObjectName`? It looks like it's not mentioned anywhere here. Is it meant to be replaced by another value or left as is?

pages/agent/v3/windows.md
HugeIRL: 100%, great change

HugeIRL: But, keep in mind the password has to be the password already determined for the account the user is wanting to use. So if the Admin account is PC\Administrator with a password of "notarealpassword", the password field in the CLI command has to be "notarealpassword" or some kind of variable/env var to match the account password. It's not meant to be something to set the password itself. So we might wanna reconsider the wording here again.

pages/agent/v3/windows.md
HugeIRL: So in this case, `ObjectName` is part of the CLI syntax and not to be touched.

pages/agent/v3/windows.md
mbelton-buildkite: I suspect the formatting is working against us here. I'm going to suggest a different style.

pages/agent/v3/windows.md
```diff
@@ -89,6 +89,15 @@ nssm start buildkite-agent
...
+```
```
mbelton-buildkite, line 99: Alternative style:

````suggestion
```
nssm set buildkite-agent ObjectName "COMPUTER_NAME\ACCOUNT_NAME" "PASSWORD"
```

Replace the following:

- `COMPUTER_NAME`: The system name under _Settings_. For example, `PC`.
- `ACCOUNT_NAME`: The name of the account you'd like to use. For example, `Administrator`.
- `PASSWORD`: The password for the account you'd like to use. You can reference a variable rather than directly specifying the value.
````

This:

- Makes the command able to be copied without the comments getting in the way.
- Separates placeholders from keywords following the [Google style guide](https://developers.google.com/style/code-syntax).
- Removes ambiguity about "changing" things.

What do you think?

HugeIRL: Much nicer! I'm all for this change.

mbelton-buildkite (approved):
Looks good to me! And thanks for adding this info. Feel free to merge when you're ready :)

PR #1893, opened 2023-02-24
Move table of contents out of main content

dannymidnight:
This PR updates the way table of contents are configured and generated and moves the nav outside the main content, fixing a text selection issue.

Visually this change moves the table of contents above the page title on smaller devices – which I think is better but keen for others opinions.

### Changes
1. Generate table of contents from markdown parser (data extractor). This makes it easy to change where the table of contents are rendered.
2. Use front matter for opting-out of rendering table of contents. Front matter is YAML metadata included in the top of a markdown template which is commonly seen in static site generators. 
3. Remove all usage of `:notoc` and `:toc`
4. Render the template in another column – some styling adjustments are required to make this work as expected.
5. Remove custom `{ :id}` syntax for overriding header IDs (only used in a single template)

| Before | After |
| --- | --- |
| Screenshot 2023-02-27 at 12 49 38 pm | Screenshot 2023-02-27 at 12 49 40 pm
| Screenshot 2023-02-27 at 12 49 24 pm | Screenshot 2023-02-27 at 12 49 28 pm


spec/scripts/graphql_api_content/render_helpers_spec.rb
```diff
@@ -703,9 +708,7 @@
             <h1 class="has-pills" data-algolia-exclude>__DirectiveLocation<span class="pill pill--object pill--normal-case pill--large"><code>OBJECT</code></span></h1>
             <!-- vale on -->
```
olyism, line 709: cool work with frontmatter 🤩 

In this case we can remove the `<!-- vale on/off -->` comments in the GraphQL API render script

app/views/layouts/application.html.erb
```diff
@@ -9,14 +9,11 @@
...
         <div class="Docs__page-container__inner PageContainer">
           <div class="Docs__article">
-            <div class="column1">
-              <div class="TextContent">
-                <%= yield %>
-              </div>
+            <%= yield :toc %>
+            <div class="TextContent">
+              <%= yield %>
               <%= render "footer", :embed_emojicom => true %>
```
olyism, line 15: Let's move the footer out of `.TextContent`. It's breaking the styles a bit, e.g. border around the emojicom iframe, changing he button styles.

dannymidnight: Good catch! 

FYI I've got a bit of another follow up CSS fix that does some larger style changes that I reckon will make this all work a bit better using CSS grids.

pages/test_analytics.md
```diff
@@ -1,3 +1,7 @@
+---
+toc: true
```
olyism, line 2: Test Analytics is a landing page so this should be false.

```suggestion
toc: false
```

spec/scripts/graphql_api_content/render_helpers_spec.rb
dannymidnight: Had a quick look at doing this but it appears we're still making use of this to get around some vale violations from

pages/test_analytics.md
olyism: Also, landing pages are an interesting one and I think a good candidate to start using frontmatter with.

Perhaps in a separate ticket we could have something like a `page_layout` attribute with values like `default|landing_page`. 

Landing pages are currently using the `landing_page.html.erb` view layout, which could possibly be refactored to just one file with conditionals in the template. We could also get rid of the `landing_pages.rb` model as well.

app/assets/stylesheets/base/_text_content.scss
```diff
@@ -1,10 +1,5 @@
 @mixin margin($margin: 1) {
   margin-bottom: #{$margin}rem;
-  margin-top: #{$margin}rem;
-
-  &:first-child {
-    margin-top: 0;
-  }
 }
 
 @mixin body {
```
olyism, line 5: hi @dannymidnight here's a suggestion for margins. I think it's beneficial to have top and bottom margins for `h1–h6` and `p` so they'll always create a gap above them for any element (such as `table`, custom `div`s etc).

```suggestion
@mixin margin($margin: 1) {
  margin-bottom: #{$margin}rem;
  margin-top: #{$margin}rem;
}

@mixin body($margin: 1) {
```

dannymidnight: Makes sense! I've pushed up something similar that brings back the top margins and simplifies usage of the sass mixins. Hopefully this does the trick :)

https://github.com/buildkite/docs/pull/1893/commits/f8a69c2c3a296a934e3bebf8969e2d7f4792f0c0

olyism (approved):
LGTM 🚀  

Awesome work! Really excited with the frontmatter addition 🤩 

@mbelton-buildkite do you have any thoughts on this PR particularly around how this changes the writing experience?

PR #1898, opened 2023-02-26
ONB-172: Add pipelines glossary

mbelton-buildkite:
Adding a glossary of terms that introduce key concepts in Pipelines.

**Preview:** https://bk-docs-pr-1898.onrender.com/docs/pipelines/glossary


pages/pipelines/glossary.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Agent
+
+An agent is a small, reliable, and cross-platform build runner that connects your infrastructure to Buildkite. It polls Buildkite for work, runs jobs, and reports results. You can install agents on local machines, cloud servers, or other remote machines. You need at least one agent to run builds.
+
+To learn more, see the [Agent overview](/docs/agent/v3).
+
+## Artifact
+
+An artifact is a file or metadata generated during a build. You can keep artifacts in a Buildkite-managed storage service or a third-party cloud storage service like Amazon S3, Google Cloud Storage, or Artifactory. Common uses include storing assets like logs and reports, or passing files between steps.
```
sj26, line 15: "metadata" is a separate thing. While artifacts might contain meta information like timing or reports about a build or job, I think it might be good to avoid the term.

pages/pipelines/glossary.md
```diff
@@ -0,0 +1,72 @@
...
+
+A hook is a method of customizing the behavior of Buildkite through lifecycle events. They let you run scripts at different points of the agent or job lifecycle. Using hooks, you can extend the functionality of Buildkite and automate tasks specific to your workflow and requirements.
+
+To learn more, see [Hooks](/docs/agent/v3/hooks).
+
+## Job
+
+A job is the execution of a command step during a build. Jobs run the commands, scripts, or plugins defined in the step.
+
+## Job queue
```
sj26, line 41: I think we just call them "queues" most of the time. Is this enough? Is the "job" prefix necessary for disambiguation?

```
## Queue

Queues contain jobs waiting to run on an agent.
```

pages/pipelines/glossary.md
```diff
@@ -0,0 +1,72 @@
...
+
+The job queue is a queue of jobs waiting to run on an agent. You can configure queues for particular groups of agents, isolating a set of jobs and agents. Typical uses for queues include separating deployment agents and pools of agents for specific pipelines or teams.
+
+To learn more, see [Queues](/docs/agent/v3/queues).
+
+## Pipeline
+
+A pipeline is a container for modeling and defining workflows. They contain a series of steps to achieve goals like building, testing, and deploying software.
+
+To learn more, see the [Pipeline overview](/docs/pipelines).
```
sj26, line 51: This is so tricky. Pipeline means so many things. But I think you've walked the line nicely here.

pages/pipelines/glossary.md
```diff
@@ -0,0 +1,72 @@
...
+To learn more, see [Plugins](/docs/plugins).
+
+## Step
+
+A step describes a single, self-contained task as part of a pipeline. You define a step in the pipeline configuration using one of the following types:
+
+- **Command step:** Runs one or more shell commands on one or more agents.
+- **Block step:** Pauses a build until unblocked.
+- **Wait step:** Pauses a build until all previous jobs have completed.
+- **Input step:** Pauses a build while collecting information from a user.
```
sj26, line 68: Close, but I think "pauses" might not be true here. The difference between "block" and "input" steps is that an input step _doesn't_ pause the pipeline by default, only steps which directly depend on it. Both can have fields which collect information.

pages/pipelines/glossary.md
mbelton-buildkite: 👍🏻 I'll remove "or metadata"

pages/pipelines/glossary.md
mbelton-buildkite: I like it. I was nervous about whether we use other types of queue, but I don't think I've come across anything.

pages/pipelines/glossary.md
mbelton-buildkite: Good catch! Thanks. I think I'm focussing on the wrong bit here, which causes the issue. The most important part of an input step is that it collects information from a user. I'll change it to that and skip the details about what it pauses on this page. Let me know what you think.

sj26 (approved):
Reads well to me, nice one :+1:

You might like another set of eyes to confirm

mbelton-buildkite:
Thanks @sj26! Is there anyone in particular you'd recommend to review it?

pages/pipelines/glossary.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Artifact
+
+An artifact is a file generated during a build. You can keep artifacts in a Buildkite-managed storage service or a third-party cloud storage service like Amazon S3, Google Cloud Storage, or Artifactory. Common uses include storing assets like logs and reports, or passing files between steps.
+
+To learn more, see [Build artifacts](/docs/pipelines/artifacts).
+
+## Build
+
+A build is a single run of a pipeline. You can trigger a build in various ways, including through the UI, API, as the result of a webhook, or on a schedule.
```
eleanorakh, line 21: @mbelton-buildkite you can also use a trigger step to create a build: https://buildkite.com/docs/pipelines/trigger-step

pages/pipelines/glossary.md
eleanorakh: Agreed, this is very well done. ✨

eleanorakh (approved):
One non-blocking comment, otherwise I think this is wonderful!

pages/pipelines/glossary.md
mbelton-buildkite: I like the way that hints at the complexity you can achieve with Buildkite. I'll include it, thanks! 😊

PR #1900, opened 2023-02-27
Fix Rspec link to instance_of

amybiyuliu:
Relishapp has historically been an official source of ruby documentation. However, the maintainer was recently made redundant from his job and the website is currently returning a 503. The ruby community will probably either fix the website or find a new site to host this content; however, in the mean time we're linking to the Internet Archive as a stop gap.


mbelton-buildkite (approved):
Thanks for fixing this! Feel free to merge when you're ready :)

PR #1904, opened 2023-02-28
[ONB-195] Ruby v3.0.5

dannymidnight:
Ruby 2.7 will be [EOL](https://endoflife.date/ruby) next month (31 March 2023) 

This change updates docs to ruby v3.0.5.


yob:
I don't know docs well enough to give this a good review, but a green build looks promising!

After this ships, @ellsclytn would be suuuper grateful (for SOC2 CVE auditing reasons) if you could configure dependabot like this: https://github.com/buildkite/buildkite/blob/bdc9e38356166dc258d2b02991b21df900dda57a/.github/dependabot.yml#L12-L21

That should occasionally bump the base docker image (without  changing the ruby version), rolling new security updates into the production images.

dannymidnight:
> After this ships, @ellsclytn would be suuuper grateful (for SOC2 CVE auditing reasons) if you could configure dependabot like this: https://github.com/buildkite/buildkite/blob/bdc9e38356166dc258d2b02991b21df900dda57a/.github/dependabot.yml#L12-L21

Sounds sensible to me 👍

PR #1910, opened 2023-03-01
Remove deprecated render config and docker image

dannymidnight:
This PR cleans up some unused configuration for Render since we now use environment managed in the web console and the same docker image as production.

Closes #1908 #1909


dannymidnight:
Render link above works so can confirm none of the files being removed in this PR are in use.

PR #1915, opened 2023-03-03
Rails 7

dannymidnight:
Bump rails to the latest version.


mbelton-buildkite:
Is this something we're still trying to do, @dannymidnight?

PR #1916, opened 2023-03-04
add clarity around methods that can be used to set meta-data

jeremybumsted:
Meta-data can only be **set** via the `buildkite-agent meta-data set` command, as it requires an agent token for auth and can't be used outside of the agent context. 

I think this could be worded better, but wanted to get something started :)

### DIFF
--- a/pages/pipelines/build_meta_data.md
+++ b/pages/pipelines/build_meta_data.md
@@ -7,7 +7,7 @@ Meta-data is intended to store data 1 kilobyte or less in size, to be used acros
 
 ## Setting data
 
-You can set meta-data using the command line or in a script. Both of these methods use the `buildkite-agent` cli with the `meta-data` command.
+You can set meta-data using the command line or in a script. Both of these methods use the `buildkite-agent` cli with the `meta-data` command. Meta-data can only be set using the `buildkite-agent meta-data set` command, setting meta-data via API or other methods is not possible.
 
 To set meta-data in the meta-data store, use the `set` command with a key/value pair:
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/build_meta_data.md
```diff
@@ -7,7 +7,7 @@ Meta-data is intended to store data 1 kilobyte or less in size, to be used acros
 
 ## Setting data
 
-You can set meta-data using the command line or in a script. Both of these methods use the `buildkite-agent` cli with the `meta-data` command.
+You can set meta-data using the command line or in a script. Both of these methods use the `buildkite-agent` cli with the `meta-data` command. Meta-data can only be set using the `buildkite-agent meta-data set` command, setting meta-data via API or other methods is not possible.
```
mbelton-buildkite, line 10: Hmmm how about simplifying this to:

```suggestion
The agent's `meta-data` command is the only method for setting meta-data. You can run the command from the command line or in a script.
```

Or, if there should be more explanation, then:

```suggestion
You can set meta-data using the command line or in a script. Both methods use the agent's `meta-data` command because setting meta-data must happen in the agent context.
```

What do you think? Is it enough to call this out as the only method, or do we need to explicitly mention where it's not possible?

jeremybumsted:
> Hey @jeremybumsted, I'm just skimming through some older PRs to see if we can resolve them. This one feels pretty close. What do you think?

@mbelton-buildkite yep, I just committed your suggestions which are excellent! ✨ I think we can merge/close if you're happy with it too!

mbelton-buildkite (approved):

PR #1917, opened 2023-03-06
[PDP-609] Add BUILDKITE_GROUP_ID/_KEY/_LABEL to job env docs

dabarrell:


dabarrell:
CleanShot 2023-03-06 at 16 47 12

^ handy screenshot of the new vars

data/content/environment_variables.yaml
```diff
@@ -215,6 +215,21 @@ variables:
     The GitHub deployment payload data as serialized JSON. Only available on builds triggered by a [GitHub Deployment](https://developer.github.com/v3/repos/deployments/).
   modifable: false
   example: "production"
+- name: BUILDKITE_GROUP_ID
+  desc: |
+    The UUID of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
```
mbelton-buildkite, line 220: A couple of thoughts:

- I think this reads better without "that." 
- What happens if someone requests the variable when the job doesn't belong to a group step? Does it fail, or return a null value? I think it would be better to describe the outcome. Thinking something like:

```suggestion
    The UUID of the [group step](/docs/pipelines/group-step) the job belongs to. Returns `Null` if the job doesn't belong to a group step.
```

What do you think? This applies to the other two values, too.

data/content/environment_variables.yaml
```diff
@@ -215,6 +215,21 @@ variables:
...
+  modifable: false
+  example: "4a331026-8c9a-4714-aff0-8aa30211a34e"
+- name: BUILDKITE_GROUP_KEY
+  desc: |
+    The value of the `key` attribute of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
+  modifable: false
+  example: "audit-tasks"
+- name: BUILDKITE_GROUP_LABEL
+  desc: |
+    The label/name of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
```
mbelton-buildkite, line 230: It would be good to avoid the "a/b" syntax here. Looking at the group step, I assume this is the `label` attribute. If so, let's follow the same pattern you used for the `key` attribute.

```suggestion
    The value of the `label` attribute of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
```

Does that work?

data/content/environment_variables.yaml
dabarrell: @mbelton-buildkite the "that" comment makes me realise that I've ended the sentence with a preposition, and grammatically it should be "The UUID of the group step to which the job belongs." I'll let you tell me if I should make that change!

Regarding the return value – these are environment variables, so `null` doesn't really apply here. In the case where the job isn't in a group, the environment variables simply won't be set. How about this?

```suggestion
    The UUID of the [group step](/docs/pipelines/group-step) the job belongs to. This variable will not be set if the job doesn't belong to a group step.
```

dabarrell: I'll note I copied this sentence from a few other env vars, so if you prefer this, I could go and update those too

data/content/environment_variables.yaml
dabarrell: Not quite - it's either the label attribute, or the name set using `group` - you can read more here: https://bk-docs-pr-1917.onrender.com/docs/pipelines/group-step#group-step-attributes

This was copied from the [BUILDKITE_LABEL](https://bk-docs-pr-1917.onrender.com/docs/pipelines/environment-variables#BUILDKITE_LABEL) docs, which work the same way, but applies to the current job, not the parent group

mbelton-buildkite: Okay, wishful thinking on my part haha. Let's leave it like you had, with the "that" removed here too.

```suggestion
    The label/name of the [group step](/docs/pipelines/group-step) the job belongs to. This variable is only available if the job belongs to a group step.
```

data/content/environment_variables.yaml
```diff
@@ -215,6 +215,21 @@ variables:
...
   modifable: false
   example: "production"
+- name: BUILDKITE_GROUP_ID
+  desc: |
+    The UUID of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
+  modifable: false
+  example: "4a331026-8c9a-4714-aff0-8aa30211a34e"
+- name: BUILDKITE_GROUP_KEY
+  desc: |
+    The value of the `key` attribute of the [group step](/docs/pipelines/group-step) that the job belongs to. This variable is only available if the job belongs to a group step.
```
mbelton-buildkite, line 225: Removing the "that," like the others.

```suggestion
    The value of the `key` attribute of the [group step](/docs/pipelines/group-step) the job belongs to. This variable is only available if the job belongs to a group step.
```

mbelton-buildkite (approved):
One nitpick with "that," but otherwise looks good to me 😊

dabarrell:
Done! will merge in

PR #1918, opened 2023-03-06
Notifications payload job states explained

lizrabuya:
Expand on the differences between Webhook Event notifications and Eventbridge notifications job states

Reference: https://secure.helpscout.net/conversation/2174018311/47841 - asked recently, and another one https://secure.helpscout.net/conversation/1010246694/7911/ asked from 3 yrs ago

### DIFF
--- a/pages/apis/webhooks/job_events.md
+++ b/pages/apis/webhooks/job_events.md
@@ -24,6 +24,7 @@
 </tbody>
 </table>
 
+
 ## Request body data
 
 <table>
@@ -55,3 +56,6 @@ Example request body:
   }
 }
 ```
+
+>📘
+> The job states of webhook events do not return <code>finished</code>, but returns <code>passed</code> or <code>failed</code> based on the  <a href="/docs/apis/rest-api/builds">REST API</a>  for legacy compatibility.  
--- a/pages/integrations/amazon_eventbridge.md
+++ b/pages/integrations/amazon_eventbridge.md
@@ -449,6 +449,8 @@ AWS EventBridge has strict limits on the size of the payload as documented in [A
   }
 }
 ```
+>📘
+> To determine if a job has <code>passed</code> or <code>failed</code>, that is reflected by the "passed" boolean which is <code>true</code> or <code>false</code>.
 
 <h3 id="events-job-activated">Job Activated</h3>
 


pages/apis/webhooks/job_events.md
```diff
@@ -24,6 +24,7 @@
 </tbody>
 </table>
 
+
```
mbelton-buildkite, line 27: Typo: Remove extra blank line.

```suggestion
```

pages/apis/webhooks/job_events.md
```diff
@@ -55,3 +56,6 @@ Example request body:
   }
 }
 ```
+
+>📘
+> The job states of webhook events do not return <code>finished</code>, but returns <code>passed</code> or <code>failed</code> based on the  <a href="/docs/apis/rest-api/builds">REST API</a>  for legacy compatibility.
```
mbelton-buildkite, line 61: I found this a bit confusing. Perhaps it should be in the "Request body data" section? But that section already links you to the REST API objects. I wonder if this could be left out. It doesn't look like any of the customer problems have arisen from confusion from this side of things. What do you think?

```suggestion
```

pages/integrations/amazon_eventbridge.md
```diff
@@ -449,6 +449,8 @@ AWS EventBridge has strict limits on the size of the payload as documented in [A
   }
 }
 ```
+>📘
```
mbelton-buildkite, line 452: It's best to include a title for notes and tips like this. Maybe:

```suggestion

>📘 Check job result
```

^ Also add a blank line to separate the element from the code block.

pages/integrations/amazon_eventbridge.md
```diff
@@ -449,6 +449,8 @@ AWS EventBridge has strict limits on the size of the payload as documented in [A
...
+> To determine if a job has <code>passed</code> or <code>failed</code>, that is reflected by the "passed" boolean which is <code>true</code> or <code>false</code>.
```
mbelton-buildkite, line 453: I think we can simplify the language a bit, and use Markdown formatting. Thinking:

```suggestion
> To check a job's result, use the `passed` field. The value is `true` when the job passed, and `false` otherwise.
```

What do you think?

pages/apis/webhooks/job_events.md
lizrabuya: Hi @mbelton-buildkite , yes this is a bit confusing. I am also not quite convinced this is the right place to add this note. But then, we really have no place to describe that a job state has different values for webhook notifications and EventBridge notifications. However, we do mention that they are different for REST API [here](https://buildkite.com/docs/pipelines/defining-steps#job-states).  I am fine not to have this PR merged. But maybe we can have better docs for our notification services integration payloads.

mbelton-buildkite: I think it would be good to merge the change on the Eventbridge page. From my understanding, that directly addresses the problem the customers ran into. I think it would be best to remove the change to this file for now. What do you think?

lizrabuya: Thanks for the inputs @mbelton-buildkite . We can close this PR and I'll raise a new one with a note at the Eventbridge notification page.

lizrabuya:
Closing this PR for https://github.com/buildkite/docs/pull/1966

PR #1925, opened 2023-03-08
[PDP-603] Update build matrix docs to list use in agent query rules

dabarrell:
We've updated matrices to support use in agent query rules. This updates the docs to reflect that, as well as adding in an example use.

The code change has hit production, so this can be merged when approved


mbelton-buildkite (approved):
Nice 😊

PR #1929, opened 2023-03-09
Break out branch limiting instructions

MelissaKaulfuss:
As per feedback from @timriley, I propose moving the optional branch limiting filter to it's own section. Github Merge Queues will work out of the box with the first steps, users only need to add the filter pattern if they already have existing branch limiting filters.

### DIFF
--- a/pages/tutorials/github_merge_queue.md
+++ b/pages/tutorials/github_merge_queue.md
@@ -18,12 +18,16 @@ To enable a merge queue for a pipeline:
 
 1. From your Buildkite dashboard, select your pipeline.
 1. Select _Pipeline Settings_ > _GitHub_.
+1. In the _GitHub Settings_ section, select the _Skip builds with existing commits_ checkbox.
+
+If you have existing branch limiting rules:
+
+1. From your Buildkite dashboard, select your pipeline.
+1. Select _Pipeline Settings_ > _GitHub_.
 1. In the _Branch Limiting_ section, add a filter for the following pattern:
 
     ```text
     gh-readonly-queue/{base_branch}/*
     ```
-
-1. In the _GitHub Settings_ section, select the _Skip builds with existing commits_ checkbox.
 
 That's it! Your pipeline supports merge queues in GitHub. 🎉


timriley (approved):
This looks good to me, @MelissaKaulfuss!

pages/tutorials/github_merge_queue.md
```diff
@@ -16,6 +16,12 @@ The merge queue creates temporary branches with a special prefix to validate pul
...
 
+1. From your Buildkite dashboard, select your pipeline.
+1. Select _Pipeline Settings_ > _GitHub_.
+1. In the _GitHub Settings_ section, select the _Skip builds with existing commits_ checkbox.
+
+If you have existing branch limiting rules:
+
 1. From your Buildkite dashboard, select your pipeline.
 1. Select _Pipeline Settings_ > _GitHub_.
 1. In the _Branch Limiting_ section, add a filter for the following pattern:
```
mbelton-buildkite, line 27: It would be good to avoid having to repeat the navigation instructions. What do you think about keeping it as one list, but outlining when you need to make changes to the branch limiting rules? Thinking:

```suggestion
1. In the _Branch Limiting_ section, check whether there are any existing branch filter patterns.
   - If none, you don't need to do anything further.
   - If there are existing patterns, add a filter for the following pattern:
```

pages/tutorials/github_merge_queue.md
```diff
@@ -24,6 +30,4 @@ To enable a merge queue for a pipeline:
     gh-readonly-queue/{base_branch}/*
     ```
```
mbelton-buildkite, line 31: If using my suggestion to keep one list, you'll need to update the indentation here:

````suggestion
        gh-readonly-queue/{base_branch}/*
        ```
````

Note, the line above also needs updating to match, but I can't comment on it.

File: pages/tutorials/github_merge_queue.md [hunk 3]
```diff
@@ -16,6 +16,12 @@ The merge queue creates temporary branches with a special prefix to validate pul
```
mbelton-buildkite, line 16: I can't comment on it, but let's update the ordering of the intro paragraph for the steps on line 15 to align with the updated instructions. Thinking:

```
The merge queue creates temporary branches with a special prefix to validate pull request changes. To avoid these temporary branches triggering multiple builds on the same pipeline for the same commit, we recommend enabling the option to skip builds on existing commits.

If you already use branch limiting rules in the pipeline, you must also add a filter to match branches with the special prefix.
```

mbelton-buildkite (commented):
One suggestion to avoid repetition, and an update to the intro paragraph to match the change.

PR #1933, opened 2023-03-09
Fix GraphQL pipeline

dannymidnight:
Some minor updates to keep GraphQL pipeline moving along.

- Fix invalid auto-merge command 😬 
- Stage file before generating file hash (using `git rev-parse`).


dannymidnight:
Whoops. I just noticed that I authored these commits as the bot. I'll shall fix this.

PR #1959, opened 2023-03-20
Assert user has permission before pushing to ECR

dannymidnight:
We frequently see permission denied errors when merging PRs from users that don't have permission to deploy. This PR changes the deployment pipeline for first check the user team earlier for manual intervention (in the form a block step) if necessary. This is what would've happened if one of these failing builds were to ever make it to the final deployment step. 

Still not the ideal workflow but it should make merging marginally simpler.


mbelton-buildkite:
I think we discussed this and decided not to do it, but I don't remember why. Is this something we should still pursue?

PR #1962, opened 2023-03-20
add clarification for agent token configuration command

HugeIRL:
This commit will add a small subsection under the Agent Token configuration that describes only needing to replace the "INSERT-YOUR-AGENT-TOKEN-HERE" for clarification.

### DIFF
--- a/pages/agent/v3/macos.md
+++ b/pages/agent/v3/macos.md
@@ -11,7 +11,7 @@ If you have <a href="http://brew.sh/">Homebrew</a> installed you can use our <a 
 brew tap buildkite/buildkite && brew install buildkite-agent
 ```
 
-Then configure your [agent token](/docs/agent/v3/tokens):
+Then configure your [agent token](/docs/agent/v3/tokens) by replacing `INSERT-YOUR-AGENT-TOKEN-HERE` with your agent token and running:
 
 ```shell
 sed -i '' "s/xxx/INSERT-YOUR-AGENT-TOKEN-HERE/g" "$(brew --prefix)"/etc/buildkite-agent/buildkite-agent.cfg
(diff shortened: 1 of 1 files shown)


pages/agent/v3/macos.md
```diff
@@ -17,6 +17,8 @@ Then configure your [agent token](/docs/agent/v3/tokens):
 sed -i '' "s/xxx/INSERT-YOUR-AGENT-TOKEN-HERE/g" "$(brew --prefix)"/etc/buildkite-agent/buildkite-agent.cfg
 ```
 
+In the above command, you only need to replace `INSERT-YOUR-AGENT-TOKEN-HERE` and then run the command.
+
```
mbelton-buildkite, line 21: Rather than adding a new line here, I think it would be good to update the lead-in sentence for the command. Thinking:

```
Then configure your [agent token](/docs/agent/v3/tokens) by replacing `INSERT-YOUR-AGENT-TOKEN-HERE` with your agent token and running:
```

What do you think?

HugeIRL: 100% agree, that's a much better place for it. The end goal is that any customer that reads this message understands they _only_ need to replace `INSERT-YOUR-AGENT-TOKEN-HERE` and nothing else within that command.

mbelton-buildkite:
I just got a review request for this, but it looks like nothing has changed since the last review?

pages/agent/v3/macos.md
HugeIRL: I've added these changes in now and removed my original.

mbelton-buildkite (approved):
Looks good to me!

mbelton-buildkite:
Hi @HugeIRL, I've made a few other changes to this section this morning so I decided to include this change in them: #1991. I'll close this PR and merge that one.

PR #1956, opened 2023-03-17
feat: add documentation for new usage page

jfhinchcliffe:
we're about to launch a new usage page for job minutes and test executions. this PR adds the documentation for the new page.

There isn't an overarching section for information on both Job Minutes and Test Analytics, so I've duplicated thigns a bit across each section.

The job minutes / usage page has been updated, and a new test analytics page added which is title 'Test Executions'. Could be convinced that this should go somewhere else, but seemed logical enough for a start.

PR #1969, opened 2023-03-23
Fdn 2321 docs cleanup

jfhinchcliffe:
Some cleanup after merging this PR (https://github.com/buildkite/docs/pull/1956) a bit preemptively after setting Usage Breakdown live for everyone.

From @mbelton-buildkite 

> I noticed a couple of things that still need to be addressed:
> 
> A few images were merged in that are not used in the doc. These should be deleted.
> The nav entry should use sentence case: Test Executions -> Test executions.


mbelton-buildkite (approved):
Looks good, thanks for sorting these 😊

PR #1980, opened 2023-03-27
fix missing quotation mark in HTTP example

HugeIRL:
When troubleshooting, I noticed we missed a double quotation mark in our HTTP example in the docs. This commit adds the missing double quotation mark.


mbelton-buildkite (approved):
Thanks for catching this and fixing it!

james2791 (approved):

PR #1960, opened 2023-03-20
Add documentation for Clusters

eleanorakh:
This PR adds documentation for the upcoming public beta of Clusters. The content focuses on an overview of the clusters feature and then offers some guidance on the feature itself and how a user can access it.

PR #1983, opened 2023-03-28
Improve clusters docs

timriley:
Addressing all the feedback from #1960. Thank you for going into this so deeply, @mbelton-buildkite!


mbelton-buildkite (approved):
Looks good to me! Thanks for making these updates :)

PR #1988, opened 2023-03-29
Fix bug in pre-bootstrap example in securing.md

mechanical-fish:
When I tried to use the example syntax for the `pre-bootstrap` script, it failed because it didn't strip the leading and trailing quotes when extracting the value of a Buildkite env var from `BUILDKITE_ENV_FILE`. Adding the appropriate `sed` call made my script work.

### DIFF
--- a/pages/agent/v3/securing.md
+++ b/pages/agent/v3/securing.md
@@ -83,7 +83,7 @@ set -euo pipefail
 
 for line in "$(grep "^BUILDKITE_REPO=" "${BUILDKITE_ENV_FILE}")"
 do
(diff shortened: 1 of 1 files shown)


mbelton-buildkite:
Hi @mechanical-fish, thanks for raising this! I'm asking someone who works on the agent to review this for us. 😊

mbelton-buildkite:
Thanks @sj26! I'll trigger the builds for this branch and then merge when they pass. 😊

mbelton-buildkite (approved):

PR #1994, opened 2023-03-31
Add automatic retry default limit

lizrabuya:


pages/pipelines/command_step.md
```diff
@@ -261,7 +261,7 @@ steps:
 ```
 
-If you retry a job, the information about the failed job(s) remains, and a new job is created. The history of retried jobs is preserved and immutable. The number of possible retries is available as an [environment variable `limit`](/docs/pipelines/command-step#retry-attributes-automatic-retry-attributes) on the job.
+If you retry a job, the information about the failed job(s) remains, and a new job is created. The history of retried jobs is preserved and immutable. The number of possible retries is available as an [environment variable `limit`](/docs/pipelines/command-step#retry-attributes-automatic-retry-attributes) on the job. When a limit is not specified on automatic retry, the default limit is 3.
```
mbelton-buildkite, line 264: I quite like the way a number sticks out here, but the general guidance from [Microsoft](https://learn.microsoft.com/en-us/style-guide/numbers) and [Google](https://developers.google.com/style/numbers) is to spell out numbers from zero to nine. Let's follow that:

```suggestion
If you retry a job, the information about the failed job(s) remains, and a new job is created. The history of retried jobs is preserved and immutable. The number of possible retries is available as an [environment variable `limit`](/docs/pipelines/command-step#retry-attributes-automatic-retry-attributes) on the job. When a limit is not specified on automatic retry, the default limit is three.
```

mbelton-buildkite (approved):
One style nitpick to fix, otherwise looks good :)

PR #2002, opened 2023-04-04
Differentiate GraphQL nullable and non-null types

dannymidnight:
It was recently reported that our GraphQL docs ain't quite right. (Thanks @jradtilbrook!) 

This PR removes the "required" label from all GraphQL fields and prefixes non-null types with the standard `!` syntax.

| Before  | After |
| --- | --- |
|  Screenshot 2023-04-04 at 4 58 23 pm | Screenshot 2023-04-04 at 4 58 27 pm |


jradtilbrook:
LGTM 🎉

scripts/graphql_api_content/render_helpers.rb
```diff
@@ -10,14 +10,17 @@ def render_html(markdown)
...
     else
-      kind = of_type["kind"] || ""
-      if name = of_type["name"]
+      kind = type["kind"] || ""
+      nullable_symbol = nullable ? "!" : ""
+
+      if name = type["name"]
         <<~HTML
-          <a href="/docs/apis/graphql/schemas/#{kind.downcase}/#{name.downcase}" class="pill pill--#{kind.downcase} pill--normal-case pill--#{size}" title="Go to #{kind} #{name}"><code>#{name}</code></a>
+          <a href="/docs/apis/graphql/schemas/#{kind.downcase}/#{name.downcase}" class="pill pill--#{kind.downcase} pill--normal-case pill--#{size}" title="Go to #{kind} #{name}"><code>#{nullable_symbol}#{name}</code></a>
```
mbelton-buildkite, line 23: The nullable symbol should be a suffix rather than a prefix. From [the spec](https://spec.graphql.org/October2021/#sec-Non-Null):

> A trailing exclamation mark is used to denote a field that uses a Non-Null type like this: `name: String!`.

mbelton-buildkite (commented):
Just need to change where the symbol displays, otherwise looks good :)

scripts/graphql_api_content/render_helpers.rb
dannymidnight: 🤦‍♂️!

mbelton-buildkite (approved):
Looks good to me!

PR #2003, opened 2023-04-04
Use Ruby 3.1.4

JuanitoFatas:
This PR bumps docs repo to use Ruby v3.1.4. See [:basecamp: internal basecamp post for context](https://3.basecamp.com/3453178/buckets/1933834/messages/6006053479).

### Changes

- [Matrix](https://github.com/ruby/matrix) now need to be installed as gem
- Psych 4 has incompatible change and breaks `YAML.load_file`, so change to use the equivalent `Psych.safe_load_file` method instead and permitted `Date` class to be handled https://bugs.ruby-lang.org/issues/17866
- `ERB.new` removed 2nd argument and accept `trim_mode` by keyword argument (old 3rd argument) now.

### Verification

- [x] tests passed
- [x] app runs ok locally
 
  CleanShot 2023-04-04 at 20 34 40@2x

  CleanShot 2023-04-04 at 20 34 51@2x

### Deployment

[Normal deploys](https://github.com/buildkite/engineering-guide/blob/main/process/deployments/index.md#normal-deploys).

Low risk: I’ve verified it works locally and no outstanding issue I can find.

### Rollback

Low risk. Revert forward with GitHub Revert.


dannymidnight (approved):
Thanks for landing this one @JuanitoFatas. Changes LGTM

PR #2007, opened 2023-04-05
Yes, we do. isn't helpful.

mcncl:
Trying to add some more helpful information for those who mightn't know what JIT is.

### DIFF
--- a/pages/integrations/sso.md
+++ b/pages/integrations/sso.md
@@ -96,7 +96,7 @@ Yes, team maintainers can select whether a user is 'required' to use SSO or whet
 ### Do you support JIT provisioning?
 Yes, we do.
 
-### What **is** JIT provisioning?
+### What is JIT provisioning?
 Just In Time (JIT) provisioning is a method of creating accounts at the point in which they are needed; you can give a user access to Buildkite in your SSO provider, but their account won't be created (and thus they won't use a "seat") until it's required, which is assumed to be their first login attempt.
 
 ### What happens if a person leaves our company?
--- a/pages/integrations/sso.md
+++ b/pages/integrations/sso.md
@@ -97,7 +97,7 @@ Yes, team maintainers can select whether a user is 'required' to use SSO or whet
 Yes, we do.
 
 ### What is JIT provisioning?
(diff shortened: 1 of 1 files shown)


pages/integrations/sso.md
```diff
@@ -96,6 +96,9 @@ Yes, team maintainers can select whether a user is 'required' to use SSO or whet
 ### Do you support JIT provisioning?
 Yes, we do.
 
+### What **is** JIT provisioning?
```
mbelton-buildkite, line 99: It doesn't look like the bolding shows up in the heading, but I also don't think it's necessary here:

```suggestion
### What is JIT provisioning?
```

pages/integrations/sso.md
```diff
@@ -96,6 +96,9 @@ Yes, team maintainers can select whether a user is 'required' to use SSO or whet
...
+Just In Time (JIT) provisioning is a method of creating accounts at the point in which they are needed; you can give a user access to Buildkite in your SSO provider, but their account won't be created (and thus they won't use a "seat") until it's required, which is assumed to be their first login attempt.
```
mbelton-buildkite, line 100: General tightening:

```suggestion
Just-in-time user provisioning (JIT provisioning) creates accounts only when needed. You can grant a user access to Buildkite through your SSO provider, but their account won't be created until it's required—typically upon their first login attempt. For billing purposes, the user doesn't exist until their account is created.
```

mbelton-buildkite, line 100: Actually, I think this can flow on without needing a separate heading:

```suggestion
Yes, we do. Just-in-time user provisioning (JIT provisioning) creates accounts only when needed. You can grant a user access to Buildkite through your SSO provider, but their account won't be created until it's required—typically upon their first login attempt. For billing purposes, the user doesn't exist until their account is created.
```

What do you think?

mbelton-buildkite: This suggests a reword, but I think it's better to use the suggestion below to include the text in the previous question heading.

mcncl:
@mbelton-buildkite made the changes you suggested, including making it a single section (see latest commit)

mbelton-buildkite (approved):
So much better than just, "Yes, we do." 🙌🏻

PR #2014, opened 2023-04-10
Commit push to branch with PR

nsuma8989:
Add a note to explain the behavior on how builds are triggered when we have an open PR and a commit is pushed to the branch associated with the PR.

### DIFF
--- a/pages/integrations/github.md
+++ b/pages/integrations/github.md
@@ -80,6 +80,9 @@ Optionally, select one or more of the following:
 
 If you want to run builds only on pull requests, set the _Branch Filter Pattern_ in the pipeline to a branch name that will never occur (such as "this-branch-will-never-occur"). Pull request builds ignore the _Branch Filter Pattern_, and all pushes to other branches that don't match the pattern are ignored.
 
+>🚧
+> When a commit is pushed to a branch which already has an one pull request then only build for the code push will happen and associate pull request related information to that build instead of running two builds one for code push and one for pull request.
+
 ## Running builds on git tags
 
 Builds are only run for tags when a [`push` event is triggered](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#push). To enable builds for `push` events for git tags, edit the _GitHub settings_ for your Buildkite pipeline, and choose the _Build Tags_ checkbox.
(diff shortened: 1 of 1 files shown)


pages/integrations/github.md
```diff
@@ -80,6 +80,9 @@ Optionally, select one or more of the following:
 
 If you want to run builds only on pull requests, set the _Branch Filter Pattern_ in the pipeline to a branch name that will never occur (such as "this-branch-will-never-occur"). Pull request builds ignore the _Branch Filter Pattern_, and all pushes to other branches that don't match the pattern are ignored.
 
+>🚧
+> When a commit is pushed to a branch which already has an one pull request then only build for the code push will happen and associate pull request related information to that build instead of running two builds one for code push and one for pull request.
```
mbelton-buildkite, line 84: I'm finding this a bit hard to understand, sorry. I think it's saying:

> When you push a commit to a branch that already has a PR open, only one build is run.

Is that right? Can you tell me a bit about where this came from? Was there a situation where customers were confused?

nsuma8989: @mbelton-buildkite Yes that is correct. For first time when a commit is pushed then it runs build for both commit push and PR but when any further commits are pushed to a branch where there is an open PR then only one build happens. We do get this question from customers where this is not clear so updating this note for their understanding

mbelton-buildkite: Sorry, I'm struggling with this one. This is the default logic I'm understanding:

- **Branch with no PR:** One build is triggered for each commit.
- **Some state:** Two builds are triggered. One for the commit, one for the PR.
- **Branch with open PR:** One build is triggered for each commit.  

Is that right? If so, what's the "Some state?" Is it when the PR is created?

nsuma8989: Yes some state is when a PR is created

mbelton-buildkite: Ahh okay, I'll post a new suggestion.

mbelton-buildkite, line 84: Okay, how about this:

```suggestion
When you create a pull request, two builds are triggered: one for the pull request and one for the most recent commit. However, any commit made after the pull request is created only triggers one build.
```

(I don't think it needs to be in a warning callout)

nsuma8989: Makes sense. I accepted the changes and committed it

mbelton-buildkite (approved):

PR #2017, opened 2023-04-11
Remove "Cloning a pipeline" section

timriley:
This only works as expeted for pipelines with legacy (UI-defined) steps, not for pipelines with YAML steps.

Remove this for now to avoid customer confusion.


jordandcarter (approved):
👍

I quickly searched our bk src for 'cloning-a-pipeline' to find any direct links and came up empty 🙌

mbelton-buildkite (approved):
Sounds good to me if this has confused customers. If it's still useful for some people to know, you could instead include a callout after the heading like:

> 🚧 **Legacy feature**
> Cloning a pipeline using the `?clone` URL parameter only works for pipelines defined in the UI. Pipelines defined in YAML are not supported.

timriley:
Thanks for checking this out, @mbelton-buildkite. Given our commitment to rolling forward with yaml steps, I'll go ahead and remove this altogether.

PR #2018, opened 2023-04-11
Add note about block/input step not supported

nsuma8989:
Currently if concurrency gates have block/input step then it wont support it

### DIFF
--- a/pages/pipelines/controlling_concurrency.md
+++ b/pages/pipelines/controlling_concurrency.md
@@ -84,6 +84,9 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
+>🚧 
+> Concurrency gates does not support if there are `input` or `block` step inside the gates.
+
 ### Controlling command order
 
 By default, steps that belong to the same concurrency group are run in the order that they are added to the pipeline.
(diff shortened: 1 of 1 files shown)


pages/pipelines/controlling_concurrency.md
```diff
@@ -84,6 +84,9 @@ steps:
 ```
 
+>🚧 
+> Concurrency gates does not support if there are `input` or `block` step inside the gates.
```
mbelton-buildkite, line 88: I think this can be worked into the existing text rather than put in a warning callout. Perhaps the first paragraph in the section could be changed to:

> Sometimes you need strict concurrency while also having jobs that would benefit from parallelism. In these situations, you can use _concurrency gates_ to control which jobs run in parallel and which jobs run one at a time. You can't use input or block steps inside concurrency gates.

What do you think?

nsuma8989: @mbelton-buildkite I am worried that it will confuse users that they cannot define input or block steps inside concurrency gates. If you think it wont be then I can change it to that

mbelton-buildkite: Sorry, I think I'm misunderstanding the change. I thought this was saying that a caveat to using concurrency gates is that you can't use block or input steps inside them. I see you've said, "does not support." What does that mean? What happens if someone does define block or input steps inside concurrency gates?

nsuma8989:
@mbelton-buildkite Updated the PR with suggestions, please let me know if we are good to merge this

mbelton-buildkite (approved):

PR #2019, opened 2023-04-11
Fix example in retry docs

francoiscampbell:
The current syntax isn't valid, it triggers the error
> `reason` is not a valid property on the retry ruleset. Please check the documentation to get a full list of supported properties.

The updated file matches the correct syntax used lower down on the page.


mbelton-buildkite (approved):
Looks good to me. Thanks for finding and fixing this @francoiscampbell! I'll run our linting against this PR and merge when that passes. 😊

PR #2020, opened 2023-04-11
Update Grace period for Tokens

stephanieatte:
Notice about Agent Token Grace Period

### DIFF
--- a/pages/agent/v3/tokens.md
+++ b/pages/agent/v3/tokens.md
@@ -14,6 +14,7 @@ The token is used by the Buildkite Agent's [start](/docs/agent/v3/cli-start#star
 It's recommended you use your platform's secret storage (such as the [AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-paramstore.html)) to allow for easier rollover and management of your agent tokens.
 
 ## Creating tokens
+Note: There is a grace period of 10 minutes after Agent Token is created. During this period,  even if the page is reloaded the Token will still be visible. 
 
 New tokens can be created using the [GraphQL API](/docs/apis/graphql-api) with the `agentTokenCreate` mutation.
 


pages/agent/v3/tokens.md
```diff
@@ -14,6 +14,7 @@ The token is used by the Buildkite Agent's [start](/docs/agent/v3/cli-start#star
 It's recommended you use your platform's secret storage (such as the [AWS Systems Manager Parameter Store](https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-paramstore.html)) to allow for easier rollover and management of your agent tokens.
 
 ## Creating tokens
+Note: There is a grace period of 10 minutes after Agent Token is created. During this period,  even if the page is reloaded the Token will still be visible.
```
mbelton-buildkite, line 17: Style tweaks:

```suggestion
There is a grace period of 10 minutes after you create a token. Even if you reload the page, the token remains visible during the grace period.
```

mbelton-buildkite (commented):
The section you're editing says that tokens can only be created through the API, but your addition sounds like someone is creating a token through the UI. Does the section need to be updated to indicate that you can create a token through the UI, too?

> It is possible to create multiple agent tokens using the GraphQL API. These tokens will show up on the [Agents page](https://buildkite.com/organizations/-/agents) in the UI, but can only be managed (created or revoked) using the API.

pages/agent/v3/tokens.md
stephanieatte: @mbelton-buildkite Users were unsure why the Token still showed on the page even after they refreshed the page. Logged in and out. We want to clarify we have a grace period where the token is still visible after creation

mbelton-buildkite:
@stephanieatte, I'm not sure if you saw [my comment](https://github.com/buildkite/docs/pull/2020#pullrequestreview-1391149713). There's no mention of what page someone is looking at when they can use this information. Where is someone creating a token that this grace period applies to?

stephanieatte:
@mbelton-buildkite you can definitely create API token in the UI. what happens is when the user creates it in the UI. It shows on the top of the page for about 10 minutes. So some user thought it was a security breach or something happening. Because usually when you create tokens on other websites it disappears right after

ozdenyilmaz:
This PR is no longer needed. The PR wanted to add a note about a 10-minute grace period after creating a token where it remains visible. The current page already addresses this clearly on line 25–27:

> 📘 An agent token's value is only displayed once
> As soon as the agent token's value is displayed, copy its value and save it in a secure location.

The page also had outstanding review feedback that was never addressed. It can be safely closed.

PR #2024, opened 2023-04-12
Remove documentation for build retention

jordandcarter:
Build retention will no longer be configurable and so this PR removes the documentation for it.

Build export used to live on the build retention page so I've moved it to it's own page build-export and updated the nav.


data/nav.yml
```diff
@@ -224,8 +224,8 @@
           path: "pipelines/artifacts"
         - name: "Pipeline tags"
           path: "pipelines/tags"
-        - name: "Build retention"
-          path: "pipelines/build-retention"
+        - name: "Build export"
```
mbelton-buildkite, line 227: This is a feature name, which assumes people know what it is. I think we could use a title that uses these key words but speaks more to a user's goal:

```suggestion
        - name: "Export build data"
```

What do you think?

pages/pipelines/build_export.md
```diff
@@ -0,0 +1,15 @@
+# Configuring build export
```
mbelton-buildkite, line 1: If you update the nav, also update this heading.

pages/pipelines/build_export.md
```diff
@@ -0,0 +1,15 @@
...
+
+Each [Buildkite plan](https://buildkite.com/pricing) has a maximum build retention
+period, after which your oldest builds will be automatically deleted.
```
mbelton-buildkite, line 4: tightening slightly while we're here:

```suggestion
period, after which your oldest builds are automatically deleted.
```

data/nav.yml
```diff
@@ -224,8 +224,8 @@
           path: "pipelines/artifacts"
         - name: "Pipeline tags"
           path: "pipelines/tags"
-        - name: "Build retention"
-          path: "pipelines/build-retention"
```
mbelton-buildkite, line 228: Can you add a redirect for this URL to the new one? I believe they're defined in [routes.rb](https://github.com/buildkite/docs/blob/main/config/routes.rb) (Is that right, @dannymidnight?)

data/nav.yml
jordandcarter: Yeah that makes sense 👍

data/nav.yml
jordandcarter: Yeah that looks right, I've added in the redirect

dannymidnight (approved):
LGTM

mbelton-buildkite (approved):
Nice!

PR #2038, opened 2023-04-17
Fix GraphQL script IA

mbelton-buildkite:
I'd removed the "Schema browser" label and moved the Cookbook page previously...but I made the changes directly to the nav file. I noticed things had gone back to before and realized the nav file is generated when the GraphQL script runs. This change moves those updates into the script and generates the nav file from that.


dannymidnight (approved):
LGTM 👍🏻

I think I'll add something to our backlog to clean this process up a bit. It's definitely not very intuitive having nav split across YAML and ruby files.

PR #2041, opened 2023-04-17
Add ability to render H3s in Table of Contents

dannymidnight:
Extract H3s from page content and optionally render them into the table of contents for pages that opt-in.

e.g. example.md

```markdown
---
h3s: true
---
```

I've used `h3s: true` as the configuration attribute but totally open to something more descriptive. Perhaps `toc_depth` or something along those lines? 

I've also left the styling wide open, perhaps @olyism has some feels on this one.

Screenshot 2023-04-17 at 4 30 07 pm

### DIFF
--- a/app/assets/javascripts/toc.js
+++ b/app/assets/javascripts/toc.js
@@ -4,7 +4,7 @@ function initToc() {
   const initCurrentLinkListener = () => {
     const scrollPadding = 125;
     const currentClassName = "Toc__link--current";
-    const tocLinkNodes = document.querySelectorAll(".Toc__link");
+    const tocLinkNodes = document.querySelectorAll(".Toc__link--h2");
     const headingNodes = document.querySelectorAll("h2.Docs__heading");
 
     const getDistanceFromTop = (element) => {
--- a/app/assets/stylesheets/components/_toc.scss
+++ b/app/assets/stylesheets/components/_toc.scss
@@ -86,6 +86,10 @@
     }
   }
 
+  .Toc__list--h3 {
+    font-size: 0.7rem;
+  }
+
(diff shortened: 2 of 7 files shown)


mbelton-buildkite:
I've had a quick look at how other tools do the naming and found the following.

[Docusaurus](https://docusaurus.io/docs/next/markdown-features/toc#table-of-contents-heading-level):

```
toc_min_heading_level: 2
toc_max_heading_level: 5
```

[Gatsby](https://www.gatsbyjs.com/plugins/gatsby-remark-table-of-contents/#use):

```
from-heading: 2
to-heading: 6
```

[Hugo](https://www.kiroule.com/article/add-table-of-contents-to-hugo-theme/#:~:text=The%20startLevel%20setting%20defines%20the,create%20a%20table%20of%20contents.)

```
startLevel = 2
endLevel = 5
```

My thoughts:

- I don't think we have any use cases for needing to set the min level.
- I don't think we have any use cases for including headings beyond H3.
- I see the implementation is specific to H3.
- "Min" and "max" are pretty confusing in this situation.
- I think the name should include `toc`.

Let's go with: `toc_include_h3: true`

clbarrell:
> For the styling, I think there should be some indentation on the H3s.

Yeah agree, some indentation would make it more visually obvious.

otherwise looks good!

clbarrell (approved):

dannymidnight:
OK made some further adjustments and fixed an issue with the link URL.

It should now look like the following:

Screenshot 2023-04-19 at 11 19 11 am

Worth noting, I've excluded the h3 links from being highlighted on scroll since there's a few more style considerations. We can definitely come back around on this one.

PR #2047, opened 2023-04-18
Note for session IP pinning to work

lizrabuya:
There was a question raised by a customer on why the session IP address pinning did not work for them after following the current steps in our doc.

After investigating on the issue, it was found out that their org's SSO was configured as optional. For the session IP address to work, it is required that the org must have Required SSO settings configured. This piece of integral information was missing in our docs and hopefully would save us time for debugging when the next customer raise the same issue.


pages/integrations/sso.md
```diff
@@ -78,7 +78,7 @@ You can configure the session duration to any timeout between 6 hours and 8,760
 > 📘 Enterprise feature
 > Pinning SSO sessions to IP addresses is only available on an [Enterprise](https://buildkite.com/pricing) plan.
 
-Session IP address pinning prompts users to re-authenticate when their IP address changes. This prevents session hijacking by restricting authorized sessions to only originate from the IP address used to create the session. If any attempt is made to access Buildkite from a different IP address, the session is instantly revoked and the user must re-authenticate.
+Session IP address pinning prompts users to re-authenticate when their IP address changes. This prevents session hijacking by restricting authorized sessions to only originate from the IP address used to create the session. If any attempt is made to access Buildkite from a different IP address, the session is instantly revoked and the user must re-authenticate. The user must be configured with Required SSO at the organisation's settings for this to work.
```
mbelton-buildkite, line 81: Looking to make this a bit clearer:

```suggestion
Session IP address pinning prompts users to re-authenticate when their IP address changes. This prevents session hijacking by restricting authorized sessions to only originate from the IP address used to create the session. If any attempt is made to access Buildkite from a different IP address, the session is instantly revoked and the user must re-authenticate. Users must be required to use SSO in the [organization's user settings](https://buildkite.com/organizations/~/users) for SSO session IP address pinning to work for them.
```

What do you think?

mbelton-buildkite (approved):

mbelton-buildkite:
@lizrabuya, I'm skimming through old PRs to see if there are any I can quickly resolve, and this looks like a good candidate. I'll update the branch and merge for you. Let me know if there are any issues with that. 😊

PR #2050, opened 2023-04-19
Scheduled builds: Show H3s in table of contents

mbelton-buildkite:


mbelton-buildkite:
FYI @dannymidnight – a small one, but first use of `toc_include_h3` going out 😊

PR #2051, opened 2023-04-19
Highlight nav items in backend markup

dannymidnight:
I've been experimenting with https://turbo.hotwired.dev/ for removing full page reloads and whilst integrating I ran into some issues due to the way we're using javascript to highlight the current nav item (+ ancestors).

This PR replaces the current JS approach with a ruby alternative which in addition to setting things up for turbo, has the major benefit of removing a flicker of content on page load prior to javascript initialising.


dannymidnight:
Interestingly this PR has pointed out some orphaned pages that aren't included in our navigation. 

e.g. https://buildkite.com/docs/agent/v2/-apt-configuration

@mbelton-buildkite can I remove these pages? They look broken and we can always bring them back.

dannymidnight:
OK @mbelton-buildkite I did end up running into two orphaned routes which I've removed and would love your input on whether or not I've done the right thing!

- https://github.com/buildkite/docs/pull/2051/commits/d1e6000a6602d82cf5f46adc4347141dc345852b - this one seemed like an accident.
- https://github.com/buildkite/docs/pull/2051/commits/c018f3410260567a12dbd1cc18c9dad5006ded3d - this one seems like it's not that useful content https://buildkite.com/docs/pipelines/my-builds

mbelton-buildkite:
> OK @mbelton-buildkite I did end up running into two orphaned routes which I've removed and would love your input on whether or not I've done the right thing!

👍🏻 Both look fine to delete.

mbelton-buildkite:
Have you seen this jumping behavior on a wide screen?

https://user-images.githubusercontent.com/119824349/233001422-fae818be-1eb8-4c3d-867a-d782a884b328.mov

It looks amazing on a standard width.

dannymidnight:
> Have you seen this jumping behavior on a wide screen? Is this solved by https://github.com/buildkite/docs/pull/2052?

Good spot. Yep that's right, this will be solved in the other PR. This PR makes the issue more obvious since prior to this change both the _visibility_ and position were both handled via JS. 

This change removes JS from the equation for visibility, https://github.com/buildkite/docs/pull/2052 removes JS from the equation for position.

dannymidnight:
I decided to merge the other change into this branch since they're better bundled together.

app/models/nav.rb
```diff
@@ -1,9 +1,53 @@
...
+    item
+  end
+
+  # Returns a hash of routes, indexed by path
+  #
+  # @example
+  #   {
+  #     "path/to/page": {
+  #       path: "path/to/page",
+  #       parents: ["parent", "grandparent"]
```
dannymidnight, line 26: It's worth calling out that I'm doing route matching by keeping track of ancestors since our URLs aren't representative of our navigation structure. If that were the case then path pattern matching would be much simpler!

e.g.

This is the current routing and the parent that should be highlighted.
`/docs/tutorials` – [Pipelines]

Ideally in this case we'd be using `/docs/pipelines/tutorials/`, for simplicities sake, since we'd then be able to determine Pipelines should be highlighted based purely on the URL. I'm aware there's more to consider before making this change, redirects etc. but if we do ever make that change then we should be able to remove a bunch of this code.

mbelton-buildkite: 👍🏻 Even when we look into updating the URLs, we may not want to include section names. That makes it easier to rename sections, move pages, and reduces the URL depth. So all the pages for pipelines may just be `/pipelines/*`. Something to consider later.

mbelton-buildkite:
Bug with screenshot position [here](https://bk-docs-pr-2051.onrender.com/docs/agent/v3/cli-annotate):
Screenshot 2023-04-20 at 12 09 14 pm

mbelton-buildkite (commented):
One bug, otherwise looks good!

mbelton-buildkite (approved):
LGTM

PR #2052, opened 2023-04-19
Use CSS grid for page layout

dannymidnight:
As a continuation from https://github.com/buildkite/docs/pull/2051 this PR replaces all JS related to nav positioning to make it possible to use https://turbo.hotwired.dev/ without shifting layout on page loads.

For this to happen, the sidebar needed to be loaded directly into markup. 

There's some very minor spacing adjustments in this PR but they should be improvements since all alignment is managed via css grids with consistent spacing.

Before:
Screenshot 2023-04-19 at 4 45 41 pm

After:
Screenshot 2023-04-19 at 4 45 11 pm


mbelton-buildkite (approved):
Looks good to me if looks good to @olyism

PR #2055, opened 2023-04-20
Add H3s to the on-page nav by default

mbelton-buildkite:
This change switches the default behavior of the on-page nav to include H3s by default. Almost all pages benefit from showing their H3s. Those that don't have been excluded in this PR.

I've also removed any usage of `toc_include_h3: true` now that it's the default behavior.

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -144,8 +144,8 @@ class Page
     defaults = {
       # Default to rendering table of contents
       "toc": true,
-      # Default to H2s being included in the table of contents
-      "toc_include_h3": false
+      # Default to H3s being included in the table of contents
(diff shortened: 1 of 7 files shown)


app/models/page.rb
```diff
@@ -144,8 +144,8 @@ def metadata
     defaults = {
       # Default to rendering table of contents
       "toc": true,
-      # Default to H2s being included in the table of contents
-      "toc_include_h3": false
+      # Default to H3s being included in the table of contents
```
dannymidnight, line 147: 😆 good fix

dannymidnight (approved):
Nice!

PR #2058, opened 2023-04-20
Fix the heading generation in CLI script

mbelton-buildkite:
The script for generating docs for the agent CLI supported two modes:

- Pages where one command was the whole page. This generated H2 headings.
- Pages with multiple commands on one page. This generated H3 headings.

This change removes the first case. H2s are already manually set on the page the generated content is included on, so we only need to generate H3s. The only heading I needed to add was on the OIDC page.

You can see the change on the `annotate` page:

[Before](https://buildkite.com/docs/agent/v3/cli-annotate):
Screenshot 2023-04-20 at 7 36 30 pm

[After](https://bk-docs-pr-2058.onrender.com/docs/agent/v3/cli-annotate):
Screenshot 2023-04-20 at 7 36 55 pm

Note, we display H3 headings in the on-page nav since #2041 and #2055.


dannymidnight (approved):
Awesome. So much simpler!

triarius (approved):
Nice!

PR #2059, opened 2023-04-20
Update BUILDKITE_MESSAGE empty scenario

nsuma8989:
This PR updates the explanation of environment variable BUILDKITE_MESSAGE where it can be empty of build is triggered from UI and user does not set the message in the build dialog box

### DIFF
--- a/data/content/environment_variables.yaml
+++ b/data/content/environment_variables.yaml
@@ -269,7 +269,7 @@ variables:
     - false
 - name: BUILDKITE_MESSAGE
   desc: |
-    The message associated with the build, usually the commit message.
+    The message associated with the build, usually the commit message. This value can be blank if message is not set when builds manually triggered using Buildkite web interface.
   modifiable: false
   example: "Added a great new feature"
 - name: BUILDKITE_ORGANIZATION_SLUG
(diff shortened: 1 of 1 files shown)


data/content/environment_variables.yaml
```diff
@@ -269,7 +269,7 @@ variables:
     - false
 - name: BUILDKITE_MESSAGE
   desc: |
-    The message associated with the build, usually the commit message.
+    The message associated with the build, usually the commit message. This value can be blank if message is not set when builds manually triggered using Buildkite web interface.
```
mbelton-buildkite, line 272: I think splitting up the ideas a bit makes it simpler to understand:

```suggestion
    The message associated with the build, usually the commit message. The value is empty when a message is not set. For example, when a user triggers a build from the Buildkite dashboard without entering a message, the variable returns an empty value.
```

What do you think?

nsuma8989: Should we say Buildkite dashboard or Buildkite web interface ? I saw for few of the variable we say Buildkite web interface so just confirming.

mbelton-buildkite: I found a post that pre-dates me about trying to move towards _Buildkite dashboard_, so have been using that for new content or updates.

nsuma8989: Thanks for confirming. I made the changes suggested

mbelton-buildkite (approved):

PR #2061, opened 2023-04-20
Explain about local and global hooks

nsuma8989:


pages/agent/v3/hooks.md
```diff
@@ -9,6 +9,8 @@ Hooks can be defined in the following locations:
 * In your pipeline's repository
 * In [plugins](/docs/plugins) applied to steps
 
+Agent hooks are also referred to as Global hooks while pipeline's repository hooks are referred to as Local hooks.
```
mbelton-buildkite, line 12: General style + use italics for introducing a new term:

```suggestion
Hooks defined on the agent are also referred to as _global hooks_, while hooks defined in the pipeline's repository are known as _local hooks_.
```

mbelton-buildkite (approved):
One suggestion for the wording, otherwise looks good :)

PR #2064, opened 2023-04-21
Experiment with using Turbo (links) ⚡️ 

dannymidnight:
This PR drops in https://github.com/hotwired/turbo-rails (formerly known as TurboLinks) as an experiment to make the docs UI a little snappier by removing full page reloads. 

In order to drop in Turbo I needed to first upgrade the docs asset pipeline to something that could import node_modules since up until this point we haven't imported any external packages (in our app). I decided to go with [Vite](https://vitejs.dev/) / [Vite Ruby](https://vite-ruby.netlify.app/) since everyone seems to be moving away from the whole webpack ecosystem with rails webpacker no longer sounding like the preferred path. 

Annoyingly, I also needed to bump our node version for Turbo so I've upgraded to the latest LTS release _and_ I needed Yarn for Vite so I've also bumped that to the latest version. 

Now that I've got this all working, I think I'd like to release this specific PR in a few chunks:

- [x] Upgrade Node + Yarn  
- [x] Use Vite for JS
- [ ] Turbo


Dockerfile
```diff
@@ -49,7 +46,7 @@ COPY . /app
 # Compile sprockets
 RUN if [ "$RAILS_ENV" = "production" ]; then \
     echo "--- :sprockets: Precompiling assets" \
-    && RAILS_ENV=production RAILS_GROUPS=assets bundle exec rake assets:precompile \
+    && RAILS_ENV=production RAILS_GROUPS=assets SECRET_KEY_BASE=xxx bundle exec rake assets:precompile \
```
dannymidnight, line 49: This is a bit weird but it matches up with https://github.com/buildkite/docs/blob/main/docker-compose.production.yml#L11 to get around https://github.com/rails/rails/issues/32947

app/assets/javascripts/.gitkeep
dannymidnight, line 1: Had to keep this folder around whilst sprockets is still being used for styles and images.

PR #2066, opened 2023-04-21
Node v18 + Yarn

dannymidnight:
Pulling the node and yarn upgrade out from https://github.com/buildkite/docs/pull/2064

Also since yarn seems to be used for some rails things, I've dropped the package-lock.json file generated by npm – it's all `yarn` from now on.


mbelton-buildkite (approved):
> it's all yarn from now on

Nice!

PR #2068, opened 2023-04-21
Work around listen gem issue

dannymidnight:
There seems to be an issue with the listen gem running on linux/amd64 which prevents the app from running on docker. 

I found a [workaround](https://github.com/evilmartians/terraforming-rails/issues/34#issuecomment-1347442704) which seems fine since we're likely to stop using sprockets (https://github.com/buildkite/docs/pull/2064)


mbelton-buildkite (approved):
Worked for me locally!

PR #2069, opened 2023-04-21
Consolidate GraphQL nav into regular nav

dannymidnight:
No more nav switching. It's all the same nav!

Screenshot 2023-04-21 at 2 40 09 pm


mbelton-buildkite (approved):
Much more straightforward :)

PR #2080, opened 2023-04-24
Update source_control.md

stephanieatte:
We kept getting questions about commit status, we needed more clarification in the docs

### DIFF
--- a/pages/integrations/source_control.md
+++ b/pages/integrations/source_control.md
@@ -13,3 +13,5 @@ Buildkite integrates with several popular code repository management systems (RM
 * [Bitbucket Server](/docs/integrations/bitbucket-server)
 * [Phabricator](/docs/integrations/phabricator)
 * [Other Git servers](/docs/integrations/git)
+
+Note: Commit Status uses the latest built commit on a pipeline


pages/integrations/source_control.md
```diff
@@ -13,3 +13,5 @@ Buildkite integrates with several popular code repository management systems (RM
 * [Bitbucket Server](/docs/integrations/bitbucket-server)
 * [Phabricator](/docs/integrations/phabricator)
 * [Other Git servers](/docs/integrations/git)
+
+Note: Commit Status uses the latest built commit on a pipeline
```
mbelton-buildkite, line 17: Hi @stephanieatte, sorry this has taken me a while to get around to. Can you please unpack this a bit more for me? What is the issue customers are running into?

stephanieatte: @mbelton-buildkite they get confused about if the commit statuses feature works with the first of latest build on a pipeline

mbelton-buildkite: It feels a bit out of place here, what do you think about including it on the relevant pages? Thinking:

- GitHub + Enterprise: In the [first paragraph](https://buildkite.com/docs/integrations/github) of each page?
- Bitbucket: [Enable commit status updates](https://buildkite.com/docs/integrations/bitbucket#enable-commit-status-updates)

Looks like it's not supported or not mentioned in the others.

Perhaps just something like:

```
The commit status on a PR shows the result of the most recent build.
```

stephanieatte: @mbelton-buildkite I see but I didn't want to repeat the informtion, but I understand it looks out of place. we might have to put it on all the pages

PR #2086, opened 2023-04-26
Replace Sprockets with Vite

dannymidnight:
I wanted some live reloading in development so switched SASS compilation over to Vite (https://github.com/buildkite/docs/pull/2070).

Side note: the removal of some no-longer used gems also makes the docker image build a little faster. 

Edit: Turns out switching over images wasn't much additional work so I've fully replaced sprockets to remove the need for installing the sassc gem.


mbelton-buildkite:
I thought this might give me live reloading when editing markdown files, but looks like it's for other things? I also saw this error when starting with Docker:

1786eca4c6b7c52a9389c8e7d8ecaff2

I'm wondering whether to abandon the Docker workflow because it seems a lot slower now anyway.

dannymidnight:
Yeah I also spotted that 😭 . Looks like sprockets (the rails asset pipeline) still needs sass in development – even though it isn't being referenced.

MelissaKaulfuss (approved):
LGTM (at a surface level) 👍🏻✨

PR #2091, opened 2023-04-27
Fix JS error on home page

dannymidnight:
There's an error on the homepage that's caused by the table of contents JS not finding an element.


mbelton-buildkite (approved):
I'm assuming it's this error:

> c.js:54 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at e (toc.js:54:16)
    at zt (toc.js:61:3)
    at be (application.js:18:3)
    at application.js:13:3

If so, I see it's not there for this branch ✅

PR #2094, opened 2023-04-28
Add note that agent-startup hook is supported

nsuma8989:
This is to add a note that agent-startup hook is support from agent version 3.42.0 or above as we recently saw a user had issues and it turned out they are running on version older than v3.42.0 where hook did not run


pages/agent/v3/hooks.md
```diff
@@ -97,7 +97,7 @@ a plugin can define.
...
-| `agent-startup` | <span class="add-icon-agent">Agent</span> | Executed at agent startup, immediately prior to the agent being registered with Buildkite. Useful for initialising resources that will be used by all jobs that an agent runs, outside of the job lifecycle. |
+| `agent-startup` | <span class="add-icon-agent">Agent</span> | Executed at agent startup, immediately prior to the agent being registered with Buildkite. Useful for initialising resources that will be used by all jobs that an agent runs, outside of the job lifecycle. This is introduced from agent version v3.42.0 and above. |
```
mbelton-buildkite, line 100: I think it would be good to put this on a new line:

```suggestion
| `agent-startup` | <span class="add-icon-agent">Agent</span> | Executed at agent startup, immediately prior to the agent being registered with Buildkite. Useful for initialising resources that will be used by all jobs that an agent runs, outside of the job lifecycle.<br /><br />Supported from agent version 3.42.0 and above. |
```

What do you think?

nsuma8989: Yes, makes sense. Updated the change

mbelton-buildkite (approved):
Looks good!

PR #2096, opened 2023-04-28
Creator variables clarifications

butterkitty:
Updated the Pipeline-> Conditionals page to be clearer around the creator variables.

Also added notes around the requirement of the user to be verified for pipeline conditionals to be used properly. According to my testing, this is the case and wasn't implied at all in documentation.


pages/integrations/slack.md
```diff
@@ -25,6 +25,8 @@ Once you have granted access to your Slack workspace, give it a description, cho
 
 <%= image "buildkite-slack-connected.png", width: 1458/2, height: 1540/2, alt: "Screenshot of Buildkite Slack Notification Settings, requesting a description, your choice of text or emoji message themes, which pipelines and branches to include, and which build states should trigger a notification" %>
 
+The function of the "fixed" option displayed above is to notify you of a successful build of a pipeline after it has failed and will not be issued based on the actions of individual steps within a pipeline. The notifications configured above are at the pipeline level, not at the step level.
```
mbelton-buildkite, line 28: This reads as if the failed build option causes it to be at the pipeline level, but I suspect that's not the case from the wording in the UI.

If it is:
```suggestion
With the configuration above, you'll receive notifications at the pipeline level. The _fixed builds_ option ensures you're notified when a failed build next passes, but not on the outcomes of individual steps within a pipeline.
```

If it's not:
```suggestion
With the configuration above, you'll receive notifications at the pipeline level but not on the outcomes of individual steps. The _fixed builds_ option ensures you're notified when a failed build next passes.
```

mbelton-buildkite (commented):
I've left a few comments to consider. Thanks for picking this up!

pages/integrations/slack.md
butterkitty: I was trying to figure out where this code came from and I was really trying to think. Then I remembered this [PR](https://github.com/buildkite/docs/pull/1848) that was never looked at from months ago. This code actually belongs there, but if you're okay, we can merge it in here as well and close out the other PR.

"We should clarify what is meant by fixed as well as how it functions at the pipeline outcome level and not the step level."

So, it would be at the pipeline level, not the step level.

pages/pipelines/conditionals.md
```diff
@@ -197,22 +197,38 @@ The following variables are supported by the `if` attribute. Note that you canno
 	<tr>
 		<td><code>build.creator.email</code></td>
 		<td><code>String</code></td>
-		<td>The <strong><a href="#unverified-commits">unverified</a></strong> email of the user who created the build. </td>
+		<td>The email of the user who created the build. This variable is set several different ways depending on how the build was created:<br><br>
+    	* Buildkite UI - Set based on who manually created the build using the Buildkite UI <br>
+		* GitHub Webhook - Set from the <strong><a href="#unverified-commits">unverified</a></strong> HEAD commit. <br>
+		* Webhook - This will be the user attached to the API Key used.<br><br>
+  		<i>In each case, the user must be a verified Buildkite user, meaning their email must be verified, for pipeline conditionals including this variable to function.</i></td>
```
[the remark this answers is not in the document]
butterkitty: These look good. I'm used to working with markdown so I forgot to do up any html instead.

This will work a lot better thanks!

mbelton-buildkite (approved):
Looks good to me! Feel free to merge when you're ready :)

pages/integrations/slack.md
mbelton-buildkite: Yeah, let's just include it here. So you can close the other PR when this one is merged :)

PR #2097, opened 2023-04-28
Added note about the agent release in which `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` is introduced

nsuma8989:
We had a user who was trying to use `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` but it did not work as this feature was introduced in agent release v3.37.0 or above. This PR updates the pages to include a note that mentions release this configuration is introduced.

### DIFF
--- a/data/content/agent_attributes.yaml
+++ b/data/content/agent_attributes.yaml
@@ -234,7 +234,7 @@ attributes:
   default_value: "false"
   required: false
   desc: |
-      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development. Introduced in [v3.37](https://github.com/buildkite/agent/releases/tag/v3.37.0).
+      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development. Available from [v3.37](https://github.com/buildkite/agent/releases/tag/v3.37.0).
 - name: priority
   env_var: BUILDKITE_AGENT_PRIORITY
   default_value: "null"
--- a/data/content/agent_attributes.yaml
+++ b/data/content/agent_attributes.yaml
@@ -234,7 +234,7 @@ attributes:
   default_value: "false"
   required: false
   desc: |
-      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development.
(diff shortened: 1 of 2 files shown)


nsuma8989:
@mbelton-buildkite Please can we review this PR as we had two users report issues where it was because this feature was introduced in v3.37.0 or later and in our docs there is no note about it

pages/plugins/writing.md
```diff
@@ -238,7 +238,7 @@ Next, add a `README.md` file to introduce the plugin to the world:
...
 
-One way to try this is to add the following step to the Buildkite pipeline where you're testing your plugin.  Configuring `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` on only one step means that other plugins, which are unlikely to be changing in the meantime, won't get unnecessarily cloned on every step invocation.
+One way to try this is to add the following step to the Buildkite pipeline where you're testing your plugin.  Configuring `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` on only one step means that other plugins, which are unlikely to be changing in the meantime, won't get unnecessarily cloned on every step invocation. This needs agent version v3.37.0 or above as configuration `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` is introduced in that release.
```
mbelton-buildkite, line 241: Just a bit simpler:

```suggestion
One way to try this is to add the following step to the Buildkite pipeline where you're testing your plugin.  Configuring `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` on only one step means that other plugins, which are unlikely to be changing in the meantime, won't get unnecessarily cloned on every step invocation. You need agent version v3.37.0 or above to use `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH`.
```

data/content/agent_attributes.yaml
```diff
@@ -234,7 +234,7 @@ attributes:
   default_value: "false"
   required: false
   desc: |
-      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development.
+      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development. Introduced in [v3.37](https://github.com/buildkite/agent/releases/tag/v3.37.0).
```
mbelton-buildkite, line 237:
```suggestion
      Force checking out a fresh clone of plugins every build. By default, if the organization, repository, and version tag of a plugin specified in a step match a plugin that is already on the agent, the agent uses that local version. Forcing fresh checkout is especially useful during plugin development. Available from [v3.37](https://github.com/buildkite/agent/releases/tag/v3.37.0).
```

mbelton-buildkite (approved):
Just a couple of nitpicks you can consider, but looks good 😊

PR #2102, opened 2023-05-02
Generate sitemap

dannymidnight:
Add configuration to generate sitemap for /docs.

TODO:

- [x] ~~Check paths are right~~
- [x] ~~Generate sitemap as part of CI~~
- [ ] Connect up to marketing sitemap (https://github.com/buildkite/site/pull/2417)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 .yarn
 
 /public/assets
+/public/docs/sitemap.xml.gz
 .byebug_history
 
 .DS_Store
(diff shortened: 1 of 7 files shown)


dannymidnight:
OK cool, had a moment to come back around on this. I'm now generating the sitemap during a docker build. Longer term I'd prefer this to happen as a seperate step in CI but this kind of follows suit with the current CI approach for docs. If we were to build this separately we may also consider hosting the sitemap in a shared S3 bucket.

For now I think hosting under https://buildkite.com/docs/sitemap.xml.gz is fine and the output seems like a decent starting point – we can always tweak priority / changefreq later.

PR #2106, opened 2023-05-03
Simplify nav generation

dannymidnight:
The current workflow for modifying the nav is a little clunky. This PR wires up the nav data on boot so that now there's no requirement to regenerate the nav when moving around items. 

The GraphQL generation script now generates a file with _only_ the GraphQL items rather than a consolidated nav.yml file.


mbelton-buildkite (approved):
I tried the workflow locally and looks good. I also pushed a change with some basic instructions in the readme (feel free to update).

PR #2108, opened 2023-05-03
Clarification on pipeline upload order

pzeballos:
When using multiple dynamic pipelines from the same step, these are shown in reverse order in the UI (i.e. steps uploaded in the second call to pipeline upload show up before the first call). This is confusing for some people and there's a workaround if they want to see them in the "expected" order

### DIFF
--- a/pages/pipelines/defining_steps.md
+++ b/pages/pipelines/defining_steps.md
@@ -260,7 +260,11 @@ To use this script, you'd save it to `.buildkite/pipeline.sh` inside your reposi
 .buildkite/pipeline.sh | buildkite-agent pipeline upload
 ```
 
-When the build is running it will execute the script and pipe the output to the `pipeline upload` command. The upload command will insert the steps from the script into the build immediately after the upload step.
+When the build is running it will execute the script and pipe the output to the `pipeline upload` command. The upload command will insert the steps from the script into the build immediately after the upload step, which means that if you execute it more than once in a single step you'll get those uploads steps appearing in reverse order.
+
+>📘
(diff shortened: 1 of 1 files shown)


pages/pipelines/defining_steps.md
```diff
@@ -260,7 +260,11 @@ To use this script, you'd save it to `.buildkite/pipeline.sh` inside your reposi
 .buildkite/pipeline.sh | buildkite-agent pipeline upload
 ```
 
-When the build is running it will execute the script and pipe the output to the `pipeline upload` command. The upload command will insert the steps from the script into the build immediately after the upload step.
+When the build is running it will execute the script and pipe the output to the `pipeline upload` command. The upload command will insert the steps from the script into the build immediately after the upload step, which means that if you execute it more than once in a single step you'll get those uploads steps appearing in reverse order.
+
+>📘
+> To avoid having the steps in a reverse order in the UI, we suggest to upload the steps in the opposite order you expect them (the things you want to run immediately last).
```
mbelton-buildkite, line 266: Let's encapsulate the whole ordering situation into the note:

```suggestion
When the build runs, it executes the script and pipes the output to the `pipeline upload` command. The upload command then inserts the steps from the script into the build immediately after the upload step.

>📘 Step ordering
> Since the upload command inserts steps immediately after the upload step, they appear in reverse order when you upload multiple steps in one command. To avoid the steps appearing in reverse order, we suggest you upload the steps in reverse order (the step you want to run first goes last). That way, they'll be in the expected order when inserted.
```

What do you think? (I've also added a title to the note and fixed up the paragraph style a bit)

mbelton-buildkite (approved):

PR #2116, opened 2023-05-09
Update docs on Elastic CI Stack Parameters to v5.20

triarius:
The documentation for the elastic ci stack parameters was derived from a vendored stack template, which needs to be manually updated. We had not done this in a while, and we had added a few new parameters since then.

Also, the page was attempting to render everything between `:` as emojis. As `:aws:` is a valid Buildkite emoji, if any stack parameter value were an ARN, it would render part of it as an emoji. It's necessary to do this as there is as a new stack parameter default value that is `arn:aws:serverlessrepo:us-east-1:172840064832:applications/buildkite-agent-scaler`, which was being rendered as in the screenshot below prior to this change.

2023-05-10T07:30:48,098543098+10:00

Now, it is rendered as:
2023-05-10T07:32:34,170521909+10:00

### DIFF
--- a/data/content/aws-stack.yml
+++ b/data/content/aws-stack.yml
@@ -1,6 +1,25 @@
 ---
 AWSTemplateFormatVersion: "2010-09-09"
-Description: "Buildkite stack v5.7.0"
+Description: "Buildkite stack v5.20.0"
+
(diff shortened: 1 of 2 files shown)


mbelton-buildkite (approved):
Thanks for updating these!

PR #2118, opened 2023-05-11
Fix link to `annotate` subcommand in annotation API doc page

nick96:
Just ran into this - it points to the [`pipeline` subcommand](https://buildkite.com/docs/agent/v3/cli-pipeline) when it should point to the [`annotate` subcommand](https://buildkite.com/docs/agent/v3/cli-annotate).

### DIFF
--- a/pages/apis/rest_api/annotations.md
+++ b/pages/apis/rest_api/annotations.md
@@ -3,7 +3,7 @@
 
 ## Annotation data model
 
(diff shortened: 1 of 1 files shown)


mbelton-buildkite (approved):
@nick96, thanks for catching this and fixing it! I'll just run the linting against this branch and merge for you when that passes.

PR #2124, opened 2023-05-18
Add resiliency to curl commands

ChrisBr:
It's always good to add retries and a timeout in CI to make the command more resilient.


mbelton-buildkite:
Hi @ChrisBr, thanks for raising this! I'm the tech writer looking after these docs. I've been considering whether this fits in the docs and think the answer is no. It is a good practice, but it is more of a good practice for using curl rather than using Buildkite. These code snippets aim to show how to use the Buildkite-specific parameters to make a call. Additional practices for using curl are out of scope.

Thanks again for raising this, though. It helped me clarify the scope for this in the docs. Let me know if you think I'm missing something. 😊

ChrisBr:
Fine by me!

PR #2125, opened 2023-05-18
Add documentation around prefix third-party fork branch names

nsuma8989:
This add information on how users can prefix third-party fork branch names

### DIFF
--- a/pages/integrations/github.md
+++ b/pages/integrations/github.md
@@ -78,6 +78,8 @@ Optionally, select one or more of the following:
 * _Rebuild pull requests when they become ready for review_
 * _Build pull requests from third-party forked repositories_. Make sure to check the [managing secrets](/docs/pipelines/secrets) guide if you choose to do this.
 
+If you want to prefix branch names for third-party forked builds to ensure they don't trigger branch conditions then you need to select _Prefix third-party fork branch names_.
+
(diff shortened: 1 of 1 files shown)


pages/integrations/github.md
```diff
@@ -78,6 +78,8 @@ Optionally, select one or more of the following:
 * _Rebuild pull requests when they become ready for review_
 * _Build pull requests from third-party forked repositories_. Make sure to check the [managing secrets](/docs/pipelines/secrets) guide if you choose to do this.
 
+If you want to prefix branch names for third-party forked builds to ensure they don't trigger branch conditions then you need to select _Prefix third-party fork branch names_.
```
mbelton-buildkite, line 81: I think it would be good to front-load the problem we're trying to solve (ensuring third-party forked builds don't trigger branch conditions). I also think we should include where someone finds the option to select. Thinking something like:

```suggestion
If you want to control which third-party forks can trigger builds in GitHub, you can prefix the branches from third-party forks with the contributor's username. For example, the `main` branch from `some-user` becomes `some-user:main`. You can then detect these using a pre-command hook or something similar before running a build. To enable prefixing the branch names, go to the GitHub settings for the pipeline and select _Prefix third-party fork branch names_.
```


What do you think? I've written it as specific to GitHub, but is it?

nsuma8989:
@mbelton-buildkite Thank you for the feedback. I accepted your suggestion.

mbelton-buildkite (approved):

PR #2127, opened 2023-05-19
Add Pausing Cluster Queues documentation

l-suzuki:
This PR is to get the ball rolling on the documentation for Pausing Cluster Queues. This is an upcoming feature that allows users to "Pause" a cluster queue. When paused, job dispatch to agents associated with that queue is prevented. When resumed, job dispatch picks up where it left off. At first, pausing queues will be possible via the Buildkite UI and GraphQL API only. We will be adding REST API support for this as part of upcoming Clusters work later on.

For this first draft I've made a 'best guess' at what to include, given the existing structure of the Clusters docs. I'm sure there's lots of room for improvement, so would love feedback on this.

### DIFF
--- a/pages/agent/clusters.md
+++ b/pages/agent/clusters.md
@@ -53,19 +53,6 @@ To create additional queues:
 1. Enter a key and description.
 1. Select _Create Queue_.
 
-### Pause a queue
-
-Pausing a queue will prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
-
-To pause a queue:
-1. Navigate to your cluster’s _Queues_.
-1. Click on the queue you wish to pause.
-1. Click _Edit_.
-1. Under _Queue Management_, click _Pause Queue_.
-1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue.
-
-To resume the queue again, click Resume Queue.
-
 ### Connect agents to a cluster
 
 Agents are associated with a cluster through the cluster’s agent tokens.
@@ -102,6 +89,23 @@ To add a maintainer to a cluster:
 1. Navigate to the cluster’s _Maintainers_.
 1. Select a user or team.
 1. Click _Add Maintainer_.
+
+### Pause a queue
+
+You can pause a queue to prevent jobs from being dispatched to agents associated with that queue.
+To pause a queue:
+
+1. Navigate to your cluster’s _Queues_.
+1. Select on the queue you wish to pause.
+1. Select _Edit_.
+1. Under _Queue Management_, select _Pause Queue_.
+1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue. The note will be displayed on the queue page, and any affected builds.
+
+Jobs which have _already_ been dispatched to agents in the queue prior to pausing will continue to run. New jobs which target the paused queue will 'wait' until the queue is resumed.
+
+Trigger steps do not rely on agents, so they will run unless they have a dependency that has been halted by the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will 'wait' until the queue is resumed. If a triggered job does not target the pasued queue, it will run as usual.
+
+To resume the queue again, select _Resume Queue_. Once resumed, job dispatch to the queue will operate as usual, and any 'waiting' jobs affected by the pause will be picked up.
 
 ### Migrate to clusters
 
--- a/pages/agent/clusters.md
+++ b/pages/agent/clusters.md
@@ -90,7 +90,11 @@ To add a maintainer to a cluster:
 1. Select a user or team.
 1. Click _Add Maintainer_.
 
-### Pause a queue
+### Migrate to clusters
+
+If you migrate all your existing agents over to clusters, make sure to add all your pipelines to the relevant clusters. Otherwise, any builds for those pipelines will never find agents to run them.
+
+## Pause a queue
 
 You can pause a queue to prevent jobs from being dispatched to agents associated with that queue.
 
@@ -109,7 +113,3 @@ Jobs _already_ dispatched to agents in the queue before pausing will continue to
 Trigger steps do not rely on agents, so they will run unless they have dependencies waiting on the paused queue. The behavior of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will wait until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
 
 Select _Resume Queue_ to resume a paused queue. Jobs will resume dispatching to the queue as usual, including any jobs waiting to run.
-
-### Migrate to clusters
-
-If you migrate all your existing agents over to clusters, make sure to add all your pipelines to the relevant clusters. Otherwise, any builds for those pipelines will never find agents to run them.
--- a/pages/agent/clusters.md
+++ b/pages/agent/clusters.md
@@ -106,7 +106,7 @@ To pause a queue:
 
 Jobs _already_ dispatched to agents in the queue before pausing will continue to run. New jobs that target the paused queue will wait until the queue is resumed.
 
-Trigger steps do not rely on agents, so they will run unless they have dependencies waiting on the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will wait until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
+Trigger steps do not rely on agents, so they will run unless they have dependencies waiting on the paused queue. The behavior of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will wait until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
 
 Select _Resume Queue_ to resume a paused queue. Jobs will resume dispatching to the queue as usual, including any jobs waiting to run.
 
--- a/pages/agent/clusters.md
+++ b/pages/agent/clusters.md
@@ -93,19 +93,22 @@ To add a maintainer to a cluster:
 ### Pause a queue
 
 You can pause a queue to prevent jobs from being dispatched to agents associated with that queue.
+
 To pause a queue:
 
 1. Navigate to your cluster’s _Queues_.
-1. Select on the queue you wish to pause.
+1. Select the queue you wish to pause.
 1. Select _Edit_.
 1. Under _Queue Management_, select _Pause Queue_.
-1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue. The note will be displayed on the queue page, and any affected builds.
+1. Enter an optional note in the dialog, and confirm that you wish to pause the queue.
 
-Jobs which have _already_ been dispatched to agents in the queue prior to pausing will continue to run. New jobs which target the paused queue will 'wait' until the queue is resumed.
+      You can use the note to explain why you're pausing the queue. The note will display on the queue page and any affected builds.
 
-Trigger steps do not rely on agents, so they will run unless they have a dependency that has been halted by the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will 'wait' until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
+Jobs _already_ dispatched to agents in the queue before pausing will continue to run. New jobs that target the paused queue will wait until the queue is resumed.
 
-To resume the queue again, select _Resume Queue_. Once resumed, job dispatch to the queue will operate as usual, and any 'waiting' jobs affected by the pause will be picked up.
+Trigger steps do not rely on agents, so they will run unless they have dependencies waiting on the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will wait until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
+
+Select _Resume Queue_ to resume a paused queue. Jobs will resume dispatching to the queue as usual, including any jobs waiting to run.
 
 ### Migrate to clusters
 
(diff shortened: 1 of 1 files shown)


pages/agent/clusters.md
```diff
@@ -53,6 +53,19 @@ To create additional queues:
...
+### Pause a queue
+
+Pausing a queue will prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
+
+To pause a queue:
+1. Navigate to your cluster’s _Queues_.
+1. Click on the queue you wish to pause.
+1. Click _Edit_.
+1. Under _Queue Management_, click _Pause Queue_.
+1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue.
```
angusfretwell, line 65: I think we should explain that a notice including the note will be shown on build pages affected by the pause.

l-suzuki: Good point - feel like we should mention this in the dialog itself too

pages/agent/clusters.md
```diff
@@ -53,6 +53,19 @@ To create additional queues:
 1. Enter a key and description.
 1. Select _Create Queue_.
 
+### Pause a queue
+
+Pausing a queue will prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
+
+To pause a queue:
+1. Navigate to your cluster’s _Queues_.
```
timriley, line 61:
```suggestion
To pause a queue:

1. Navigate to your cluster’s _Queues_.
```

I think we need this extra newline to make sure the list displays properly. Right now it's not being picked up (from the branch preview site):

Screenshot 2023-05-22 at 1 23 16 pm

pages/agent/clusters.md
```diff
@@ -53,6 +53,19 @@ To create additional queues:
...
+1. Click on the queue you wish to pause.
+1. Click _Edit_.
+1. Under _Queue Management_, click _Pause Queue_.
+1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue.
+
+To resume the queue again, click Resume Queue.
```
timriley, line 67:
```suggestion
To resume the queue again, select _Resume Queue_.
```

We're italicising the other bits of UI text here, so this one probably needs it too?

Also, we use "select" instead of "click" as our house style for directly interacting with buttons and other UI elements.

pages/agent/clusters.md
```diff
@@ -53,6 +53,19 @@ To create additional queues:
...
 
+### Pause a queue
+
+Pausing a queue will prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
+
+To pause a queue:
+1. Navigate to your cluster’s _Queues_.
+1. Click on the queue you wish to pause.
+1. Click _Edit_.
+1. Under _Queue Management_, click _Pause Queue_.
```
timriley, line 64:
```suggestion
1. Select the queue you wish to pause.
1. Select _Edit_.
1. Under _Queue Management_, select _Pause Queue_.
```

Just a few other "click"->"select" changes.

pages/agent/clusters.md
```diff
@@ -53,6 +53,19 @@ To create additional queues:
 1. Enter a key and description.
 1. Select _Create Queue_.
 
+### Pause a queue
+
+Pausing a queue will prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
```
timriley, line 58:
```suggestion
You can pause a queue to prevent jobs from being dispatched to agents associated with that queue. This may affect builds targeting the paused queue.
```

Just tossing around the idea of phrasing this in "You can" language like a few of the other sentences on this page, given this is an action we expect our users to take directly.

timriley: I also wonder if we can make this sentence clearer?

> This may affect builds targeting the paused queue.

Does this mean trigger jobs from other pipelines targeting the queue to be paused? If so, perhaps we could include that specific detail too?

timriley (commented):
Nice work getting the docs ready, @l-suzuki!

I took a pass at this and shared some things that I learnt from the review of my own first contribution to our docs. Let me know what you think :)

I do have one other question about the _placement_ of this new section. Right now we have it within the "Set up" section, _before_ the section about connecting agents to the cluster. Since this new section is all about impacting whether jobs make it to the agents or not, I wonder if it should come somewhere _after_ the agents section?

Right now our sections under "Set up a cluster" are:

- Set up queues
- **Pause a queue**
- Connect agents to a cluster
- Add pipelines to a cluster
- Add maintainers to a cluster
- Migrate to clusters

Since it only makes sense to pause a queue after everything is already fully set up, I wonder if this order might be better?

- Set up queues
- Connect agents to a cluster
- Add pipelines to a cluster
- Add maintainers to a cluster
- **Pause a queue**
- Migrate to clusters

What do you reckon? I'm don't feel super strongly about this, but I thought it might be worth some quick consideration while we're still getting this change in place.

The upside of this new order would be that pausing comes after all the other sections related to getting properly set up for the first time. The downside is that our two sections dedicated specifically to queues are now further apart.

pages/agent/clusters.md
l-suzuki: @timriley yeah, I wasn't sure how much detail to go into here! With trigger jobs, they don't run on agents at all so don't really care about queues being paused. If the job they're triggering relies on the paused queue, it won't run while the queue is paused. If the job they're triggering targets another queue that is not paused, then it'll go ahead and run, unless an explicit dependency has been set

pages/agent/clusters.md
l-suzuki: whoops!

l-suzuki:
> Nice work getting the docs ready, @l-suzuki!
> 
> I took a pass at this and shared some things that I learnt from the review of my own first contribution to our docs. Let me know what you think :)
> 
> I do have one other question about the _placement_ of this new section. Right now we have it within the "Set up" section, _before_ the section about connecting agents to the cluster. Since this new section is all about impacting whether jobs make it to the agents or not, I wonder if it should come somewhere _after_ the agents section?
> 
> Right now our sections under "Set up a cluster" are:
> 
> * Set up queues
> * **Pause a queue**
> * Connect agents to a cluster
> * Add pipelines to a cluster
> * Add maintainers to a cluster
> * Migrate to clusters
> 
> Since it only makes sense to pause a queue after everything is already fully set up, I wonder if this order might be better?
> 
> * Set up queues
> * Connect agents to a cluster
> * Add pipelines to a cluster
> * Add maintainers to a cluster
> * **Pause a queue**
> * Migrate to clusters
> 
> What do you reckon? I'm don't feel super strongly about this, but I thought it might be worth some quick consideration while we're still getting this change in place.
> 
> The upside of this new order would be that pausing comes after all the other sections related to getting properly set up for the first time. The downside is that our two sections dedicated specifically to queues are now further apart.

Thanks for the in depth reviews @timriley  - with the order of things, I went back and forth on this for ages:  should it belong next to the "Set up a queue" because its queue related? or should it go later because of the reasons you said? In the end I just picked one and thought I'd see how people responded. Looking at it again now I think what you suggest makes more sense.  The Clusters docs are currently structured more like a setup guide so "chronological" ordering makes sense. If Clusters docs evolve into a bigger beast with sub sections etc etc then I'd expect this info to go in a Clusters > Cluster Queues subsection, but we're not at that stage yet :)

pages/agent/clusters.md
l-suzuki: I've tried a best pass at explaining in more detail. Let me know what you think :)

pages/agent/clusters.md
```diff
@@ -90,6 +90,23 @@ To add a maintainer to a cluster:
...
+
+1. Navigate to your cluster’s _Queues_.
+1. Select on the queue you wish to pause.
+1. Select _Edit_.
+1. Under _Queue Management_, select _Pause Queue_.
+1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue. The note will be displayed on the queue page, and any affected builds.
+
+Jobs which have _already_ been dispatched to agents in the queue prior to pausing will continue to run. New jobs which target the paused queue will 'wait' until the queue is resumed.
+
+Trigger steps do not rely on agents, so they will run unless they have a dependency that has been halted by the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will 'wait' until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
```
timriley, line 106: Also a good clarifying note!

Just took the quotes off "wait" here too:

```suggestion
Trigger steps do not rely on agents, so they will run unless they have a dependency that has been halted by the paused queue. The behaviour of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will wait until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
```

My only other thought here is whether "halted" is the right term. We're not using it anywhere else in this section. I wonder if we could rephrase that first sentence to avoid it...

> Trigger steps do not rely on agents, so they will run unless they have dependencies that are also waiting with the paused queue.

Maybe?

pages/agent/clusters.md
```diff
@@ -90,6 +90,23 @@ To add a maintainer to a cluster:
...
+
+To resume the queue again, select _Resume Queue_. Once resumed, job dispatch to the queue will operate as usual, and any 'waiting' jobs affected by the pause will be picked up.
```
timriley, line 108:
```suggestion
Select _Resume Queue_ to resume a paused queue. Job dispatch to the queue will operate as usual and any previously waiting jobs will be run.
```

timriley (commented):
Great improvements here, @l-suzuki! I have another round of feedback and I promise this'll be my last :)

pages/agent/clusters.md
```diff
@@ -90,6 +90,23 @@ To add a maintainer to a cluster:
...
+### Pause a queue
+
+You can pause a queue to prevent jobs from being dispatched to agents associated with that queue.
+To pause a queue:
+
+1. Navigate to your cluster’s _Queues_.
+1. Select on the queue you wish to pause.
+1. Select _Edit_.
+1. Under _Queue Management_, select _Pause Queue_.
+1. Enter an optional note in the dialog if needed, and confirm that you wish to pause the queue. The note will be displayed on the queue page, and any affected builds.
```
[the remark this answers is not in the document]
mbelton-buildkite: I think the "if needed" might be playing another role. For me, it's hinting at the reason why you might or might not leave a note. Maybe it would be helpful to be more explicit about that? Thinking:

```suggestion
1. Enter an optional note in the dialog, and confirm that you wish to pause the queue. 
   
    You can use the note to explain why you're pausing the queue. The note will display on the queue page and any affected builds.
```

Note, the spacing should display the clarifying sentence indented with the list. I think I've got the spacing right, but worth double checking.

pages/agent/clusters.md
```diff
@@ -90,6 +90,23 @@ To add a maintainer to a cluster:
...
+
+Jobs which have _already_ been dispatched to agents in the queue prior to pausing will continue to run. New jobs which target the paused queue will 'wait' until the queue is resumed.
```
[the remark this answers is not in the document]
mbelton-buildkite: This could be further tightened:

```suggestion
Jobs _already_ dispatched to agents in the queue before pausing will continue to run. New jobs that target the paused queue will wait until the queue is resumed.
```

But, being more verbose could help with clarity. You can decide which is the better way to go here. :)

pages/agent/clusters.md
mbelton-buildkite: +1 to avoiding adding a new term here if we can avoid it. Another option for tightening:

> Trigger steps do not rely on agents, so they will run unless they have dependencies waiting on the paused queue.

mbelton-buildkite, line 106: Use US English:

```suggestion
Trigger steps do not rely on agents, so they will run unless they have a dependency that has been halted by the paused queue. The behavior of the jobs they trigger depends on their configuration. If a triggered job targets the paused queue, it will 'wait' until the queue is resumed. If a triggered job does not target the paused queue, it will run as usual.
```

pages/agent/clusters.md
mbelton-buildkite: I'll add a further suggestion to the second sentence:

```suggestion
Select _Resume Queue_ to resume a paused queue. Jobs will resume dispatching to the queue as usual, including any jobs waiting to run.
```

mbelton-buildkite (commented):
Nice! I've left a few suggestions, but I think this is looking good. 😊

mbelton-buildkite (commented):
Nice! I've left a few suggestions, but I think this is looking good. 😊

mbelton-buildkite:
One larger change in my mind that I'd like both of your opinions on is the location of this page. I misunderstood clusters when this was first being worked on and thought they were specific to agents. I now understand that they're a container concept above agents and pipelines. With this much content, perhaps it's a good chance to create a dedicated section. I've put this into a [quick PR to see the idea](https://github.com/buildkite/docs/pull/2137/files).

Ignoring this is okay because it's no worse than now, but I think it would be a good change. What do y'all think?

(Note, don't merge that change because it doesn't include the suggested changes)

l-suzuki:
@mbelton-buildkite @timriley thanks for your feedback, I've made the suggested changes, and I think it's displaying correctly here: 

Screenshot 2023-05-24 at 2 42 26 PM

mbelton-buildkite (approved):
Nice! This looks good. I'll follow up about restructuring the section.

l-suzuki:
@mbelton-buildkite I've just been doing a final check of this - I'm wondering whether "Pause a queue" should actually be an h2 after "migrate to clusters". My reason for asking is because the h3s up until "migrate to clusters" are all steps in the setup process. "Pause a queue" isn't part of setup - it's a feature you'd use *after* you've set things up, probably once you'd gotten the hang of clusters. What do you think?

mbelton-buildkite:
> I'm wondering whether "Pause a queue" should actually be an h2 after "migrate to clusters"

That sounds like a good idea to me, @l-suzuki 👍🏻

PR #2129, opened 2023-05-19
Add metrics-datadog-distributions documentation

abhishekmukherg:
This adds documentation for metrics-datadog-distributions, which was added in https://github.com/buildkite/agent/pull/1433.

Please note that I don't actually know how to test this at all -- do I get a rendered version of the site in the PR?

### DIFF
--- a/data/content/agent_attributes.yaml
+++ b/data/content/agent_attributes.yaml
@@ -154,6 +154,16 @@ attributes:
   required: false
   desc: |
         The DogStatsD instance to send metrics to using UDP.
+- name: metrics-datadog-distributions
+  env_var: BUILDKITE_METRICS_DATADOG_DISTRIBUTIONS
+  default_value: "false"
+  requires: false
+  desc: |
+      Use Datadog Distributions for Timing metrics. By default, we submit metrics via standard gauges. This will unfortunately cause issues if you have multiple agents running,
+      as components like buildkite.jobs.duration.success.median becomes a roll up of the median execution time _from each agent_. As each agent is likely only running a single job
+      at a time, this causes all time related histogram metrics (avg, max, median, 95percentile) from a given agent to be the same value.
+      Distributions, however, aggregate the metrics server side at Datadog, thus the median and 95percentile is calculated given the set of builds executing across all agents.
+      Learn more about [Datadog Distributions](Use Datadog Distributions for Timing metrics)
 - name: name
   env_var: BUILDKITE_AGENT_NAME
   default_value: "%hostname-%spawn"
--- a/data/content/agent_attributes.yaml
+++ b/data/content/agent_attributes.yaml
@@ -159,7 +159,7 @@ attributes:
   default_value: "false"
   requires: false
   desc: |
-      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This advisable when running multiple agents. Otherwise, metrics from multiple agents may be rolled up and appear to have the same value.
+      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This is advisable when running multiple agents. Otherwise, metrics from multiple agents may be rolled up and appear to have the same value.
 - name: name
   env_var: BUILDKITE_AGENT_NAME
   default_value: "%hostname-%spawn"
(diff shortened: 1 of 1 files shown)


abhishekmukherg:
My concern on this PR is that maybe i gave _too_ much information and should just link out to the Datadog docs

data/content/agent_attributes.yaml
```diff
@@ -154,6 +154,16 @@ attributes:
...
+- name: metrics-datadog-distributions
+  env_var: BUILDKITE_METRICS_DATADOG_DISTRIBUTIONS
+  default_value: "false"
+  requires: false
+  desc: |
+      Use Datadog Distributions for Timing metrics. By default, we submit metrics via standard gauges. This will unfortunately cause issues if you have multiple agents running,
+      as components like buildkite.jobs.duration.success.median becomes a roll up of the median execution time _from each agent_. As each agent is likely only running a single job
+      at a time, this causes all time related histogram metrics (avg, max, median, 95percentile) from a given agent to be the same value.
+      Distributions, however, aggregate the metrics server side at Datadog, thus the median and 95percentile is calculated given the set of builds executing across all agents.
+      Learn more about [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types)
```
mbelton-buildkite, line 166: As you mentioned, I think this is a bit too much information. I see that most values match the descriptions given in the [command references](https://buildkite.com/docs/agent/v3/cli-start#metrics-datadog-distributions), but adding a sentence for why someone would use this is helpful. Thinking:

```suggestion
      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This is helpful when running multiple agents as Datadog aggregates metrics server side.
```

What do you think?

mbelton-buildkite (commented):
Hi @abhishekmukherg, thanks for putting this together!

> Please note that I don't actually know how to test this at all -- do I get a rendered version of the site in the PR?

You can run the site locally if you'd like using the instructions in the [README](https://github.com/buildkite/docs#run-the-development-server). I've done this and confirmed it looks good, though. So all good to skip it.

I'll work with you on the language for this change and add someone who works on the agent to provide a technical review. Once it looks good, I'll run the linting on the change and merge for you. 😊

data/content/agent_attributes.yaml
triarius: I think the disclaimer is still useful to have. Perhaps, it's needed elsewhere too, but we can phrase it here like
```suggestion
      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This advisable when running multiple agents. Otherwise, metrics from multiple agents may be rolled up and appear to have the same value.
```

triarius (commented):
From what you've described @amukherjeetwilio, the metrics are not useful unless this option is set to true. I would want users encountering that problem to be able to find this page while searching for the symptoms.

It also might be a good idea to change this to be true by default in some future version of the agent.

abhishekmukherg:
> From what you've described @amukherjeetwilio, the metrics are not useful unless this option is set to true. I would want users encountering that problem to be able to find this page while searching for the symptoms.

That was our perspective too, yes, which is why we added this feature to the `buildkite-agent`. _Some_ of the existing metrics without this are still useful, specifically the mean and mode, I think, will work. The trade off is that distributions are _significantly_ costlier

triarius:
> The trade off is that distributions are _significantly_ costlier

Ahh ok, makes sense to leave it `false` by default then.

triarius (approved):
LGTM 🙇‍♂️

data/content/agent_attributes.yaml
```diff
@@ -154,6 +154,12 @@ attributes:
   required: false
   desc: |
         The DogStatsD instance to send metrics to using UDP.
+- name: metrics-datadog-distributions
+  env_var: BUILDKITE_METRICS_DATADOG_DISTRIBUTIONS
+  default_value: "false"
+  requires: false
+  desc: |
+      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This advisable when running multiple agents. Otherwise, metrics from multiple agents may be rolled up and appear to have the same value.
```
mbelton-buildkite, line 162: Typo: Missing word

```suggestion
      Use [Datadog Distributions](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types) for timing metrics. This is advisable when running multiple agents. Otherwise, metrics from multiple agents may be rolled up and appear to have the same value.
```

mbelton-buildkite (approved):
I've just left one suggestion to add a missing word. Since it's a small typo, I'll accept the change and run the linting.

data/content/agent_attributes.yaml
```diff
@@ -154,6 +154,12 @@ attributes:
   required: false
   desc: |
         The DogStatsD instance to send metrics to using UDP.
+- name: metrics-datadog-distributions
+  env_var: BUILDKITE_METRICS_DATADOG_DISTRIBUTIONS
+  default_value: "false"
+  requires: false
```
mbelton-buildkite, line 160: The linting picked up a typo here with the field name:

```suggestion
  required: false
```

mbelton-buildkite:
This is good to go, so I'll merge it for you. Thanks again for putting this together, @abhishekmukherg! And thanks for your help @triarius 😊

PR #2130, opened 2023-05-19
Added how to get number of builds between two dates

butterkitty:


pages/apis/graphql/graphql_cookbook.md
```diff
@@ -906,3 +906,25 @@ mutation UpdateTeamPipelineReadonly {
   }
 }
 ```
+
+## Test Analytics
+
+### Get number of builds between 2 dates
```
mbelton-buildkite, line 912: Style nitpicks: 

- Write out number for 0-9.
- Add a blank line between headings and paragraphs.

```suggestion
### Get number of builds between two dates
 
```

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -906,3 +906,25 @@ mutation UpdateTeamPipelineReadonly {
   }
 }
 ```
+
+## Test Analytics
+
+### Get number of builds between 2 dates 
+There is a correlation between number of builds and job minutes used. Though this query won't get you how many job minutes have been used, but it will give an idea of whether or not there is an increase in the number of builds, which will cause more job minutes to be used.
```
mbelton-buildkite, line 913: What's the main task that people are trying to do here? Is it to look for usage trends and this is the recommended way? If so, maybe the heading should be something like, _Check usage in a period_ and the text:

```suggestion
This query helps you understand how many job minutes you've used by looking at the number of builds. While not equivalent, there's a correlation between the number of builds and job minutes. So, looking at the number of builds in different periods gives you an idea of how the job minutes would compare in those periods.
```

Also, is this connected to TA? I thought job minutes were a Pipelines thing? And the equivalent in TA is test executions?

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -906,3 +906,25 @@ mutation UpdateTeamPipelineReadonly {
...
+
+```graphql
+query PipelineBuildCountForBranchQuery {
```
mbelton-buildkite, line 916: I think this is a copy/paste hangover, so I think it should be updated. Perhaps:

```suggestion
query PipelineBuildCountForPeriod {
```

pages/apis/graphql/graphql_cookbook.md
butterkitty: You are correct! For some reason I was thinking TA, but it is under the pipeline subheading in the rest of the documentation. I blame the Friday evening slump.

I'll correct this!

mbelton-buildkite (commented):
I think the content is all good, but the position is not quite right. Since this is more about builds than pipelines, how about moving it under the `## Builds` heading?

butterkitty:
> I think the content is all good, but the position is not quite right. Since this is more about builds than pipelines, how about moving it under the `## Builds` heading?

it's true that it does have a lot to do with builds, but the usage is centered around job minutes, so I'm not sure if it's in builds or pipelines. I can see the utility for both. I can move it to Builds, if you feel it's better there. TBH, I'm not massively attached to it. 

I would end up changing the wording on the explanation a bit though if we do that. Maybe something like, this will get the total number of builds between two dates. This can also be used for job minutes blah blah blah, etc etc...

:slightly_smiling_face:

mbelton-buildkite (approved):

PR #2137, opened 2023-05-23
Idea: Add a Clusters section

mbelton-buildkite:
**DO NOT MERGE**

This adds a new section for Clusters content and splits the existing page into two.

I misunderstood clusters when this was first being worked on and thought they were specific to agents. I now understand that they're a container concept above agents and pipelines.


l-suzuki:
@mbelton-buildkite This idea makes a lot of sense to me. Especially given that the idea is to make Clusters a 'first class' concept, with Agents as a nested thing underneath

mbelton-buildkite:
Closing for now. Will open a new PR when doing this change for real.

PR #2142, opened 2023-05-25
Add more state to concurrency-group query

pzeballos:
We have a handy graphQL query to check all the concurrency groups from jobs, but it was missing to check for two more states from the job: `WAITING` and `ASSIGNED`

You also need to define `group` as a parameter of `input`


mbelton-buildkite (approved):
Feel free to merge when the conflict is resolved. Sorry for missing this one before the other went in.

mbelton-buildkite:
(Resolving conflict and merging)

PR #2150, opened 2023-05-26
add documentation for cluster tokens rest api

angusfretwell:
Add docs for https://github.com/buildkite/buildkite/pull/12395 which implements REST API support for cluster tokens. This should only be merged once the API implementation is deployed.

I [raised a question](https://3.basecamp.com/3453178/buckets/27588172/messages/6170557636#__recording_6194022492) about how we want the navigation to work. I'll update this PR based on the outcome of that. 

image


pages/apis/rest_api/cluster_tokens.md
```diff
@@ -0,0 +1,198 @@
+# Cluster Tokens API
+
+A [cluster token](https://buildkite.com/docs/agent/clusters#set-up-a-cluster-connect-agents-to-a-cluster) is used to connect agents to a cluster.
```
angusfretwell, line 3: I think this could be improved, wasn't sure how best to explain cluster tokens.

pages/apis/rest_api/cluster_tokens.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+Required scope: `read_clusters`
+
+Success response: `200 OK`
+
+## Create a token
+
+> 📘 Token values
+> The token value is only be included in the response for the request to create it. To ensure the security of tokens, their value isn't included in subsequent responses.
```
angusfretwell, line 87: This is crap, would like some help wording it better. We include the token value ("token" in the JSON response) only when the token is first created. Subsequent GET requests etc. won't include the token for security reasons.

angusfretwell:
@mbelton-buildkite I've restructured this per your feedback on Basecamp so its on the Clusters API page 😄

pages/apis/rest_api/clusters.md
```diff
@@ -227,3 +231,199 @@ Error responses:
...
+}
+```
+
+Required scope: `read_clusters`
+
+Success response: `200 OK`
+
+### Create a token
+
+> 📘 Token values
```
mbelton-buildkite, line 317: I think this heading could convey the intention of the note more clearly. To me, it looks like the note is about security, but specifically when someone can and can't see the value of a token. Perhaps:

```suggestion
> 📘 Token visibility
```

pages/apis/rest_api/clusters.md
```diff
@@ -227,3 +231,199 @@ Error responses:
...
+> The token value is only be included in the response for the request to create it. To ensure the security of tokens, their value isn't included in subsequent responses.
```
mbelton-buildkite, line 318: I think we should lead with the security aspect. How about:

```suggestion
> To ensure the security of tokens, the value is only included in the response for the request to create the token. Subsequent responses do not contain the token value.
```

mbelton-buildkite (commented):
Just a couple of thoughts, but otherwise looks good 😊

angusfretwell:
Great suggestions thanks @mbelton-buildkite - have applied them

mbelton-buildkite (approved):

PR #2153, opened 2023-05-29
Restructure the Clusters page into a new section

mbelton-buildkite:
This change:
- Creates a new section for information about clusters. 
- Splits the content from the [existing page](https://buildkite.com/docs/agent/clusters) into two new pages.
- Updates any internal links to the old page.
- Adds a redirect from the old page to the new overview.
- Renames a couple of section headings from "Set up" to "Create."

The change should not change the meaning or intent of any of the existing content.

We should ensure any links to the old page in the product are updated as part of this going live.

**Preview:** [Clusters overview](https://bk-docs-pr-2153.onrender.com/docs/clusters/overview)


l-suzuki (approved):
Nice! This makes way more sense than nesting clusters within the buildkite-agent docs ✨

mbelton-buildkite:
I confirmed locally that the redirect catches links to H2s, so will go ahead with merging. The links in the app will be updated in: https://github.com/buildkite/buildkite/pull/12513/files

PR #2156, opened 2023-05-29
document adding pipelines to a cluster

angusfretwell:
* Document the `cluster_id` property on the Pipelines API page
* ~Document how to add pipelines to a cluster on the Clusters API page, and link to Pipelines API~

Closes EDGE-984.

### DIFF
--- a/pages/apis/rest_api/clusters.md
+++ b/pages/apis/rest_api/clusters.md
@@ -227,17 +227,3 @@ Error responses:
   <tr><th><code>422 Unprocessable Entity</code></th><td><code>{ "message": "Reason the cluster couldn't be deleted" }</code></td></tr>
 </tbody>
 </table>
-
-## Add pipelines to a cluster
-
-Use the [Pipelines API](/docs/apis/rest-api/pipelines) to control which cluster a pipeline runs on by setting the `cluster_id` property.
-
-Set `cluster_id` to `null` to remove a pipeline from a cluster.
-
-For example, to add an existing pipeline to a cluster:
-
-```bash
-curl -X PATCH "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{slug}" \
-  -H "Content-Type: application/json" \
-  -d '{ "cluster_id": "42f1a7da-812d-4430-93d8-1cc7c33a6bcf" }'
-```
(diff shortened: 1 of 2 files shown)


pages/apis/rest_api/clusters.md
```diff
@@ -227,3 +227,17 @@ Error responses:
   <tr><th><code>422 Unprocessable Entity</code></th><td><code>{ "message": "Reason the cluster couldn't be deleted" }</code></td></tr>
 </tbody>
 </table>
+
+## Add pipelines to a cluster
```
mbelton-buildkite, line 231: Hmmm I don't think this is in the right spot. This page is reference information where this is more how-to/cookbook content. There's a GraphQL cookbook, but nothing for REST at the moment. What would you think about including it under [Add pipelines to a cluster](https://buildkite.com/docs/agent/clusters#set-up-a-cluster-add-pipelines-to-a-cluster)? Thinking something like:

````
### Add pipelines to a cluster

Add a pipeline to a cluster to ensure the pipeline’s builds run only on agents connected to that cluster. You can add pipelines to a cluster through the Buildkite dashboard or the REST API.

To add a pipeline to a cluster from the Buildkite Dashboard:

1. Navigate to the _Pipeline Settings_ for the pipeline.
1. Under _Cluster Settings_, select the relevant cluster.

To add a pipeline to a cluster from the REST API, run the following:

```bash
curl -X PATCH "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{slug}" \
  -H "Content-Type: application/json" \
  -d '{ "cluster_id": "CLUSTER_ID" }'
```

Replacing:

- `{org.slug}` with your organization slug.
- `{pipeline.slug}` with the slug of the pipeline you want to update.
- `CLUSTER_ID` with the ID of the cluster you want to add the pipeline to.

You can set the cluster ID to `null` to remove a pipeline from all clusters.
````

angusfretwell: It seems a bit advanced to add to the main docs. Maybe it's OK to just drop this section?

mbelton-buildkite: Hmmmm I don't think it's too advanced for that section. I expect curl commands to be familiar to any developer reading the docs. I'm happy for you to make the decision for whether to include it or not, though 😊

angusfretwell: I guess I just don't think that's where you'd expect to find it. Having cookbooks for REST in the future feels like a better place for it to live. I'm happy to just have the attribute documented in the pipeline API docs for now.

mbelton-buildkite (approved):

PR #2161, opened 2023-05-30
IAM policies on Secret Manager on single Elastic CI Stack note 

james2791:
This change:

- Adds a note underneath the main page within the `docs/pages/agent/v3/aws/secrets_manager.md` page (Storing agent tokens) to clarify the need for only one IAM role to be set in a corresponding secret's policy. 
- Came out from community forum [discussion](https://forum.buildkite.community/t/buildkite-elastic-stack-minsuccessfulinstancespercent/2931/6) and testing (Autoscaling Lambda/components not made in the single case)

### DIFF
--- a/pages/agent/v3/aws/secrets_manager.md
+++ b/pages/agent/v3/aws/secrets_manager.md
@@ -46,6 +46,9 @@ To ensure your Elastic CI Stack for AWS has access to the secret:
 }
 ```
 
+>📘 Single instance Elastic CI Stacks
+> If you have set `minSize` and `maxSize` parameters equal to 1 in your [Elastic CI Stack parameters](/docs/agent/v3/elastic-ci-aws/parameters), you will only need to set the `IAMRole` to your Secret Manager secret's resource policy as above. The  `AutoscalingLambdaExecutionRole` IAM role and corresponding autoscaling resources are not created in this Elastic CI Stack setup.
+
 ## Multi-region replication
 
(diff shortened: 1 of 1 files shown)


pages/agent/v3/aws/secrets_manager.md
```diff
@@ -46,6 +46,9 @@ To ensure your Elastic CI Stack for AWS has access to the secret:
 }
 ```
 
+>📘 Single instance Elastic CI Stacks
+> If you have set `minSize` and `maxSize` parameters equal to 1 in your [Elastic CI Stack parameters](/docs/agent/v3/elastic-ci-aws/parameters), you will only need to set the `IAMRole` to your Secret Manager secret's resource policy as above. The  `AutoscalingLambdaExecutionRole` IAM role and corresponding autoscaling resources are not created in this Elastic CI Stack setup.
```
mbelton-buildkite, line 50: I *think* this is a little simpler/clearer:

```suggestion
>📘 Single instance stacks
> If you have set `minSize` and `maxSize` [parameters](/docs/agent/v3/elastic-ci-aws/parameters) equal to `1`, don't include the `AutoscalingLambdaExecutionRole` IAM role in the resource policy. Only the `IAMRole` is required since this setup doesn't need autoscaling resources.
```

But it's a little less precise, so happy to go with your version if you think it works better. Let me know what you think.

PR #2165, opened 2023-05-31
Generate preview URLs using Netlify

dannymidnight:
Render is costing a lot for what it does. 

This PR adds an alternative preview URL solution by instead generating a static site and using Netlify for managing previews.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -43,7 +43,7 @@ module ApplicationHelper
     render(partial: "quick_reference/#{attribute}", formats: [:md]).to_json.html_safe
   end
 
-  def logo_image_url
+  def logo_image_path
     image = 'logo.svg'
 
     # Pride month is June in The United States,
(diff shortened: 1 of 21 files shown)


app/helpers/application_helper.rb
```diff
@@ -43,7 +43,7 @@ def render_attribute_content(attribute)
     render(partial: "quick_reference/#{attribute}", formats: [:md]).to_json.html_safe
   end
 
-  def logo_image_url
+  def logo_image_path
```
mbelton-buildkite, line 46: Just out of curiosity: Why did these need to be updated?

mbelton-buildkite:
Looks sensible to me! Can we see it work or does this need to be merged first?

dannymidnight:
Preview URL: https://false--bk-docs-preview.netlify.app/

dannymidnight:
Preview URL: https://false--bk-docs-preview.netlify.app/

app/helpers/application_helper.rb
dannymidnight: The `_url` suffix in rails outputs absolute URLs and unfortunately when scraping a local version of the site it meant the host was being output as localhost:3000. By using relative URLs I was able to output compatible markup.

dannymidnight:
I've made a few tweaks to the main pipeline so going to re-open another PR to see if it all works as expected...

PR #2167, opened 2023-05-31
Update nav proposal

ozdenyilmaz:
This is a proposed nav update for @mbelton-buildkite to have a look at.

It adds an evaluating buildkite page which adds like a directory of information. 

It also adds getting started guides according to your organization size, infrastructure preferences and code architecture. 

Note this is only a thought exercise, I am not planning to merge it until those pages are incrementally added.

### DIFF
--- a/data/nav.yml
+++ b/data/nav.yml
@@ -4,8 +4,32 @@
   children:
     - name: "Overview"
       path: "pipelines"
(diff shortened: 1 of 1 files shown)


mbelton-buildkite:
We'll still have this history for inspo, but I'll close the PR

PR #2172, opened 2023-05-31
Trigger step `soft_fail` doesn't accept `exit_status`

pda:
This pull request removes mention of `exit_status` from , since trigger steps don't run a command, so they don't have an exit status.

The pipeline parser fails with this error if `exit_status` is provided:

> The `soft_fail` ruleset on a `trigger` step must either be `true` or `false`

Because there's no remaining optional Soft fail attributes, I've deleted that entire section.

### DIFF
--- a/pages/pipelines/trigger_step.md
+++ b/pages/pipelines/trigger_step.md
@@ -110,11 +110,8 @@ _Optional attributes:_
   <tr>
     <td><code>soft_fail</code></td>
     <td>
-      Allow specified non-zero exit statuses not to fail the build.
-      Can be either an array of allowed soft failure exit statuses or <code>true</code> to make all exit statuses soft-fail.<br>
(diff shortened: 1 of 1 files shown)


mcncl (approved):
LGTM

123sarahj123:
LGTM :)

pda:
image

The “Example: `true`” looks a bit silly, and I see some other boolean options describe their default rather than giving an example. I think I'll fix that.

pda:
Aside: we've gotta fix that word wrapping of the attribute name etc. Here's what the table looks like with `white-space: nowrap;` on the `<code>` element:
image

I'm not sure the best way to roll that out without risking other breakages / accessibility issues / etc.

pda:
> The “Example: true” looks a bit silly, and I see some other boolean options describe their default rather than giving an example. I think I'll fix that.

This is a bit better:

image

PR #2186, opened 2023-06-06
Add missing "#" to notify -> slack -> channels example for step-level Slack notifications

sachit-shroff:
If you omit this, you get the following error message when editing pipeline steps:

```
The `slack` notification is invalid: Each channel should be defined as `#channel-name`, `team-name#channel-name`, 'team-name@user-name', '@user-name'` message 
```

### DIFF
--- a/pages/pipelines/notifications.md
+++ b/pages/pipelines/notifications.md
@@ -294,7 +294,7 @@ steps:
     notify:
       - slack:
           channels:
-            - "general"
+            - "#general"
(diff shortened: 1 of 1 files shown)


mbelton-buildkite (approved):
Thanks for picking this up and fixing it, @sachit-shroff! I'll run the linting and then merge for you 😊

sachit-shroff:
Thank you @mbelton-buildkite!

PR #2191, opened 2023-06-07
Add placeholder preview URL comment

dannymidnight:
Post placeholder site and Github comment to let users know something is happening. 

#### Highlights:

1. Deploy a lightweight placeholder site to get a preview URL (it might even be worth making this nice one day)
2. Avoid dependencies and interact with REST APIs so that we can post a Github ASAP, rather than waiting for docker images to build which takes aaaaaaages.

Seems to work alright...
Screenshot 2023-06-08 at 1 12 41 pm

Placeholder site:
Screenshot 2023-06-08 at 1 15 11 pm

It's worth noting that I'm blowing away the preview website on every deployment. I think that's probably fine but if not then I'm sure could also do some extra checks using some Netlify APIs.

### DIFF
--- a/.buildkite/pipeline.preview.yml
+++ b/.buildkite/pipeline.preview.yml
@@ -1,4 +1,19 @@
 steps:
+  - label: "Prepare preview"
+    command: bin/prepare-preview
+    plugins:
+      - docker-compose#v4.12.0:
+          run: agent
+          config: docker-compose.yml
+          mount-checkout: true
+          mount-buildkite-agent: true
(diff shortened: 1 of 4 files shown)


mbelton-buildkite (approved):
Is this the flow for someone updating the docs?

1. Open PR.
2. See a comment with a preview URL link.
3. If they open the link immediately, they see the placeholder site.
4. Eventually, the link becomes the deployed site.

If so, sounds good to me 😊

PR #2197, opened 2023-06-08
add docs for creating, update and deleting cluster queues with REST API

angusfretwell:
Add docs for https://github.com/buildkite/buildkite/pull/12471

### DIFF
--- a/pages/apis/rest_api/clusters.md
+++ b/pages/apis/rest_api/clusters.md
@@ -285,7 +285,7 @@ curl "https://api.buildkite.com/v2/organizations/{org.slug}/clusters/{cluster.id
   {
     "id": "01885682-55a7-44f5-84f3-0402fb452e66",
     "graphql_id": "Q2x1c3Rlci0tLTQyZjFhN2RhLTgxMmQtNDQzMC05M2Q4LTFjYzdjMzNhNmJjZg==",
-    "key": "Default Queue",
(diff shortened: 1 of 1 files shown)


jordandcarter (commented):
Looks good! I'll let @mbelton-buildkite give it the tick though

pages/apis/rest_api/clusters.md
```diff
@@ -346,6 +346,138 @@ Required scope: `read_clusters`
 
 Success response: `200 OK`
 
+### Create a queue
+
+```bash
+curl -X POST "https://api.buildkite.com/v2/organizations/{org.slug}/clusters/{cluster.id}/queues" \
+  -H "Content-Type: application/json" \
+  -d '{ "key": "default", "description": "The default queue for this cluster" }'
```
mbelton-buildkite, line 354: I thought a cluster already had a default queue when created? If so, perhaps this should create a queue with a different key and description to be more realistic?

mbelton-buildkite (approved):
Looks good to me! I've added one minor comment to consider, but happy for you to decide how to proceed with it. 😊

pages/apis/rest_api/clusters.md
angusfretwell: They do when created through the UI however we decided to change this behaviour for the API so it's more predictable and the user is in full control!

PR #2210, opened 2023-06-15
Add 'coordinating multiple agents'

DrJosh9000:
A new section of the parallel builds guide.

Note to self: this will need updating for the next agent release.

### DIFF
--- a/pages/tutorials/parallel_builds.md
+++ b/pages/tutorials/parallel_builds.md
@@ -35,6 +35,30 @@ buildkite-agent start --tags queue=test
 buildkite-agent start --tags queue=deploy
 ```
 
+#### Coordinating multiple agents
+
+>🛠️ Experimental feature
+> To use it, set <code>experiment="agent-api"</code> in your <a href="/docs/agent/v3/configuration#experiment"> agent configuration</a>.
+> This also requires Agent v3.47.0 or later.
+
+Multiple agents on a single host can sometimes interfere with one another. For example, a pipeline might contain commands like `docker prune` or `apt upgrade`, but these commands fail if another job is running the same command at the same time.
+
+To coordinate access to shared resources on the same host, you can use agent locks. Locking is advisory (nothing prevents a buggy command from ignoring a lock), but it can help avoid multiple agents interfering with each other.
+
+Here's how you could use locks to make sure a command is run by only one agent at a time:
+
+```bash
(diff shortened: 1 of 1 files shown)


pages/tutorials/parallel_builds.md
```diff
@@ -35,6 +35,30 @@ buildkite-agent start --tags queue=test
 buildkite-agent start --tags queue=deploy
 ```
 
+#### Coordinating multiple agents
+
+>🛠️ Experimental feature
+> To use it, set <code>experiment="agent-api"</code> in your <a href="/docs/agent/v3/configuration#experiment"> agent configuration</a>.
+> This also requires Agent v3.47.0 or later.
```
mbelton-buildkite, line 42: I don't feel too strongly about this, but I think you can ditch the "also."

```suggestion
> This requires Agent v3.47.0 or later.
```

pages/tutorials/parallel_builds.md
```diff
@@ -35,6 +35,30 @@ buildkite-agent start --tags queue=test
...
+
+Multiple agents on a single host can sometimes interfere with one another. For example, a pipeline might contain commands like `docker prune` or `apt upgrade`, but these commands fail if another job is running the same command at the same time.
```
mbelton-buildkite, line 44: General tightening:

```suggestion
Multiple agents on a single host can sometimes interfere with one another. For example, a pipeline might contain commands like `docker prune` or `apt upgrade`, but these commands fail if another job runs the same command at the same time.
```

You could also consider changing "at the same time" to "simultaneously," but I'm not sure it makes it any clearer.

pages/tutorials/parallel_builds.md
```diff
@@ -35,6 +35,30 @@ buildkite-agent start --tags queue=test
...
+
+To coordinate access to shared resources on the same host, you can use agent locks. Locking is advisory (nothing prevents a buggy command from ignoring a lock), but it can help avoid multiple agents interfering with each other.
+
+Here's how you could use locks to make sure a command is run by only one agent at a time:
```
mbelton-buildkite, line 48: Where would these commands be defined? I assume in a script? Could we mention something about that in this lead in? Thinking something like:

```suggestion
Here's how you could use locks in a script to make sure a command is run by only one agent at a time:
```

mbelton-buildkite (approved):

moskyb (approved):

PR #2212, opened 2023-06-16
s/Golang/Go/ because the latter is its name

pda:
Replace “Golang” with “Go” in prose and headings, so that we're using the correct name of the language.

I've also linked “Go” to its official https://go.dev/ site in one place in case that helps clarity now that we're not saying “the Go language” or “Golang”, and done the same for a couple of other languages (Rust, Ruby) but not all, so maybe that's silly, but I figure it's somewhat helpful to disambiguate e.g. Go from https://en.wikipedia.org/wiki/Go_(game) and Rust from https://en.wikipedia.org/wiki/Rust_(video_game) and Ruby from https://en.wikipedia.org/wiki/Ruby_character 🤷‍♂️

Anyway… back to Go:

> **Is the language called Go or Golang?**
>
> The language is called Go. The "golang" moniker arose because the web site was originally golang.org. (There was no .dev domain then.) Many use the golang name, though, and it is handy as a label. For instance, the Twitter tag for the language is "#golang". The language's name is just plain Go, regardless.

-- https://go.dev/doc/faq#go_or_golang

### DIFF
--- a/data/nav.yml
+++ b/data/nav.yml
@@ -343,7 +343,7 @@
           pill: "beta"
         - name: "Python"
           path: "test-analytics/python-collectors"
-        - name: "Golang"
+        - name: "Go"
           path: "test-analytics/golang-collectors"
         - name: ".NET"
           path: "test-analytics/dotnet-collectors"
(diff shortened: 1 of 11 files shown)


triarius (approved):
🙏

PR #2221, opened 2023-06-19
Update defining_steps.md

stephanieatte:
Users keep getting confused about Build States and Job states. It makes sense to give description of jobs state in our docs


pages/pipelines/defining_steps.md
```diff
@@ -124,6 +123,31 @@ When you run a pipeline, a build is created. Each of the steps in the pipeline e
 
 <%= image "job-states.svg", size: "1482x690", alt: "Job state diagram" %>
 
+Job State             | Description
```
mbelton-buildkite, line 126: Style: Use sentence casing.

```suggestion
Job state             | Description
```

pages/pipelines/defining_steps.md
```diff
@@ -124,6 +123,31 @@ When you run a pipeline, a build is created. Each of the steps in the pipeline e
...
+----------------------| -----------------------------------------
+PENDING               | The job has just been created and doesn't have a state yet
```
mbelton-buildkite, line 128: Style:

- In the diagram above and text below it looks like these are normally written in lower case and using code styling. I think this should be consistent with that unless there's a good reason not to.
- End each description with a period.

```suggestion
`pending`               | The job has just been created and doesn't have a state yet.
```

^ This applies to each row.

mbelton-buildkite (commented):
This is a good addition! Just a few style nitpicks to fix.

gilesgas (approved):
Thanks very much for your contribution @stephanieatte !

This is now ready to merge 🎉

PR #2222, opened 2023-06-19
Add "Cluster" to the glossary

mbelton-buildkite:
With clusters becoming generally available in the upcoming release and being an important concept to understand, I think it should have an entry in the glossary. I've taken a first pass from the info on the existing page, but very open to changing it.


pages/pipelines/glossary.md
```diff
@@ -18,6 +18,10 @@ To learn more, see [Build artifacts](/docs/pipelines/artifacts).
 
 A build is a single run of a pipeline. You can trigger a build in various ways, including through the dashboard, API, as the result of a webhook, on a schedule, or even from another pipeline using a trigger step.
 
+## Cluster
+
+A cluster is a grouping of agents and pipelines. They allow teams to self-manage their agent pools, let admins create isolated sets of agents and pipelines within the one Buildkite organization, and help to make agents and queues more discoverable across your organization.
```
timriley, line 23:
```suggestion
A cluster groups queues of agents along with pipelines. They allow teams to self-manage their agent pools, let admins create isolated sets of agents and pipelines within the one Buildkite organization, and help to make agents and queues more discoverable across your organization.
```

timriley (approved):
This is a good addition!

If I were to suggest one thing, it would be to emphasise that clusters contain _queues_ of agents, since that's the main concept that people end up interacting with when they use clusters. I know you include queues in the latter part of the description, but that feels a bit like it's "burying the lede" IMHO :)

You might have a better idea about how to incorporate this into a short description, but I've provided one suggestion in case it helps :)

Otherwise, I'd still be very happy to see this added as-is. We can finesse it later.

gilesgas:
Thanks @timriley for reminding me of this PR! 🤦 

I was meant to merge it a while back. I'll do so once its build completes successfully

pages/pipelines/glossary.md
gilesgas: Thanks!

gilesgas (approved):

PR #2223, opened 2023-06-19
Update cli_step.md

stephanieatte:
Customers have recently wanted to be be able update each job label individually and it is not possible. Because the Buildkite-agent does the label update on a step level

### DIFF
--- a/pages/agent/v3/cli_step.md
+++ b/pages/agent/v3/cli_step.md
@@ -9,6 +9,7 @@ Use this command in your build scripts to update an attribute of a step.
 
 <%= render 'agent/v3/help/step_update' %>
 
+label update works on a step level and not on the job level, it would not be possible to update each job in a step 
 ## Getting a step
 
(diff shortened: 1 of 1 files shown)


pages/agent/v3/cli_step.md
```diff
@@ -9,6 +9,7 @@ Use this command in your build scripts to update an attribute of a step.
 
 <%= render 'agent/v3/help/step_update' %>
 
+label update works on a step level and not on the job level, it would not be possible to update each job in a step
```
mbelton-buildkite, line 12: Could you explain a bit more about when someone might find themselves in this situation? For example, a correct version of something like:

> Labels are applied to steps, not jobs, so you can't update a job label. When a step contains multiple jobs, all the jobs use the same label.

I also wonder if this might fit in better above the example where it's already talking about step labels.

mbelton-buildkite: If there's a ticket in Helpscout or a post in the forum, it might help me understand the details a bit better 😊

PR #2227, opened 2023-06-20
Update pipeline.yml to pull from AWS ECR

gchan:
To avoid DockerHub's rate limiting.

We want to use the public ECR so we don't need to authenticate during
the build.

Guide
https://guide.buildkite.net/engineering/getting-started/ci/docker/index.html#use-a-pull-through-cache


swebb (approved):
Assuming the tests pass LGTM 👍🏻

PR #2228, opened 2023-06-20
Bugfix curl command for Suite#create (localhost => com)

amybiyuliu:
I just spotted a bug on our docs:

Screenshot 2023-06-20 at 2 40 57 pm

I think it should be buildkite.com rather than buildkite.localhost


gchan (approved):
🐛 🛳️

mbelton-buildkite (approved):
Thanks for finding and fixing!

PR #2231, opened 2023-06-20
Update sso.md

stephanieatte:
We get Users who cannot access their account because they have sso disabled and in the user settings sso is required. This update let's the know what the issue probably i


mbelton-buildkite:
What do you think of reframing this as a question in the FAQ section of this page? I think it might help by framing it around the behavior someone sees when they run into this issue.

PR #2233, opened 2023-06-20
[PIE-1810] Fix typo in suites#create documentation (http->https)

gchan:
PIE-1810

The documentation for creating a suite via API [1] uses HTTP rather than
HTTPS.

Attempting to make a request via HTTP will result in a 301 Move
Permanently response which will ultimately fail as the HTTP client may
make a GET request when following the redirect.

[1]
https://buildkite.com/docs/apis/rest-api/analytics/suites#create-a-suite


mbelton-buildkite (approved):

amybiyuliu (approved):
LGTM

PR #2236, opened 2023-06-21
Pipeline's UUID and info added to graphql cookbook

butterkitty:
Adds to the GraphQL cookbook the ability to retrieve a pipeline's UUID and use that UUID inorder to get in depth info about that pipeline.

Reference HelpScout Ticket: https://secure.helpscout.net/conversation/2280173033/53169/

### DIFF
--- a/pages/apis/graphql/graphql_cookbook.md
+++ b/pages/apis/graphql/graphql_cookbook.md
@@ -82,6 +82,48 @@ query RecentPipelineSlugs {
 }
 ```
 
+### Get a pipeline's UUID
+
+Get a pipeline's UUID by searching for it in the API. Search term can match a pipeline slug.
+
+_Note: Pipeline slugs are modifiable and can change_
+
+```graphql
+query GetPipelineUUID {
+  organization(slug: "organization-slug") {
+    pipelines(first: 50, search: "part of slug") {
+      edges {
+        node {
+          slug
+          uuid
+        }
+      }
+    }
+  }
+}
+```
+
+### Get a pipeline's information
+
+You can get specific pipeline information for each of your pipeline. You can retrieve information for each build, jobs, and any other information listed on [this](https://buildkite.com/docs/apis/graphql/schemas/object/pipeline) page.
+
+```graphql
+query GetPipelineInfo {
+  pipeline(uuid: "pipeline-uuid") {
+    slug
+    uuid
+    builds(first:50){
+      edges {
+        node {
+          state
+          message
+        }
+      }
+    }
+  }
+}
+```
+
 ### Get pipeline metrics
 
 The _Pipelines_ page in Buildkite shows speed, reliability, and builds per week, for each pipeline. You can also access this information through the API.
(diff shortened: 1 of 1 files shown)


pages/apis/graphql/graphql_cookbook.md
```diff
@@ -82,6 +82,48 @@ query RecentPipelineSlugs {
 }
 ```
 
+### Get a pipeline's UUID
```
mbelton-buildkite, line 85: I think it would be good to frame this heading more around the task rather than the answer. It looks like the question came from someone wanting a way to refer to pipelines that won't change over time. Perhaps:

```suggestion
### Get a pipeline's persistent identifier
```

What do you think? I feel like there might be a better term. I like "permanent" more than persistent but it looks like [persistent identifiers](https://en.wikipedia.org/wiki/Persistent_identifier) are a thing. I also thought of "fixed" but don't want to cause any confusion with fixing a failing pipeline.

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -82,6 +82,48 @@ query RecentPipelineSlugs {
...
+
+Get a pipeline's UUID by searching for it in the API. Search term can match a pipeline slug.
+
+_Note: Pipeline slugs are modifiable and can change_
```
mbelton-buildkite, line 89: I think we should pull out some more context from the forum post. Maybe:

```suggestion
While a pipeline's name and slug can change over time, its UUID is permanent. Use the UUID when you need a way to reference a pipeline that won't change.
```

pages/apis/graphql/graphql_cookbook.md
butterkitty: Hmm, I do really like the idea. I think the problem is that it is a UUID (universal unique identifier) which denotes it's uniqueness as an identifier.

What if instead, we used the wording "Get a pipeline's persistent UUID"

mbelton-buildkite: 👍🏻 I'm good with that

butterkitty: Updated with new commit! :D

mbelton-buildkite:
I'll merge `main` for you to fix the linting error.

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -105,7 +103,7 @@ query GetPipelineUUID {
 
 ### Get a pipeline's information
 
-You can get specific pipeline information for each of your pipeline. You can retrieve information for each build, jobs, and any other information listed on [this](https://buildkite.com/docs/apis/graphql/schemas/object/pipeline) page.
+You can get more specific pipeline information for each of your pipelines. You can retrieve information for each build, jobs, and any other information listed on [this](https://buildkite.com/docs/apis/graphql/schemas/object/pipeline) page.
```
mbelton-buildkite, line 106:
```suggestion
You can get specific information about your pipelines, including details of each build, job, and anything from the [Pipeline object](/docs/apis/graphql/schemas/object/pipeline).
```

mbelton-buildkite (approved):
One last suggestion, but otherwise looks good :)

PR #2196, opened 2023-06-08
Update docs for Agent v3.48.0

DrJosh9000:

PR #2240, opened 2023-06-23
Pattern for promoted experiments

mbelton-buildkite:
This change suggest a pattern for handling agent experiments that have been promoted:

- Note version it became a full feature in the configuration reference.
- List them in a new section on the agent overview, newest first.

I'd like to come back to this in the future, but think it solves the immediate need from #2196. What do you think?

### DIFF
--- a/data/content/agent_attributes.yaml
+++ b/data/content/agent_attributes.yaml
@@ -91,7 +91,7 @@ attributes:
   experimental: false
   required: false
   desc: |
-      Flags to pass to the `git clone` command when used for mirroring. Refer to [Git mirrors](/docs/agent/v3#promoted-experiments-git-mirrors) for more information on this feature.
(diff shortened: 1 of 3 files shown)


data/content/agent_attributes.yaml
mbelton-buildkite, line 1: Matches pattern from when [redacted variables from promoted](https://github.com/buildkite/docs/pull/1073/files).

pages/agent/v3.md
```diff
@@ -153,38 +147,39 @@ The API is exposed using a Unix Domain Socket. Unlike the `job-api`, the path to
 
 ## Promoted experiments
 
-We've recently promoted some experiments into full features.
+The following features started as experiments before being promoted to fully supported features.
```
mbelton-buildkite, line 150: Avoids being tied to a particular time, then lists features from newest to oldest.

DrJosh9000 (approved):
Looks good, makes sense, reads well!

PR #2241, opened 2023-06-23
update allow_dependency_failure section with behavior for cancelled jobs

nsuma8989:
This PR adds information regarding behavior we can expect when a job is cancelled with respective to allow_dependency_failure set to true.

For example if a pipeline has two steps A and B where B has depends_on: A but has allow_dependency_failure set to true so that in case of failure of step A, step B can still run. However when step A is cancelled then in that scenario even if allow_dependency_failure is set to true step B will not run.


pages/pipelines/dependencies.md
```diff
@@ -130,7 +130,7 @@ Steps that are in a `concurrency_group` run in the order they are created in and
 
 ## Allowing dependency failures
 
-You can add the `allow_dependency_failure` attribute to any step that has dependencies. The step will then run when the depended-on jobs complete, fail or even did not run. However, even if you continue to run the next step, if there are any failures at all, the build will still end up failing.
+You can add the `allow_dependency_failure` attribute to any step that has dependencies. The step will then run when the depended-on jobs complete, fail or even did not run but not after a cancelation. If you cancel a job, any subsequent steps with `allow_dependency_failure: true` do not execute. However, even if you continue to run the next step, if there are any failures at all, the build will still end up failing.
```
mbelton-buildkite, line 133: I think this is a bit clearer:

```suggestion
You can add the `allow_dependency_failure` attribute to any step that has dependencies. The step will then run when the depended-on jobs complete, fail, or do not run. However, if you cancel a job, any subsequent steps with `allow_dependency_failure: true` do not execute. Note that even if you continue to run the next step, the build will still fail if there are any failures.
```

What do you think?

nsuma8989:
@mbelton-buildkite I updated PR with suggested changes. Please review

mbelton-buildkite (approved):

mbelton-buildkite:
I'll merge `main` for you, which should fix the linting error. 😊

PR #2242, opened 2023-06-23
Add note about substitution of user with build creator env variable v…

nsuma8989:
This PR adds a note that we cannot substitute user with value of build creator env variable as that variable interpolation is not possible

### DIFF
--- a/pages/pipelines/notifications.md
+++ b/pages/pipelines/notifications.md
@@ -299,6 +299,9 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
+>🚧
+> We cannot substitute `user` with the build creator environment variable value
+
 ### Conditional Slack notifications
 
 You can also add [conditionals](/docs/pipelines/notifications#conditional-notifications) to restrict the events on which notifications are sent:


pages/pipelines/notifications.md
```diff
@@ -299,6 +299,9 @@ steps:
 ```
 
+>🚧
```
mbelton-buildkite, line 302: Let's add a heading to this callout. Maybe:

```suggestion
>🚧  Build creator environment variable
```

pages/pipelines/notifications.md
```diff
@@ -299,6 +299,9 @@ steps:
...
+> We cannot substitute `user` with the build creator environment variable value
```
mbelton-buildkite, line 303: Make this more about them than us + add a period:

```suggestion
> You cannot substitute `user` with the build creator environment variable value.
```

mbelton-buildkite: Is there anything we can say about why this is the case? You mentioned environment variable interpolation in the PR description.

nsuma8989:
@mbelton-buildkite I updated the PR with suggested changes. Please review

mbelton-buildkite (approved):

mbelton-buildkite:
I'm going to merge `main` for you, which should fix the linting error. 😊

PR #2255, opened 2023-06-27
Add documentation for pipeline templates (EDGE-963)

mhaylock:
Introduces documentation for the new Pipeline Templates feature.

Per guidance from @ozdenyilmaz, this is under a new "Governance" section.

This is my first crack at a documentation update, I look forward to learning how to make this better! 🙌 

Missing:
~~- [ ] Updated GraphQL docs (should have this ready tomorrow)~~
~~- [ ] An example in the REST API docs (not sure if this is needed?)~~

PR #2259, opened 2023-06-28
Proposal: Move Build exports page into Governance section

mbelton-buildkite:
Pipeline templates is a new feature in the upcoming release. We're including documentation for it in a new section of the docs called "Governance" in #2255. I thought it would be good to include any existing pages that would fit better in that section and @ozdenyilmaz mentioned that Build exports would be a good candidate.

What do you think?


123sarahj123 (approved):
LGTM :)

PR #2265, opened 2023-06-29
Adds a section on input validation

karensawrey:
This PR adds a section on input validation - the very existence of this option and guidelines about the syntax to use based on the following Buildkite Forum post: https://forum.buildkite.community/t/regex-validation-for-step-input/1260/3

### DIFF
--- a/pages/pipelines/input_step.md
+++ b/pages/pipelines/input_step.md
@@ -293,4 +293,4 @@ steps:
         format: "[0-9a-f]+"
 ```
 
-The format must be a regular expression implicitly anchored to the beginning and end of the input and is functionally equivalent to the [HTML5 pattern attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).
+The `format` must be a regular expression implicitly anchored to the beginning and end of the input and is functionally equivalent to the [HTML5 pattern attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).
--- a/pages/pipelines/input_step.md
+++ b/pages/pipelines/input_step.md
@@ -284,13 +284,13 @@ If you associate a RegEx to a field, the field outline will turn red when an inv
 
 To do it, use the following sample syntax:
 
-```
+```yml
 steps:
-- block: "Click me!"
-  fields:
-  - text: Must be hexadecimal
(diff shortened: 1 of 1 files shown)


pages/pipelines/input_step.md
```diff
@@ -275,3 +275,22 @@ steps:
             value: "stable"
 ```
+
+## Input validation
+
+To prevent the users from entering invalid values when using an input step with text inputs (for example, to gather some deployment information), you can use input validation.
```
mbelton-buildkite, line 281: I think this could be a little tighter, but happy for you to leave it as-is if you think it's clearer:

```suggestion
To prevent users from entering invalid text values in input steps  (for example, to gather some deployment information), you can use input validation.
```

pages/pipelines/input_step.md
```diff
@@ -275,3 +275,22 @@ steps:
...
+
+If you associate a RegEx to a field, the field outline will turn red when an invalid value is entered.
```
mbelton-buildkite, line 283: Let's not abbreviate this:

```suggestion
If you associate a regular expression to a field, the field outline will turn red when an invalid value is entered.
```

pages/pipelines/input_step.md
```diff
@@ -275,3 +275,22 @@ steps:
...
+```yml
+steps:
+  - input: "Click me!"
+    fields:
+      - text: Must be hexadecimal
+        key: hex
+        format: "[0-9a-f]+"
+```
+
+The format must be a regular expression implicitly anchored to the beginning and end of the input and is functionally equivalent to the [HTML5 pattern attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).
```
mbelton-buildkite, line 296: Use code formatting when referring to the field:

```suggestion
The `format` must be a regular expression implicitly anchored to the beginning and end of the input and is functionally equivalent to the [HTML5 pattern attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).
```

mbelton-buildkite (approved):
One suggestion and a couple of small things to update, otherwise looks good to me 😊

karensawrey:
Great, thanks for the review, @mbelton-buildkite!

PR #2268, opened 2023-06-30
Update monitors.md

stephanieatte:
We keep getting questions about when Test Analytics Monitoring and Alerts will be back. The docs is misleading it says it coming back soon. But Test Analytics confirmed they do not have a concrete timeline

Reference Ticket
https://secure.helpscout.net/conversation/2288060702/53565/

### DIFF
--- a/pages/test_analytics/monitors.md
+++ b/pages/test_analytics/monitors.md
@@ -6,7 +6,7 @@ Monitors track what matters to you. They trigger alerts under conditions that yo
 ## Monitors
 
 >🚧 Feature temporarily unavailable
-> Monitors and alerts for Test Analytics are unavailable as of September 05, 2022. We expect them to be back soon in the capacity outlined on this page.
+> Monitors and alerts for Test Analytics are unavailable as of right. We are working on this, but do not have a concrete timeline for when it will be released
 
 There are three different types of monitors in Test Analytics:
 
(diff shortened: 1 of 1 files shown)


pages/test_analytics/monitors.md
```diff
@@ -30,4 +30,4 @@ When you get an alert from one of your monitors, see
 
 Mark it as resolved, mute it, or leave it open.
 
-**Coming soon:** web hooks and integrations with platforms such as Slack, GitHub, and issue tracking tools so you can assign ownership of problematic tests, and see alerts in your preferred platform.
+**Added new Features:** web hooks and integrations with platforms such as Slack, GitHub, and issue tracking tools so you can assign ownership of problematic tests, and see alerts in your preferred platform.
```
meiqimichelle, line 33: Recommend leaving this as-is, or removing the entire "coming soon" section if you're trying to avoid questions/confusion. "Added new features" to me sounds as if this is already released, which is not accurate.

meiqimichelle:
Hello! I had another look -- @mbelton-buildkite , Monitors and Alerts are unlikely to return in their current format. Should we remove the entire sub-section of our TA docs called 'Monitoring' for now?

mbelton-buildkite:
> Monitors and Alerts are unlikely to return in their current format. Should we remove the entire sub-section of our TA docs called 'Monitoring' for now?

I think so, yes. Based on that note, they haven't been available for ~10 months. Let's remove it to give you a clean slate when launching their new form when you're ready.

@meiqimichelle, are there any customers or comms we need to manage? Or could we update this PR to remove the page and be done?

PR #2269, opened 2023-06-30
clarify supported bucket encryption

pzeballos:
If customers want to use their own bucket for logs, we only support SSE-S3 encryption

### DIFF
--- a/pages/pipelines/managing_log_output.md
+++ b/pages/pipelines/managing_log_output.md
@@ -157,6 +157,8 @@ The folder structure and file format are as follows and are not customizable:
 To set up a private build log archive storage:
 
 1. Create an Amazon S3 bucket in *us-east-1* location (the only region that is currently supported).
+>📘 SSE-S3 Encryption Support
+> We don't support SSE-KMS encryption for customer log buckets, only SSE-S3. 
 2. Provide *read* and *write* access permission policy for the Buildkite's AWS account `032379705303`.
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/managing_log_output.md
```diff
@@ -157,6 +157,8 @@ The folder structure and file format are as follows and are not customizable:
 To set up a private build log archive storage:
 
 1. Create an Amazon S3 bucket in *us-east-1* location (the only region that is currently supported).
+>📘 SSE-S3 Encryption Support
+> We don't support SSE-KMS encryption for customer log buckets, only SSE-S3.
```
mbelton-buildkite, line 161: Do you think we could incorporate this into the intro paragraph instead? Maybe:

> By default, build logs are stored in encrypted form in Buildkite's managed Amazon S3 buckets, but you can instead store the archived build logs in your private AWS S3 bucket. If you decide to store the logs in your S3 bucket, they're encrypted using SSE-S3. SSE-KMS encryption is not supported. After storing the logs in your S3 bucket, Buildkite does not retain a copy of the logs.

pages/pipelines/managing_log_output.md
```diff
@@ -157,6 +157,7 @@ The folder structure and file format are as follows and are not customizable:
 To set up a private build log archive storage:
 
 1. Create an Amazon S3 bucket in *us-east-1* location (the only region that is currently supported).
+
```
mbelton-buildkite, line 160: Nitpick: Unnecessary blank line.

```suggestion
```

mbelton-buildkite (approved):
Looks good to me! Just one formatting fix you could do before merging.

PR #2229, opened 2023-06-20
rate limits documentation

himalb001:
Public documentation of GraphQL resource limits.

More context:
https://3.basecamp.com/3453178/buckets/1933834/messages/6221093655
https://3.basecamp.com/3453178/buckets/23328660/messages/6161786100#__recording_6176276293

This PR 
* adds a nav item under API > Graphql > Limits
* Adds markdown file with graphql rate limits documentation

PR #2276, opened 2023-07-03
Add Limits to GraphQL nav data

JuanitoFatas:
Limits doc was introduced in https://github.com/buildkite/docs/pull/2229 and it should have updated this `scripts/graphql_api_content/nav_data.rb` instead of the resulting `data/nav_graphql.yml`.


dannymidnight (approved):
👍 Looks like there's just one overly specific test case to fix up.

PR #2283, opened 2023-07-05
Update dev setup guide

kalopilato:
While setting up my development environment I hit a couple snags running `bin/setup`:

1. a couple of gems couldn't build because I didn't have Xcode Command Line Tools installed

Screenshot 2023-07-05 at 9 20 14 AM

Screenshot 2023-07-05 at 9 20 29 AM

2. I wasn't running a compatible node version:

Screenshot 2023-07-05 at 9 19 32 AM

This PR:

- updates the README to include Xcode Command Line Tools in the setup instructions
- adds an `.nvmrc` file so that we can use nvm to use a compatible node version. we could probably make the `bin/setup` script a bit smarter in future but this should do the trick for now


dannymidnight (approved):
Nice one. Thanks for this 👍

PR #2284, opened 2023-07-05
Update docs for 'Creating a pipeline template'

kalopilato:
We recently modified the user flow for creating a Pipeline template so we need docs to reflect this change. We no longer immediately create a template, we now follow a more standard creation flow: click create > fill out a form > click save.


kalopilato:
weird, I've pushed a commit to revert the explicit list numbering but it's not shown up here yet 🤔

kalopilato:
I'm gonna close/recreate the PR and see if that resolves things...

PR #2285, opened 2023-07-05
Update docs for 'Creating a pipeline template'

kalopilato:
We recently modified the user flow for creating a Pipeline template so we need docs to reflect this change. We no longer immediately create a template, we now follow a more standard creation flow: click create > fill out a form > click save.

### DIFF
--- a/pages/pipelines/templates.md
+++ b/pages/pipelines/templates.md
@@ -17,7 +17,9 @@ To create a template:
 
 1. Navigate to your [organization’s pipeline templates](https://buildkite.com/organizations/-/pipeline-templates).
 1. If this is your first template, select _Create a Template_. Otherwise, select _New Template_.
-1. Enter the name and description for your new template, update the default step configuration and select _Create Template_ to create the template.
+1. Enter the name and description for your new template.
+1. Update the default step configuration.
+1. Select _Create Template_.
 
 An administrator can add multiple templates to use across the organization. Making changes and saving a template will apply those changes to all pipelines using that template.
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/templates.md
```diff
@@ -16,9 +16,8 @@ Only administrators can create or update pipeline templates.
 To create a template:
 
 1. Navigate to your [organization’s pipeline templates](https://buildkite.com/organizations/-/pipeline-templates).
-1. If this is your first template, select _Create a Template_. Otherwise, select _New Template_. A pipeline template is created automatically for you.
-1. Update the name and description of your new template and select _Apply_ to save your changes.
-1. Select _Edit_ if you want to change the step configuration for the template.
+1. If this is your first template, select _Create a Template_. Otherwise, select _New Template_.
+1. Enter the name and description for your new template, update the default step configuration and select _Create Template_ to create the template.
```
mbelton-buildkite, line 20: It's a good practice to do one instruction per step, which I think we can here:

```suggestion
1. Enter the name and description for your new template.
1. Update the default step configuration.
1. Select _Create Template_.
```

kalopilato: Nice one 👌 Should've just clicked commit but have applied in [c82fd48](https://github.com/buildkite/docs/pull/2285/commits/c82fd48bbc6a2c7fbc1529ef9c1e3c37fb4acbd0)

mbelton-buildkite (approved):
Looks good to me!

PR #2289, opened 2023-07-06
Clarify how to view query complexity of a query

himalb001:
We had some confusion from customers about the value of the header required to access the query complexity. 

This change makes it explicit that the value of `true` needs to be passed in the header.

### DIFF
--- a/pages/apis/graphql/graphql_resource_limits.md
+++ b/pages/apis/graphql/graphql_resource_limits.md
@@ -154,7 +154,7 @@ RateLimit-Reset: 120
 
 ### View query complexity
 
-To include the complexity data in responses, set `Buildkite-Include-Query-Stats` header to `true` in GraphQL requests. This returns the complexity data in the response like the following:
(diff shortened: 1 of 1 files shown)


pages/apis/graphql/graphql_resource_limits.md
```diff
@@ -154,7 +154,7 @@ RateLimit-Reset: 120
 
 ### View query complexity
 
-To include the complexity data in responses, use the `Buildkite-Include-Query-Stats` header in GraphQL requests. This returns the complexity data in the response like the following:
+To include the complexity data in responses, set `Buildkite-Include-Query-Stats` header to `true` in GraphQL requests. This returns the complexity data in the response like the following:
```
mbelton-buildkite, line 157: Missing "the"

```suggestion
To include the complexity data in responses, set the `Buildkite-Include-Query-Stats` header to `true` in GraphQL requests. This returns the complexity data in the response like the following:
```

mbelton-buildkite (approved):
One nitpick, otherwise all good :)

PR #2292, opened 2023-07-10
Use `spawn` for multiple agents

DrJosh9000:
... on Debian, Red Hat, and Ubuntu docs.

### DIFF
--- a/pages/agent/v3/_linux_multiple_agents.md
+++ b/pages/agent/v3/_linux_multiple_agents.md
@@ -1,0 +1,7 @@
+You can run as many parallel agent workers on the one machine as you wish with
+the `spawn` configuration setting, or by passing the `--spawn` flag.
+
+```ini
+# Start 5 workers. Each one will independently fetch and execute jobs.
+spawn=5
+```
--- a/pages/agent/v3/_linux_multiple_agents.md
+++ b/pages/agent/v3/_linux_multiple_agents.md
@@ -2,6 +2,6 @@ You can run as many parallel agent workers on the one machine as you wish with
 the `spawn` configuration setting, or by passing the `--spawn` flag.
 
 ```ini
-# Start 5 workers. Each one will independently fetch and execute jobs.
+# Start 5 workers. Each one independently fetches and executes jobs.
 spawn=5
 ```
--- a/pages/agent/v3/debian.md
+++ b/pages/agent/v3/debian.md
@@ -88,25 +88,7 @@ On Debian, the Buildkite agent runs as user `buildkite-agent`.
 
 ## Running multiple agents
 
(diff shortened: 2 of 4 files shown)


pages/agent/v3/_linux_multiple_agents.md
```diff
@@ -0,0 +1,7 @@
+You can run as many parallel agent workers on the one machine as you wish with
+the `spawn` configuration setting, or by passing the `--spawn` flag.
+
+```ini
+# Start 5 workers. Each one will independently fetch and execute jobs.
```
mbelton-buildkite, line 5: Style nitpick:

```suggestion
# Start 5 workers. Each one independently fetches and executes jobs.
```

pages/agent/v3/_linux_multiple_agents.md
```diff
@@ -0,0 +1,7 @@
...
+spawn=5
+```
```
mbelton-buildkite, line 7: Ending the file with a blank line should fix the linting error:

````suggestion
```

````

**Review by mbelton-buildkite — APPROVED** [review]
Nice! Just a couple of small things to update

**Comment by DrJosh9000** [comment]
I'm wondering whether to keep the old text, or at least allude to using the systemd template as an alternative, since we still ship it.

**Comment by mbelton-buildkite** [comment]
> I'm wondering whether to keep the old text, or at least allude to using the systemd template as an alternative, since we still ship it.

Riiiight, so the existing text is still technically correct, it's just not our recommended approach? If that's the case, do you have a sense of how important it is to know the other method? Will anyone not be able to use the new approach you're documenting?

If it's needed, we can document it alongside this but make sure to emphasize which is the preferred method. If it's an edge case that shouldn't be needed, I'd rather omit it to keep things simple. I'm happy to talk it through if that's easier.

**Comment by DrJosh9000** [comment]
That helps, thanks @mbelton-buildkite. The `--spawn`-based setup is more generally compatible and anyone using a systemd unit template should be able to use `--spawn`. I see the main downside to deleting the old method might be that anyone who set it up that way might want to refer to some documentation to figure out what they did.

File: pages/agent/v3/_linux_multiple_agents.md [hunk 3]
```diff
@@ -0,0 +1,7 @@
+You can run as many parallel agent workers on the one machine as you wish with
+the `spawn` configuration setting, or by passing the `--spawn` flag.
+
+```ini
```
triarius, line 4: If only this were true 😂

triarius (approved):
Nice avoidance of duplication

pages/agent/v3/_linux_multiple_agents.md
DrJosh9000: Details, details...

PR #2300, opened 2023-07-14
Remove extra full stop on agent docker page

hannahcancode:
- as it says on in the title, double full stop (not sure if it should go inside or outside the `a` tag though?

### DIFF
--- a/pages/agent/v3/docker.md
+++ b/pages/agent/v3/docker.md
@@ -21,7 +21,7 @@ docker run -d -t --name buildkite-agent buildkite/agent:3-ubuntu start --token "
 ```
 
 >🚧 Caveats for builds that need Docker access.
-> If your build jobs require Docker access, and you're passing through the Docker socket, you must ensure the build path is consistent between the Docker host and the agent container. See <a href="#allowing-builds-to-use-docker">Allowing builds to use Docker</a> for more details</a>.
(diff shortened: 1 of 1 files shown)


pages/agent/v3/docker.md
```diff
@@ -21,7 +21,7 @@ docker run -d -t --name buildkite-agent buildkite/agent:3-ubuntu start --token "
 ```
 
 >🚧 Caveats for builds that need Docker access.
-> If your build jobs require Docker access, and you're passing through the Docker socket, you must ensure the build path is consistent between the Docker host and the agent container. See <a href="#allowing-builds-to-use-docker">Allowing builds to use Docker</a> for more details.</a>.
+> If your build jobs require Docker access, and you're passing through the Docker socket, you must ensure the build path is consistent between the Docker host and the agent container. See <a href="#allowing-builds-to-use-docker">Allowing builds to use Docker</a> for more details</a>.
```
mbelton-buildkite, line 24: 👍🏻 Full-stop outside the tag, but unless I'm missing something, it also looks like that closing tag shouldn't be there:

```suggestion
> If your build jobs require Docker access, and you're passing through the Docker socket, you must ensure the build path is consistent between the Docker host and the agent container. See <a href="#allowing-builds-to-use-docker">Allowing builds to use Docker</a> for more details.
```

mbelton-buildkite (approved):
Thanks for fixing this. I've left a "yes, and" 😊

pages/agent/v3/docker.md
hannahcancode: Ah, good catch!

PR #2306, opened 2023-07-17
Add Update the revoke inactive tokens after to GraphQL cookbook

JuanitoFatas:
Update the feature name

CleanShot 2023-07-24 at 12 30 00@2x

Add cookbook for how to use [revoke inactive tokens after](https://buildkite.com/releases/2023-06/api-token-expiry-policies) via GraphQL API.

CleanShot 2023-07-24 at 12 30 20@2x

### DIFF
--- a/pages/apis/graphql/graphql_cookbook.md
+++ b/pages/apis/graphql/graphql_cookbook.md
@@ -691,6 +691,23 @@ mutation UpdateSessionDuration {
 }
 ```
 
+### Update inactive API token revocation
+
+On the Enterprise plan, you can control when inactive API tokens are revoked. By default, they are never (`NEVER`) revoked, but you can set your token revocation to either 30, 60, 90, 180, or 365 days.
+
+```graphql
+mutation UpdateRevokeInactiveTokenPeriod {
+  organizationRevokeInactiveTokensAfterUpdate(input: {
+    organizationId: "organization-id",
+    revokeInactiveTokensAfter: DAYS_30
+  }) {
+    organization {
+      revokeInactiveTokensAfter
+    }
+  }
+}
+```
+
 ### Pin SSO sessions to IP addresses
 
 You can require users to re-authenticate with your SSO provider when their IP address changes with the following call, replacing `ID` with the GraphQL ID of the SSO provider:
--- a/pages/apis/managing_api_tokens.md
+++ b/pages/apis/managing_api_tokens.md
@@ -66,14 +66,14 @@ If you'd like to limit access to your organization by IP address, you can create
 
 You can also manage the allowlist with the [`organizationApiIpAllowlistUpdate`](/docs/apis/graphql/schemas/mutation/organizationapiipallowlistupdate) mutation in the GraphQL API.
 
-## Revoking inactive tokens automatically
+## Inactive API tokens revocation
 
 > 📘 Enterprise feature
-> Revoking inactive tokens automatically is only available on an [Enterprise](https://buildkite.com/pricing) plan.
+> Revoking inactive API tokens automatically is only available on an [Enterprise](https://buildkite.com/pricing) plan.
 
-To enable the automatic revocation of inactive tokens, navigate to your [organization's security settings](https://buildkite.com/organizations/~/security) and specify the maximum timeframe for inactive tokens to remain valid.
+To enable the inactive API tokens revocation, navigate to your [organization's security settings](https://buildkite.com/organizations/~/security) and specify the maximum timeframe for inactive tokens to remain valid.
 
-Inactive tokens refer to those that have not been used within the specified duration. When an API token surpasses the configured setting, Buildkite will automatically revoke the token's access to your organization.
+Inactive API tokens refer to those that have not been used within the specified duration. When an API token surpasses the configured setting, Buildkite will automatically revoke the token's access to your organization.
 
 Upon token revocation, Buildkite will notify the owner of their change in access.
 


pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
 }
 ```
 
+### Update the revoke inactive tokens after
+
+You can control how many days to revoke inactive API access tokens. By default that's `NEVER` revoke inactive API access tokens, but you can set to 30, 60, 90, 180, 365 days. This is only available to Enterprise customers.
```
mbelton-buildkite, line 683: General tightening + I think it's good to be upfront about the Enterprise restriction:

```suggestion
On the Enterprise plan, you can control when inactive API access tokens are revoked. By default, they are never (`NEVER`) revoked, but you can set your token expiry policy to either 30, 60, 90, 180, or 365 days.
```

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
...
+
+```graphql
+mutation UpdateRevokeInactiveTokenAfter {
```
mbelton-buildkite, line 686: I think it's better if the label can stand on its own. Maybe:

```suggestion
mutation UpdateRevokeInactiveTokenPeriod {
```

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
...
+  organizationRevokeInactiveTokensAfterUpdate(input: {
```
mbelton-buildkite, line 687: Is it too late to update the name of this? I'd go with `organizationRevokeInactiveTokenPeriodUpdate` orrr if it's not an issue to remove "organization" then `revokeInactiveTokenPeriodUpdate`.

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
 }
 ```
 
+### Update the revoke inactive tokens after
```
mbelton-buildkite, line 681: Ending on "after" here doesn't feel complete. How about:

```suggestion
### Update when inactive access tokens are revoked
```

pages/apis/graphql/graphql_cookbook.md
JuanitoFatas: It is following the convention ([internal link](https://github.com/buildkite/buildkite/tree/main/app/models/api/graph/mutations)), so we may need to revisit the mutation naming if we really want to rename them!

mbelton-buildkite: Hmmmm I might be missing the convention you mean, but it looks like the current name doesn't match the others. The main convention I see is:

> [thing] + [action]

- clusterAgentToken + Update
- jobTypeCommand + Cancel
- pipelineSchedule + Delete

I see there are some exceptions, but most look to follow that pattern.

"organizationRevokeInactiveTokensAfter" is not a thing. Changing that part to "organizationRevokeInactiveTokensPeriod" would make it a thing and follow the convention.

Does that make sense?

JuanitoFatas: > [thing] + [action]

Yup.

`OrganizationRevokeInactiveTokensAfter` + `UpdateMutation` same as
`OrganizationAPIIPAllowlist` + `UpdateMutation`. 

They all started with `Organization[thing]` + `action`.


The field is not called `Period` [on UI](https://buildkite.com/organizations/~/security), so was keeping the same as the UI. If you strongly feel we should rename all to Period (which is possible), could you raise an escalation on Foundation team? Thanks!

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
 }
 ```
 
+### Update when inactive access tokens are revoked
```
JuanitoFatas, line 681:
```suggestion
### Update API access tokens revocation policy
```

@mbelton-buildkite Hello, we are planning to [name this feature](https://github.com/buildkite/buildkite/pull/13085#discussion_r1270106986), feel free to leave suggestion 🙏🏻

pages/apis/graphql/graphql_cookbook.md
mbelton-buildkite: Got it, thanks! I'll raise an escalation to see how others feel. 😊

pages/apis/graphql/graphql_cookbook.md
mbelton-buildkite: I'll have a look at that link. I'm 50/50 on that suggestion. While accurate, I think it sounds more complex than it needs to. Ideally, I'd like these headings to match the words someone might use to describe the task they want to do. They shouldn't have to know the feature name. 🤔

Have a think about that and I'm happy for you to update it if you think it helps more.

pages/apis/graphql/graphql_cookbook.md
```diff
@@ -678,6 +678,23 @@ mutation UpdateSessionDuration {
...
+
+On the Enterprise plan, you can control when inactive API access tokens are revoked. By default, they are never (`NEVER`) revoked, but you can set your token expiry policy to either 30, 60, 90, 180, or 365 days.
```
mbelton-buildkite, line 683: Probably best to update this if the feature name is changing:

```suggestion
On the Enterprise plan, you can control when inactive API access tokens are revoked. By default, they are never (`NEVER`) revoked, but you can set your token revocation policy to either 30, 60, 90, 180, or 365 days.
```

mbelton-buildkite (approved):
There are a couple of decisions to make, but I think this is looking good in general. I'll also tag you in the escalation with Foundation :)

PR #2307, opened 2023-07-17
correct BUILDKITE_REPO description to be modifiable

HugeIRL:
As seen as far back as here: https://github.com/buildkite/agent/blob/a4ee642d625171b69c255a663a8aae4f0737334d/CHANGELOG.md?plain=1#L1069C22-L1069C22

BUILDKITE_REPO has been modifiable since agent version 3.7.0. Our current docs state it is not modifiable, which
has caused some confusion.


pzeballos:
Just in case, I wouldn't change it to true @HugeIRL , but I would add a clarification that it can only be updated via a hook

HugeIRL:
Closing PR, we discussed a larger change potentially being investigated.

PR #2309, opened 2023-07-17
Fix typo: modifable -> modifiable

mbelton-buildkite:
@HugeIRL [pointed out this typo](https://3.basecamp.com/3453178/buckets/28474015/chats/5132790929@6357316472), which makes the field display `false` even if the value is meant to be `true`.

I've raised an escalation with Pipelines to get a review.


sj26 (approved):
Typo fix, should be fine?

pda (approved):

PR #2315, opened 2023-07-19
Update permissions.md

stephanieatte:
Ticket: https://secure.helpscout.net/conversation/2286383801/53474/

The customer was confused about permissions for teams and pipelines. Its best to clarify about the order the precedence so they know the expected behaviour

### DIFF
--- a/pages/pipelines/permissions.md
+++ b/pages/pipelines/permissions.md
@@ -135,6 +135,8 @@ From the Member Permissions page, organization admins can toggle whether or not 
 
 If your organization has teams enabled, the pipeline creation permissions are managed at a team level. Pipeline creation permission controls can be found on the Teams Settings page. Without teams enabled, the pipeline creation permission control can be found on the Member Permissions page.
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/permissions.md
```diff
@@ -135,6 +135,8 @@ From the Member Permissions page, organization admins can toggle whether or not
 
 If your organization has teams enabled, the pipeline creation permissions are managed at a team level. Pipeline creation permission controls can be found on the Teams Settings page. Without teams enabled, the pipeline creation permission control can be found on the Member Permissions page.
 
+Note:  The Team-Level permissions on a given pipeline, will override the lower priviledge member permisions in the team settings.
```
mbelton-buildkite, line 138: I find this quite hard to parse. Could you try expanding it a bit more or perhaps providing an example? Thinking like:

> Note that team-level permissions override member permissions. For example, if the same user has a team level permission to X on a pipeline but only a X member permission on that pipeline, they can...

PR #2320, opened 2023-07-21
Clusters: Update agent connection instructions to include token creation

l-suzuki:
Linear ticket: [EDGE-1055](https://linear.app/buildkite/issue/EDGE-1055/ui-hide-agentregistrationtokentoken-and-clustertokentoken)

Updates the docs to reflect the new cluster token creation flow, as per the changes in [this PR](https://github.com/buildkite/buildkite/pull/13136/files).

### DIFF
--- a/pages/clusters/manage_clusters.md
+++ b/pages/clusters/manage_clusters.md
@@ -37,7 +37,7 @@ To connect an agent:
 1. Enter a description.
 1. Select _Create Token_.
 1. Select _Copy to Clipboard_ and save the token somewhere secure.
-1. Select _Okay, I'm done!_.
+1. Select _Okay, I'm done!_
 1. [Use the token](/docs/agent/v3/tokens#using-and-storing-tokens) with the relevant agents, along with [the key from the relevant cluster queue](/docs/agent/v3/queues#setting-an-agents-queue).
 
 You can also create, edit, and revoke other agent tokens from the cluster’s _Agent tokens_.
--- a/pages/clusters/manage_clusters.md
+++ b/pages/clusters/manage_clusters.md
@@ -33,7 +33,11 @@ Agents are associated with a cluster through the cluster’s agent tokens.
 To connect an agent:
 
 1. Navigate to the cluster's _Agent tokens_.
-1. Copy the auto-generated token.
+1. Select _New Token_.
+1. Enter a description.
+1. Select _Create Token_.
+1. Select _Copy to Clipboard_ and save the token somewhere secure.
+1. Select _Okay, I'm done!_.
 1. [Use the token](/docs/agent/v3/tokens#using-and-storing-tokens) with the relevant agents, along with [the key from the relevant cluster queue](/docs/agent/v3/queues#setting-an-agents-queue).
 
 You can also create, edit, and revoke other agent tokens from the cluster’s _Agent tokens_.
--- a/pages/clusters/overview.md
+++ b/pages/clusters/overview.md
@@ -20,6 +20,7 @@ Enabling clusters also changes access to agent tokens. Rather than being availab
 
 To enable clusters:
 
+1. Securely save any existing agent tokens you need because these won't be available after enabling clusters.
 1. Navigate to your [organization’s pipeline settings](https://buildkite.com/organizations/~/pipeline-settings).
 1. In _Clusters_, select _Enable Clusters_.
 
--- a/pages/clusters/overview.md
+++ b/pages/clusters/overview.md
@@ -18,7 +18,6 @@ Any Buildkite administrator can enable clusters for an organization. Once you en
 
 Enabling clusters also changes access to agent tokens. Rather than being available in the Buildkite dashboard, agent tokens are only visible upon creation, ensuring greater security for your applications.
 
-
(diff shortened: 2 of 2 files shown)


l-suzuki:
@mbelton-buildkite it's outside the scope of this PR, but this change got me thinking about whether there should be an "Unclustered" section in the Clusters docs. 

I'm also wondering if we need to put a warning on the "Enable clusters" toggle to let users know that if they are currently retrieving their Agent Tokens via the UI, they will no longer be able to do this once they switch to clusters... it'd be pretty annoying if you switched over assuming you'd still have access to those values, only to find they're gone!

pages/clusters/manage_clusters.md
```diff
@@ -33,7 +33,11 @@ Agents are associated with a cluster through the cluster’s agent tokens.
 To connect an agent:
 
 1. Navigate to the cluster's _Agent tokens_.
-1. Copy the auto-generated token.
+1. Select _New Token_.
+1. Enter a description.
+1. Select _Create Token_.
+1. Select _Copy to Clipboard_ and save the token somewhere secure.
+1. Select _Okay, I'm done!_.
```
mbelton-buildkite, line 40: Style: Just use the punctuation from the button text:

```suggestion
1. Select _Okay, I'm done!_
```

mbelton-buildkite (approved):
One minor nitpick, otherwise looks good :)

mbelton-buildkite:
> it's outside the scope of this PR, but this change got me thinking about whether there should be an "Unclustered" section in the Clusters docs.

That sounds like a good idea, but I don't think I know enough to suggest the approach. Happy to either have a chat to work it out or go through your thoughts in a PR. 😊

mbelton-buildkite:
> I'm also wondering if we need to put a warning on the "Enable clusters" toggle to let users know that if they are currently retrieving their Agent Tokens via the UI, they will no longer be able to do this once they switch to clusters... it'd be pretty annoying if you switched over assuming you'd still have access to those values, only to find they're gone! Maybe we could put this on the Clusters overview page somewhere? I'd be keen for advice on how to word it though, we dont want to deter people from switching over...

Good call. I think we could include it in the intro text for the Enable clusters section on the Overview page. I think that's where you need to know that information but it's not jumping out as a warning. Thinking something like:

> Any Buildkite administrator can enable clusters for an organization. Once you enable clusters, you can only disable them by contacting support.
>
> Enabling clusters also changes how you access agent tokens. Rather than being available in the Buildkite dashboard, you need to retrieve agent tokens using the API.

^ I made up the API part to illustrate the point but maybe it's true haha.

l-suzuki:
> > I'm also wondering if we need to put a warning on the "Enable clusters" toggle to let users know that if they are currently retrieving their Agent Tokens via the UI, they will no longer be able to do this once they switch to clusters... it'd be pretty annoying if you switched over assuming you'd still have access to those values, only to find they're gone! Maybe we could put this on the Clusters overview page somewhere? I'd be keen for advice on how to word it though, we dont want to deter people from switching over...
> 
> Good call. I think we could include it in the intro text for the Enable clusters section on the Overview page. I think that's where you need to know that information but it's not jumping out as a warning. Thinking something like:
> 
> > Any Buildkite administrator can enable clusters for an organization. Once you enable clusters, you can only disable them by contacting support.
> > Enabling clusters also changes how you access agent tokens. Rather than being available in the Buildkite dashboard, you need to retrieve agent tokens using the API.
> 
> ^ I made up the API part to illustrate the point but maybe it's true haha.

So the difference is that tokens will only be available upon creation. So if they're migrating to clusters they'll need to make sure they've got their existing agent tokens saved somewhere secure, as they wont be able to access them from the UI any more. 
I've added a sentence into the overview - let me know what you think!

mbelton-buildkite:
> So the difference is that tokens will only be available upon creation. So if they're migrating to clusters they'll need to make sure they've got their existing agent tokens saved somewhere secure, as they wont be able to access them from the UI any more.
> I've added a sentence into the overview - let me know what you think!

Ahhhhh, that's a bigger deal than I realised. I think our options are to use a warning callout or include it in the steps to enable. I think putting it in the steps might be a better way to avoid it being a blocker to enabling. Thinking:

```
To enable clusters:

1. Securely save any existing agent tokens you need because these won't be available after enabling clusters.
1. Navigate to...
```

What do you think?

l-suzuki:
> > So the difference is that tokens will only be available upon creation. So if they're migrating to clusters they'll need to make sure they've got their existing agent tokens saved somewhere secure, as they wont be able to access them from the UI any more.
> > I've added a sentence into the overview - let me know what you think!
> 
> Ahhhhh, that's a bigger deal than I realised. I think our options are to use a warning callout or include it in the steps to enable. I think putting it in the steps might be a better way to avoid it being a blocker to enabling. Thinking:
> 
> ```
> To enable clusters:
> 
> 1. Securely save any existing agent tokens you need because these won't be available after enabling clusters.
> 1. Navigate to...
> ```
> 
> What do you think?

Looks good to me, I'll make that change now

mbelton-buildkite (approved):
Looks good to me :)

PR #2321, opened 2023-07-24
Clarify the behavior of using multiple stacks per queue

mbelton-buildkite:
@catkins wrote an [internal post](https://3.basecamp.com/3453178/buckets/23112918/messages/6363965965) about scaling up redundant agents if you use multiple elastic stacks per queue. This change aims to use a stronger recommendation not to do that and clarify the behavior if you do.

### DIFF
--- a/pages/agent/v3/elastic_ci_aws/managing_elastic_ci_stack.md
+++ b/pages/agent/v3/elastic_ci_aws/managing_elastic_ci_stack.md
@@ -68,7 +68,7 @@ If you configure `MinSize` < `MaxSize` in your AWS autoscaling configuration, th
 
 This means you can scale down to zero when idle, which means you can use larger instances for the same cost.
 
-Metrics are collected with a Lambda function, polling every 10 seconds based on the queue the stack is configured with. The autoscaler monitors only one queue, and the monitoring drives the scaling of the stack. You should only use one Elastic CI Stack for AWS per queue to avoid scaling up redundant agents. If you target the same queue with multiple stacks, each stack will scale up to handle the work.
+Metrics are collected with a Lambda function, polling every 10 seconds based on the queue the stack is configured with. The autoscaler monitors only one queue, and the monitoring drives the scaling of the stack. You should only use one Elastic CI Stack for AWS per queue to avoid scaling up redundant agents. If you target the same queue with multiple stacks, each stack will independently scale up additional agents as if it were the only stack running, leading to over provisioning.
 
 ## Terminating the instance after the job is complete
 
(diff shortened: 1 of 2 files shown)


pages/agent/v3/elastic_ci_aws/managing_elastic_ci_stack.md
```diff
@@ -68,7 +68,7 @@ If you configure `MinSize` < `MaxSize` in your AWS autoscaling configuration, th
 
 This means you can scale down to zero when idle, which means you can use larger instances for the same cost.
 
-Metrics are collected with a Lambda function, polling every 10 seconds based on the queue the stack is configured with. The autoscaler monitors only one queue, and the monitoring drives the scaling of the stack. This means that usually you need one Elastic CI Stack for AWS per queue.
+Metrics are collected with a Lambda function, polling every 10 seconds based on the queue the stack is configured with. The autoscaler monitors only one queue, and the monitoring drives the scaling of the stack. You should only use one Elastic CI Stack for AWS per queue to avoid scaling up redundant agents. If you target the same queue with multiple stacks, each stack will scale up to handle the work.
```
catkins, line 71: Hmm, trying to think of how to best articulate this concisely 🤔 The key issue is that the stacks are unaware of each other, so will each try to "scale out to handle the work", but over-provision as it can't know that other stacks are _also_ scaling out.

```suggestion
Metrics are collected with a Lambda function, polling every 10 seconds based on the queue the stack is configured with. The autoscaler monitors only one queue, and the monitoring drives the scaling of the stack. You should only use one Elastic CI Stack for AWS per queue to avoid scaling up redundant agents. If you target the same queue with multiple stacks, each stack will independently scale up additional agents as if it were the only stack running, leading to over provisioning.
```

@triarius might have some better wordy-words?

catkins (approved):
👍 thanks for looping back to this @mbelton-buildkite 👏

triarius (approved):
LGTM 🎉

triarius:
@mbelton-buildkite this might be good to add to a FAQ. Working back from the effect to the cause, it would be something like:

> Why are my stacks are scaling out but not running any jobs?
> If you have multiple Elastic CI Stacks, check that they are listening to unique queues. This is determined by the `BuildkiteQueue` parameter. Each Elastic CI Stack that you launch through CloudFormation should have a unique value for this paramter. Otherwise, each of them will scale out independantly to service all the jobs on the queue, but the jobs will distributed amongst them. This will mean that your stacks are overprovisioned. This could also happen if you have buildkite agents that are not part of an Elastic CI Stack [started with a tag](https://buildkite.com/docs/agent/v3/cli-start#tags) of the form `queue=<name of queue>`. Any agents that are started like this could compete with and Elastic CI Stack for jobs, but the Elastic CI Stack will scale out as if this competition did not exist.

triarius (approved):

mbelton-buildkite:
> @mbelton-buildkite this might be good to add to a FAQ

@triarius, I've added a section in the troubleshooting page using the content you suggested

triarius (approved):
Amazing!

PR #2322, opened 2023-07-24
Update stack template for v5.22.1 and remove details of versions of installed software

triarius:
The version numbers that have been removed are horribly out of date. It is a
large burden to update them all, so we are removing them.

We will investigate a better way to surface this to our customers in the future,
but in the meantime, we will remove them as they are misleading.

So for now, customers will need to follow the release announcements in the repository.


mbelton-buildkite (approved):
Looks good. Thanks for mentioning where someone can find the extra info.

PR #2334, opened 2023-07-27
Add graphql_id to suites REST API

gchan:
PIE-1880

To manage team permissions for suites via the Terraform provider, we
should support a graphql_id attribute on the suites REST API response.


mbelton-buildkite (approved):
Looks good to me :)

KatieWright26 (approved):

PR #2336, opened 2023-07-28
Document a way to deny jobs based on the enviroment variables set for them

triarius:


pages/agent/v3/securing.md
```diff
@@ -104,6 +102,22 @@ do
 done
 ```
 
+You can see from the previous example that `$BUILDKITE_ENV_FILE` is the location of file that contains the environment variables that the control plane passes to a job. You may use this to block jobs from executing if certain environment variables are set.
```
mbelton-buildkite, line 105: I think this should connect to the codeblock it introduces. A couple of options would be:

1. Add a colon:

```suggestion
You can see from the previous example that `$BUILDKITE_ENV_FILE` is the location of file that contains the environment variables that the control plane passes to a job. You may use this to block jobs from executing if certain environment variables are set:
```

2. Introduce the example:

```suggestion
You can see from the previous example that `$BUILDKITE_ENV_FILE` is the location of file that contains the environment variables that the control plane passes to a job. You may use this to block jobs from executing if certain environment variables are set. For example, the following `pre-bootstrap` hook blocks a job from executing if the `ENVIRONMENT_VARIABLE_TO_DENY` environment variable is set.
```

pages/agent/v3/securing.md
```diff
@@ -104,6 +102,22 @@ do
...
+
+```bash
+#!/bin/bash
+
+set -euo pipefail
+
+if grep '^ENVIRONMENT_VARIABLE_TO_DENY=' "$BUILDKITE_ENV_FILE" > /dev/null
+then
+  echo "Rejecting job because it the environment variable ENVIRONMENT_VARIABLE_TO_DENY has been set"
```
mbelton-buildkite, line 114: Typo:

```suggestion
  echo "Rejecting job because the environment variable ENVIRONMENT_VARIABLE_TO_DENY has been set"
```

pages/agent/v3/securing.md
```diff
@@ -104,6 +102,22 @@ do
...
+  exit 1
+fi
+```
+
+But remember to that some [environment variables may be essential](https://buildkite.com/docs/pipelines/environment-variables) to the execution of jobs, so deny-listing them in this manner is not advisable.
```
mbelton-buildkite, line 119: This is mainly personal preference, so feel free to disagree. The one part that should change is using a relative link to the docs page.

```suggestion
But also remember that some [environment variables may be essential](/docs/pipelines/environment-variables) to the execution of jobs, so adding them to a blocklist in this manner is not advisable.
```

mbelton-buildkite (approved):

PR #2351, opened 2023-08-04
Remove web archive link to Ruby best practices

mbelton-buildkite:
We run a link checker as part of linting the docs, and this link keeps causing it to fail by returning a 429. We don't want Google penalizing us for broken outbound links. From my understanding of the paragraph, we're mainly linking there to add legitimacy to our suggestion, but our suggestion provides the solution. If that's true, I don't think we need this link.

If I'm missing something, is there another resource we could link to or some information from that page we could pull into our docs?


dannymidnight (approved):
Makes sense to me.

mbelton-buildkite:
(I've raised an escalation with TA to get a review)

blaknite (approved):
Makes sense to me too. As a Ruby shop, I think we can lean on our expertise and do this without a source.

PR #2363, opened 2023-08-10
Update sso.md with new entry in FAQ section for login related errors

nsuma8989:
Updated faq section with couple of errors users commonly encounter and provided guidance on how to approach those errors.


nsuma8989:
@mbelton-buildkite @pzeballos Please can you review this PR

nsuma8989:
> @mbelton-buildkite @pzeballos Please can you review this PR

Gentle remainder on this PR review request

pages/integrations/sso.md
```diff
@@ -141,3 +141,10 @@ Both authentication and authorization are necessary because SSO using one Buildk
...
+There are few reasons why error "this email is already being used by another user" can occur. One such possibility is you are using shared accounts. The email is associated with another account. You will need to remove that association from your Email Personal Settings.
+
+Other possibility is the account already exists in Buildkite but probably forgot about its existence. The faster and easier solution is to delete the account but sometimes that may not resolve the issue as there might be some SSO authorization records on Buildkite end for that account which needs to be cleaned up. In scenarios where even after deleting the account and adding the user does not resolve the issue, please reach out to support@buildkite.com for assistance.
+
+For scenarios, where you get error "we couldn't find an account with that email address" it is likely due to trying to log in from the wrong place. You need to log in to https://buildkite.com/sso and follow the link from the email you received. If the issue still persists, please reach out to support@buildkite.com for assistance.
```
mbelton-buildkite, line 150: I think this should be split into a separate question for each error:

```suggestion
### Why do I get the error "this email is already being used by another user" when logging in?

There are two common reasons. The first is that you are using shared accounts, so the email is associated with another account. To resolve that, you need to remove the association from your Email Personal Settings.

The second is that the account already exists in Buildkite. If you have access to the old account, delete it before continuing. You may also need to clean up any SSO authorization records on Buildkite for the old account. If that doesn't resolve the issue or you don't have access to the account, please reach out to support@buildkite.com for assistance.

### Why do I get the error "we couldn't find an account with that email address" when logging in?

This is likely caused by trying to log in from the wrong place. You need to log in from https://buildkite.com/sso and follow the link from the email you receive. If the issue persists, please reach out to support@buildkite.com for assistance.
```

mbelton-buildkite: Is there a link we can provide to Email Personal Settings?

mbelton-buildkite: For the second reason in the first question, should we recommend they try to recover the account using the forgot password flow?

nsuma8989: > For the second reason in the first question, should we recommend they try to recover the account using the forgot password flow?

No, in this case they need to login through SSO link we provided in the answer. Forgot password flow wont work in this case.

nsuma8989: > 

No, we wont be able to provide this as it is specific to their email provider

nsuma8989:
@mbelton-buildkite I made changes as suggested, please take a look

mbelton-buildkite (approved):

PR #2365, opened 2023-08-10
Update input_step.md

nsuma8989:
Goal of this PR is to add a note to let users know that input step fields should not be used to pass secret information

### DIFF
--- a/pages/pipelines/input_step.md
+++ b/pages/pipelines/input_step.md
@@ -44,7 +44,7 @@ For an example pipeline, see the [Input step example pipeline](https://github.c
 
 <a class="Docs__example-repo" href="https://github.com/buildkite/input-step-example"><span class="detail">:pipeline: Input Step Example Pipeline</span> <span class="repo">github.com/buildkite/input-step-example</span></a>
 
->📘 Input step should not be used to store sensitive information like secrets. 
+>📘 Input step should not be used to store sensitive information like secrets.
 
 ## Input step attributes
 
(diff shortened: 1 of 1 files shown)


nsuma8989:
@mbelton-buildkite Please can you review this PR and share your feedback

nsuma8989:
> @mbelton-buildkite Please can you review this PR and share your feedback

@mbelton-buildkite Gentle remainder on this review request

pages/pipelines/input_step.md
```diff
@@ -44,6 +44,8 @@ For an example pipeline, see the [Input step example pipeline](https://github.c
 
 <a class="Docs__example-repo" href="https://github.com/buildkite/input-step-example"><span class="detail">:pipeline: Input Step Example Pipeline</span> <span class="repo">github.com/buildkite/input-step-example</span></a>
 
+>📘 Input step should not be used to store sensitive information like secrets.
```
mbelton-buildkite, line 47: I think we should use the warning format for this, and separate out a title from the body text. In the body, it would be good to add the reason why you shouldn't do this. Please correct this, but thinking something like:

```suggestion
>🚧 Don't store sensitive data in input steps
> You shouldn't use input steps to store sensitive information like secrets because the data is available in subsequent steps.
```

nsuma8989: @mbelton-buildkite Updated with suggestion but with slight modification to see data will be stored in build metadata

mbelton-buildkite (approved):

PR #2368, opened 2023-08-13
Update ubuntu first step wording

sj26:
The first paragraph is introductory, the second is the first action, so saying "Next, ..." doesn't make sense to me — unless I'm missing something!


sj26:
Also on this page — 

https://buildkite.com/docs/agent/v3/ubuntu

image

This copy button should be careful not to include a newline so that the line can be edited to include the token before being run. Running it again will require changing the pattern being replaced etc.

mbelton-buildkite (approved):

PR #2371, opened 2023-08-14
UPD about concurrency groups

karensawrey:
A long-overdue upd following https://secure.helpscout.net/conversation/2306810185/54487/
Discussed and approved by @jeremybumsted

### DIFF
--- a/pages/pipelines/build_matrix.md
+++ b/pages/pipelines/build_matrix.md
@@ -11,7 +11,8 @@ The following [command step](/docs/pipelines/command-step) attributes can contai
 * [plugins](/docs/pipelines/command-step#plugins)
 * [agents](/docs/pipelines/command-step#agents)
 
-You can't use matrix values in other attributes, including step keys.
+You can't use matrix values in other attributes, including step keys, but you can use them in [concurrency groups](/docs/pipelines/controlling-concurrency#concurrency-groups).
+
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/build_matrix.md
```diff
@@ -11,7 +11,8 @@ The following [command step](/docs/pipelines/command-step) attributes can contai
 * [plugins](/docs/pipelines/command-step#plugins)
 * [agents](/docs/pipelines/command-step#agents)
 
-You can't use matrix values in other attributes, including step keys.
+You can't use matrix values in other attributes, including step keys, but you can use them in [concurrency groups](/docs/pipelines/controlling-concurrency#concurrency-groups).
```
mbelton-buildkite, line 14: Can I check the logic. On Helpscout, I see "having concurrency groups inside of a matrix is not possible." Then, I think this says you can have matrix values inside concurrency groups. So you can't have a concurrency group in a matrix, but you can have a matrix in a concurrency group. Is that right?


Is this right? On Helpscout, I see  But this says you can?

karensawrey: Hi, @mbelton-buildkite! "So you can't have a concurrency group in a matrix, but you can have a matrix in a concurrency group. Is that right?" - yes, this is correct according to my discussion with @jeremybumsted. I also had to double-check as this is a confusing point.
Plz re-review.

mbelton-buildkite (approved):
Thanks for clarifying! This one was confusing 🫠

PR #2375, opened 2023-08-15
Add page meta descriptions

dannymidnight:
Adds meta descriptions to all pages with content pulled from YAML front matter or falling back to the first paragraph.

Closes ONB-317, ONB-326

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -161,21 +161,33 @@ class Page
     @name.to_s.gsub(/[^0-9a-zA-Z\-\_\/]/, '').underscore
   end
 
-  def metadata
-    defaults = {
-      # Default to rendering table of contents
-      "toc": true,
-      # Default to H3s being included in the table of contents
-      "toc_include_h3": true
-    }
-    if file.front_matter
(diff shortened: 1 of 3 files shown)


app/models/page.rb
```diff
@@ -161,22 +157,39 @@ def basename
...
-      "toc_include_h3": true
-    }
-    if file.front_matter
-      defaults.merge(file.front_matter.symbolize_keys)
-    else
-      defaults
-    end
+  # Page description, either from front matter or extracted from the markdown
+  def description
+    front_matter.fetch(:description, extracted_data.fetch("shortDescription"))
```
mbelton-buildkite, line 162: Does this mean the frontmatter field is `description`? I wonder if it should be `meta_description` just so its purpose is clearer 🤔

mbelton-buildkite (approved):
🙌🏻 Yay! One question/suggestion but minor

app/models/page.rb
dannymidnight: I think `description` is pretty clear and allows for other uses – for example, it's also used as the value of the `og:description` tag. Yeah ok, that's also a meta tag but you get the drift!

PR #2376, opened 2023-08-15
Add unique title tags to GraphQL pages

dannymidnight:
- Remove some duplicate titles by adding unique titles for each of the GraphQL pages.
- Add meta descriptions to each of the pages.

Closes DOC-627


scripts/graphql_api_content/render_helpers.rb
```diff
@@ -213,6 +213,8 @@ def render_page(schema_type_data)
 
     page_content = <<~HTML.strip
       ---
+      title: #{name} | #{namespace} | GraphQL API Reference
```
mbelton-buildkite, line 216: Any particular reason for using `|` between each? I was expecting `-` and the just `|` between this and `Buildkite Documentation` (For example: `agent - Queries - GraphQL API | Buildkite Documentation`). Maybe that's unnecessarily complicated.

I also think we can drop "Reference" to keep the title slightly shorter. Thoughts?

dannymidnight: No reason other than for aesthetics! I'll update.

dannymidnight: OK cool, I've updated the formatting in https://github.com/buildkite/docs/pull/2376/commits/193b445aaa15ee6b1a0a430e5efa706059568837. Worth noting ` | Buildkite Documentation` is appended to all page titles which is why it's not included in that `title` attribute. 

https://github.com/buildkite/docs/blob/193b445aaa15ee6b1a0a430e5efa706059568837/app/views/application/_head.html.erb#L8

PR #2385, opened 2023-08-17
Add Datadog RUM performance monitoring

dannymidnight:
As part of a recent SEO audit we've added some tasks in our backlog to improve page performance. Before tackling any of those tasks I'd first like to get a better handle on how the pages are actually performing to better understand the impact of any future changes. 

As a bit of an experiment, this PR integrates the docs with [Datadog Real User Monitoring](https://docs.datadoghq.com/real_user_monitoring/browser/monitoring_page_performance/#overview) to measure page performance and give us a baseline to improve on.

I've disabled most parts of the integration (ie. session tracking etc.) since the focus is on page performance.


dannymidnight:
Hmm.. 🤔 I wonder how those previews are going to interact with this. I might need to switch out the environment for those pages.

mbelton-buildkite (approved):
Metrics!

mbelton-buildkite:
> Hmm.. 🤔 I wonder how those previews are going to interact with this. I might need to switch out the environment for those pages. They would definitely muddy the results since those pages should be loading much faster.

💯 Prod is the priority, but would be interesting to know how much better the static site performs 🤔

PR #2389, opened 2023-08-18
Explicit dependencies on a wait step can break the wait intention 

lizrabuya:
We received a customer feedback where jobs following a wait step were ran without waiting for previous steps to continue as opposed to what we say in the docs on wait steps. After investigation, it was the caused by the use of `depends_on` attribute with a missing dependency. Details at this HS ticket [56014]( https://secure.helpscout.net/conversation/2334371567/56014/) .

### DIFF
--- a/pages/pipelines/wait_step.md
+++ b/pages/pipelines/wait_step.md
@@ -45,6 +45,10 @@ _Optional attributes:_
   </tr>
 </table>
 
+>🚧 <code>depends_on</code> and <code>wait</code>
+> Use explicit dependencies in a wait step  with care. A wait step with the <code>depends_on</code> attribute might break the intention of waiting when you have missing dependencies. Jobs following the wait step no longer waits for previous steps to complete to continue if their dependencies are missing.
+
+
 ## Conditional wait
 
 You can use a conditional to only wait when certain conditions are met:


pages/pipelines/wait_step.md
```diff
@@ -45,6 +45,10 @@ _Optional attributes:_
   </tr>
 </table>
 
+>🚧 <code>depends_on</code> and <code>wait</code>
+> Use explicit dependencies in a wait step  with care. A wait step with the <code>depends_on</code> attribute might break the intention of waiting when you have missing dependencies. Jobs following the wait step no longer waits for previous steps to complete to continue if their dependencies are missing.
```
mbelton-buildkite, line 49: If I understand this correctly, the key point is that using `depends_on` means a wait step no longer waits for any proceeding step that wasn't named to complete. I think the information in the table could be updated rather than an extra callout here. Maybe:

```html
<tr>
    <td><code>depends_on</code></td>
    <td>
      A list of step keys that this step depends on. This step will only proceed after the named steps have completed. It will not wait for any proceeding steps to complete that weren't named. See <a href="/docs/pipelines/dependencies">managing step dependencies</a> for more information.<br>
      <em>Example:</em> <code>"test-suite"</code>
    </td>
</tr>
```

What do you think?

PR #2392, opened 2023-08-21
Doc 654 add reference to queues on branch configurations

clbarrell:
Purpose of this was to add **Agent Queues** as an alternative way of filtering pipelines and steps from being done – and where (agents) they are completed.

Then I added another example to the `Queues` page so it was clear the `queue` YAML attribute could be used as either root level or on a specific step.

Then I added a reference to branch filtering from the Queues page for completeness.

**QUERY:** I wasn't sure about referencing things as either 'build steps' or 'jobs'. It differs page to page. Welcome your input @mbelton-buildkite

### DIFF
--- a/pages/agent/v3/queues.md
+++ b/pages/agent/v3/queues.md
@@ -28,12 +28,17 @@ buildkite-agent start --tags "queue=building,queue=testing"
 
 ## Targeting a queue
 
-Target specific queues using the `agents` attribute on your pipeline steps.
+Target specific queues using the `agents` attribute on your pipeline steps or at the root level for the entire pipeline.
 
-For example, the following build step matches only agents running on the `deploy` queue (and ignores the agents running the `default` queue):
+For example, the following pipeline would run on the `priority` queue as determined by the root level `agents` attribute (and ignores the agents running the `default` queue). The `tests.sh` build step matches only agents running on the `deploy` queue.
 
 ```yaml
+agents:
+  queue: "priority"
+
 steps:
+  - command: echo "hello"
+
   - command: tests.sh
     agents:
       queue: "deploy"
--- a/pages/agent/v3/queues.md
+++ b/pages/agent/v3/queues.md
@@ -45,4 +45,4 @@ steps:
 
 ## Alternative methods
 
-Alternatively you can use [branch patterns](/docs/pipelines/branch-configuration) to control which pipelines and build steps are built based on the branch names.
+[Branch patterns](/docs/pipelines/branch-configuration) are another way to control what work is done. You can use branch patterns to determine which pipelines and steps run based on the branch name.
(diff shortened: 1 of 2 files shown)


mbelton-buildkite:
I'll need to sleep on this one. It's useful information that I think we should include, but it's not clear if this is the right way to go about it. 

I believe the proper place to draw this connection is on an explanation page that talks about controlling when and where jobs run on agents. But I don't believe that page exists.

Without that, I see three ways to incorporate it on these pages:

1. As a normal part of the intro content, like you have it. My concern with this is that it immediately suggests going to a different page before people get into the content of this page.
1. A callout in the intro content. This would help separate it from the regular content, but my concern is that it'll highlight it too much.
1. Add a new section at the end of the page, "Related concent/concepts." The drawback here is that you might miss it.

I'm leaning towards 3. What do you think?

mbelton-buildkite:
> I see three ways to incorporate it on these pages:
> 
> 1. As a normal part of the intro content, like you have it. My concern with this is that it immediately suggests going to a different page before people get into the content of this page.
> 2. A callout in the intro content. This would help separate it from the regular content, but my concern is that it'll highlight it too much.
> 3. Add a new section at the end of the page, "Related concent/concepts." The drawback here is that you might miss it.

I think 3 is the way to go, with a new section "## Alternative methods" at the end of the page. I won't have time to look at it today, but can help tomorrow.

clbarrell:
Thanks Will do 👍

clbarrell:
can you re -review @mbelton-buildkite

pages/agent/v3/queues.md
```diff
@@ -23,18 +22,27 @@ buildkite-agent start --tags "queue=building,queue=testing"
...
+  - command: echo "hello"
+
   - command: tests.sh
     agents:
       queue: "deploy"
 ```
+
+## Alternative methods
+
+Alternatively you can use [branch patterns](/docs/pipelines/branch-configuration) to control which pipelines and build steps are built based on the branch names.
```
mbelton-buildkite, line 48: I don't think this sounds parallel enough to quite fit. Perhaps:

```suggestion
[Branch patterns](/docs/pipelines/branch-configuration) are another way to control what work is done. You can use branch patterns to determine which pipelines and steps run based on the branch name.
```

pages/pipelines/branch_configuration.md
```diff
@@ -68,3 +68,7 @@ The following are examples of patterns, and the branches that they will match:
 * `v* !v1.*` will match any branch that begins with a `v` unless it also begins with `v1.`, such as `v2.3`, but not `v1.1`
 
 For more advanced step filtering, see the [Using conditionals](/docs/pipelines/conditionals) guide.
+
+## Alternative methods
+
+Alternatively you can use [Agent Queues](/docs/agent/v3/queues) to control which pipelines and build steps run on which agents.
```
mbelton-buildkite, line 74: As with the other page:

```suggestion
[Queues](/docs/agent/v3/queues) are another way to control what work is done. You can use queues to determine which pipelines and steps run on particular agents.
```

mbelton-buildkite (commented):
One more tweak and then I think it'll be good to go

mbelton-buildkite (approved):

PR #2393, opened 2023-08-21
Split up GraphQL cookbooks

dannymidnight:
The current GraphQL cookbook page is big. This PR breaks up the current `/graphql/graphql-cookbooks` route into seperate cookbook routes. 

Thoughts @mbelton-buildkite?


config/routes.rb
```diff
@@ -82,6 +82,7 @@
   get "/docs/test-analytics/js-collectors",       to: redirect("/docs/test-analytics/javascript-collectors")
   get "/docs/tutorials/elastic-ci-stack-aws",     to: redirect("/docs/agent/v3/elastic-ci-aws")
   get "/docs/agent/clusters",                     to: redirect("/docs/clusters/overview")
+  get "/docs/apis/graphql/graphql-cookbook",      to: redirect("/docs/apis/graphql/cookbooks/pipelines")
```
mbelton-buildkite, line 85: I think we should keep an overview page at `/docs/apis/graphql/graphql-cookbook`. This means you can still talk about "the cookbook" and have one link to share. It also avoids the need for a redirect.

data/nav_graphql.yml
```diff
@@ -3,8 +3,22 @@
   path: apis/graphql-api
 - name: Console and CLI tutorial
   path: apis/graphql/graphql-tutorial
-- name: Cookbook
-  path: apis/graphql/graphql-cookbook
+- name: Cookbooks
+  children:
+  - name: Pipelines
```
mbelton-buildkite, line 8: Based on my other comment, something like:

```suggestion
- name: Cookbook
  children:
  - name: Overview
    path: apis/graphql/graphql-cookbook
  - name: Pipelines
```

mbelton-buildkite (commented):
I haven't done full review but looks like what I expected. I've just left one thought about keeping an overview page so there's still one point of reference to link people to.

config/routes.rb
dannymidnight: Makes sense. So that page would essentially have links out to all of these sub-pages?

mbelton-buildkite: Yeah, it would mention what a cookbook is and link out to the child pages. Similar to: https://buildkite.com/docs/agent/v3/cli-reference

I can help with the content for that page

mbelton-buildkite (approved):

PR #2396, opened 2023-08-22
Smooth scroll the table of contents

dannymidnight:
This PR makes the table of contents links scroll smoothly to the content on the page. 

Whilst building this out, I ended up rewriting most of the logic for table of contents link highlighting to use the more performant [`IntersectionObserver` API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) and wrapped content in `section` tags.


dannymidnight:
No longer relevant.

PR #2399, opened 2023-08-22
Add example for queue and tag targeting rule

nsuma8989:
In this PR added an example to specify that when a job has targeting rules set with both tag and queue then an agent should be present which has both in order for the job to be scheduled.

### DIFF
--- a/pages/agent/v3/cli_start.md
+++ b/pages/agent/v3/cli_start.md
@@ -66,6 +66,16 @@ The `queue` tag works differently from other tags, and can be used for isolating
 
 If you specify a `queue` and [agent `tags`](#agent-targeting), your build will only run on agents that match **all** of the specified criteria.
 
+For example, if a job has agent targeting rules set as shown below then an agent should be present which has both queue=test and postgres=1.9.4 otherwise job will not schedule to an agent.
+
+```yaml
+steps:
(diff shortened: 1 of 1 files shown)


pages/agent/v3/cli_start.md
```diff
@@ -66,6 +66,16 @@ The `queue` tag works differently from other tags, and can be used for isolating
 
 If you specify a `queue` and [agent `tags`](#agent-targeting), your build will only run on agents that match **all** of the specified criteria.
 
+For example, if a job has agent targeting rules set as shown below then an agent should be present which has both queue=test and postgres=1.9.4 otherwise job will not schedule to an agent.
```
mbelton-buildkite, line 69: General style nitpicks:

```suggestion
For example, if a job has the following agent targeting rules, an agent with both `queue=test` and `postgres=1.9.4` should be present. Otherwise, the job will not dispatch to an agent.
```

mbelton-buildkite (commented):
One minor update, then looks good to me!

nsuma8989:
> One minor update, then looks good to me!

@mbelton-buildkite Made the suggested change, please take a look

mbelton-buildkite (approved):

PR #2400, opened 2023-08-22
Add note about conditional on build state should be only used at pipe…

nsuma8989:
Conditional expressions in notify for build state should only be used at pipeline level and not at step level.

### DIFF
--- a/pages/pipelines/conditionals.md
+++ b/pages/pipelines/conditionals.md
@@ -93,6 +93,9 @@ notify:
 ```
 {: codeblock-file="pipeline.yml"}
 
+>📘 Evaluating conditionals based on build state
+> Conditional expressions on build state will be available only at pipeline level and should not be used at step level.
+
 ## Conditionals and the broken state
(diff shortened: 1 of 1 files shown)


pages/pipelines/conditionals.md
```diff
@@ -93,6 +93,9 @@ notify:
 ```
 
+>📘 Evaluating conditionals based on build state
+> Conditional expressions on build state will be available only at pipeline level and should not be used at step level.
```
mbelton-buildkite, line 97: I think using the callout draws too much attention to this point. It can probably be included as regular text:

```suggestion
Note that conditional expressions on the build state are only available at the pipeline level. You can't use them at the step level.
```

What do you think?

nsuma8989:
@mbelton-buildkite Made the suggested change. Please let me know if we are good to merge this

mbelton-buildkite (approved):
Looks good to me, thanks!

PR #2404, opened 2023-08-23
Remove the agent upgrading page

mbelton-buildkite:
We've had multiple pieces of feedback that the current [Upgrading page](https://buildkite.com/docs/agent/v3/upgrading) looks incomplete and bad. It also recently showed up in a report showing that it's a bad page for our SEO since it lacks content.

This change:

- Moves the information into the Installation page. I think that's the logical place for it. I also updated the style.
- Removes the Upgrading page.
- Adds a redirect to the new location.
- Displays the table of contents on the Installation page now that there are two headings to show.


triarius (approved):
Nice, I think it makes total sense to have upgrading as a sub heading of installation.

PR #2406, opened 2023-08-23
Update agent cli reference docs generation

triarius:
This is the script `cli2md.rb` that generates the agent cli reference docs.
Its input is the output of `buildkite-agent` commands with the flag `--help`.

We've changed some of the practices since when this script was first written. Now:

- We will encase code blocks in `shell` backtick blocks. In the input, we will use 4 spaces to indicate the start of a code block.
- We will also replace prime signs with backticks in all the input, previously it was just the descriptions of the arguments.
- We also remove the user's home directory from the docs, since it is not relevant to the docs.
- We also make the state machine explicit, instead of using 2 boolean variables to control it.

There is a companion [PR](https://github.com/buildkite/agent/pull/2317) to adjust the cli commands in the agent to work consistently with this script. If you checkout that version of the agent, compile it and run `scripts/update-agent-help.sh`, it should be a no-op.

### DIFF
--- a/scripts/cli2md.rb
+++ b/scripts/cli2md.rb
@@ -46,6 +46,7 @@ ARGF.each_with_index do |line, line_num|
     # helptext for... reasons.
     # See: https://github.com/buildkite/agent/blob/main/clicommand/prime-signs.md
     desc.tr!("′", "`")
+    desc.gsub!(Dir.home, "$HOME")
     print "<tr id=\"#{command}\">"
     print "<th><code>--#{command} #{value}</code> <a class=\"Docs__attribute__link\" href=\"##{command}\">#</a></th>"
     print "<td><p>#{desc}"
--- a/scripts/cli2md.rb
+++ b/scripts/cli2md.rb
@@ -2,11 +2,11 @@
 
 require "cgi"
 
-# are we in a code block?
-in_code = false
+STATE_INITIAL = :inital
+STATE_CODE = :code
+STATE_TABLE = :table
 
-# are with in the table of parameters?
-in_table = false
+state = STATE_INITIAL
 
 # read from stdin or files in the args
 ARGF.each_with_index do |line, line_num|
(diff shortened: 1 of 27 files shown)


scripts/cli2md.rb
```diff
@@ -2,37 +2,51 @@
...
+in_code = false
+
+# are with in the table of parameters?
+in_table = false
 
 # read from stdin or files in the args
 ARGF.each_with_index do |line, line_num|
-  # Headings
+  # Replace all prime symbols with backticks. We use prime symbols instead of backticks in CLI
+  # helptext because Go does not support escaping backticks in backtick delimited strings.
```
moskyb, line 14: > because Go does not support escaping backticks in backtick delimited strings

this is true, but not why we avoid backticks in usage strings; we actually avoid them because urfave/cli has [special handling](https://cli.urfave.org/v1/examples/flags/#placeholder-values) for backtick-enclosed substrings in usage docs, and mangles usage strings that have backticked strings in them

moskyb: admittedly i probably could have put that there instead of just writing "reasons"

mbelton-buildkite (approved):

scripts/cli2md.rb
```diff
@@ -1,66 +1,94 @@
...
-            first_param = true
-        end
-        command_and_value = $1.rstrip
-        command = command_and_value.split[0][2..-1]
-        value = command_and_value.split[1]
-        desc    = $2
+  # Replace all prime symbols with backticks. We use prime symbols instead of backticks in CLI
+  # helptext because Go does not support escaping backticks in backtick delimited strings.
+  # See: https://github.com/buildkite/agent/blob/main/clicommand/prime-signs.md
+  line.tr!("′", "`")
```
triarius, line 16: Woah, my browser is not rendering this very well, but you be able to copy this into your text editor and see that is prime sign.

moskyb (changes requested):

moskyb (approved):

PR #2407, opened 2023-08-23
Update clusters diagram to new visual language

bravonovember:
Update Clusters diagram to new visual language. This diagram introduces the idea of clusters A, B, C to the page.


mbelton-buildkite (approved):
I'd originally planned to talk more about the clusters in the diagram but think this simple text change does enough for now.

PR #2408, opened 2023-08-23
Update exit status troubleshooting note

BrettJay:
The way the note is currently written is a bit confusing:

CleanShot 2023-08-23 at 16 46 12@2x

We could rewrite this callout while maintaining a link to the REST API, but to be as clear as we can (and assuming my interpretation is accurate), I think we should either say exit status is “only available when using the GraphQL API”, or “is not available when using the REST API”, rather than trying to word this note in such a way that it refers to _both_ APIs, while trying to indicate it’s only available in one.

A smaller rewrite could read:

```
Exit status information is available in the <a href="/docs/apis/graphql-api">GraphQL API</a>, but not the <a href="/docs/apis/rest-api">REST API</a>
```

### DIFF
--- a/pages/pipelines.md
+++ b/pages/pipelines.md
@@ -94,4 +94,4 @@ Job exit status may include the exit signal reason, which indicates whether the 
 <%= image "exit-status.png", width: 2048/2, height: 880/2, alt: "Exit status of a job" %>
 
 >🚧
-> Exit status information available in the <a href="/docs/apis/graphql-api">GraphQL API</a> but not the <a href="/docs/apis/rest-api">REST API</a>.
+> Exit status information is only available using the <a href="/docs/apis/graphql-api">GraphQL API</a>.
--- a/pages/pipelines.md
+++ b/pages/pipelines.md
@@ -93,5 +93,4 @@ Job exit status may include the exit signal reason, which indicates whether the 
 
 <%= image "exit-status.png", width: 2048/2, height: 880/2, alt: "Exit status of a job" %>
 
->🚧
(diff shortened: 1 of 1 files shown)


pages/pipelines.md
```diff
@@ -94,4 +94,4 @@ Job exit status may include the exit signal reason, which indicates whether the
 <%= image "exit-status.png", width: 2048/2, height: 880/2, alt: "Exit status of a job" %>
 
 >🚧
-> Exit status information available in the <a href="/docs/apis/graphql-api">GraphQL API</a> but not the <a href="/docs/apis/rest-api">REST API</a>.
+> Exit status information is only available using the <a href="/docs/apis/graphql-api">GraphQL API</a>.
```
BrettJay, line 97: Possible alternatives

```suggestion
> Exit status information is only available in the <a href="/docs/apis/graphql-api">GraphQL API</a>, not the <a href="/docs/apis/rest-api">REST API</a>.
```

```suggestion
> Exit status information is only available in the <a href="/docs/apis/graphql-api">GraphQL API</a>.
```

```suggestion
> Exit status information is not available in the <a href="/docs/apis/rest-api">REST API</a>.
```

mbelton-buildkite: I don't think this warrants the callout treatment. Callouts have been used fairly liberally as quick solutions but using too many undermines the ones that are actually important. How about we just make this regular text, like:

```
If you want to access the exit status through an API, it's only available in the [GraphQL API](/docs/apis/graphql-api).
```

What do you think? (I couldn't add a suggestion across the lines 😔)

BrettJay: Agreed!

mbelton-buildkite (approved):
LGTM!

PR #2409, opened 2023-08-23
Fix pluralisation on webhooks callout

BrettJay:
Small pluralisation fix for the Webhooks callout note.


mbelton-buildkite (approved):
Thanks for catching and fixing this!

PR #2410, opened 2023-08-23
Fix Algolia crawler 

dannymidnight:
It appears our self-run Algolia crawler is over indexing a bunch of records because we're been including each item of the nav and table of contents, resulting in ~500 additional records per-page.

This PR updates out algolia config to exclude irrelevant content and fixes up the selectors to match the latest page hierarchy.

For reference:

- https://docsearch.algolia.com/docs/legacy/config-file
- https://docsearch.algolia.com/docs/legacy/how-do-we-build-an-index

### DIFF
--- a/config/algolia.json
+++ b/config/algolia.json
@@ -11,11 +11,7 @@
     "lvl5": ".Page article h6",
     "text": ".Page article p, .Page article li, .Page article td"
   },
-  "selectors_exclude": [
-    "[data-algolia-exclude]",
-    ".Docs__toc",
-    ".HomepageCategory"
-  ],
+  "selectors_exclude": ["[data-algolia-exclude]", ".Toc", ".Footer", ".Nav"],
   "custom_settings": {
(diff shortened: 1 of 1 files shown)


config/algolia.json
```diff
@@ -1,39 +1,18 @@
...
-    "synonyms": [
-      [
-        "Pipeline.yml",
-        "Pipeline.yaml",
-        "config.yml",
-        "config.yaml"
-      ]
-
-    ]
+    "synonyms": [["Pipeline.yml", "Pipeline.yaml", "config.yml", "config.yaml"]]
```
mbelton-buildkite, line 16: Huh, that's cool. Will keep an eye out for any other obvious terms to include here

mbelton-buildkite (approved):

mbelton-buildkite:
> ~500 additional records per-page

🤯

config/algolia.json
dannymidnight: I believe these are also configurable directly inside Algolia.

PR #2416, opened 2023-08-24
Add dockerized version of the docs

dannymidnight:
I keep running into issues with conflicting gems between buildkite and our docs.

This PR dockerizes vite so now everything can be run with the following:

```sh
docker-compose up -d app
```


mbelton-buildkite:
What does this mean for running locally? Out with `foreman start`, in with `docker-compose up -d app`?

PR #2433, opened 2023-08-25
Update Docs wordmark and open graph image

BrettJay:
Updates the Buildkite Docs wordmark, and adds a new open graph image for social sharing.

CleanShot 2023-08-25 at 12 34 25@2x


images/docs/opengraph_default.png
olyism, line 1: I do wonder if we could pick a different page that's not the homepage 🤔 Like a page with code snippets or a more zoomed in view.

That being said, this is a big improvement on what we have currently ✨ , so happy to approve this and revisit later 👍

olyism (approved):

mbelton-buildkite (approved):
Love! :shipit:

images/docs/opengraph_default.png
BrettJay: That’s a really good point! Definitely will have a think about that for a future uplift :)

PR #2438, opened 2023-08-27
Add ids to h1 tags for search

dannymidnight:
Sometimes I'm seeing search results with some blank `#main` links. 

I noted that we didn't include IDs on our h1's which I think means the default h1 indexing doesn't have anything to link to. 

Maybe this will improve things 🤷


mbelton-buildkite (approved):
Sure 🤷🏻‍♂️

PR #2446, opened 2023-09-01
Fix center alignment of footer paragraph

BrettJay:
Max width is applied so in certain contexts, the footer paragraph text looks misaligned:

**Before:**
CleanShot 2023-09-01 at 08 00 01@2x

**After:**
CleanShot 2023-09-01 at 08 00 17@2x

### DIFF
--- a/app/assets/stylesheets/components/_footer.scss
+++ b/app/assets/stylesheets/components/_footer.scss
@@ -4,6 +4,9 @@
   margin-bottom: 30px;
   margin-top: 60px;
   text-align: center;
+  p {
+    margin-inline: auto;
+  }
 }
 
 .Footer__heading {
--- a/app/assets/stylesheets/components/_footer.scss
+++ b/app/assets/stylesheets/components/_footer.scss
@@ -4,9 +4,6 @@
   margin-bottom: 30px;
   margin-top: 60px;
   text-align: center;
-  p {
-    margin-inline: auto;
-  }
 }
 
 .Footer__heading {
--- a/app/assets/stylesheets/components/_page.scss
+++ b/app/assets/stylesheets/components/_page.scss
@@ -87,16 +87,5 @@ $lg-screen-margin-top: map-get($header-height, "one-row");
     &:nth-child(2) {
       max-width: 1074px;
     }
-
-    h1,
-    h2,
-    h3,
-    h4,
-    h5,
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/components/_footer.scss
```diff
@@ -4,6 +4,9 @@
   margin-bottom: 30px;
   margin-top: 60px;
   text-align: center;
+  p {
```
olyism, line 7: I think we should address `max-width` holistically, similar to how [Tailwind's `prose`](https://tailwindcss.com/docs/typography-plugin) handles maximum reading length.

We should address max-width handling on the container, rather than the lines of text themselves. See `_page.scss`:

```scss
.Page__content {
  grid-area: content;
  width: 100%;
  max-width: map-get($max-w, "prose");
  margin-inline: auto;
  padding-block: 24px;

  @media (min-width: $screen-lg) {
    justify-self: center;
    padding-block: 40px;
    &:nth-child(2) {
      max-width: 1074px;
    }

    h1,
    h2,
    h3,
    h4,
    h5,
    h6,
    p,
    ul {
      max-width: map-get($max-w, "prose");
    }
  }
}
```

So, rather than assigning max-width on h1...6, p and ul, we should set it on a container so it can cover other content elements, such as ol, table, dl/dt/dd et al.

I suggest we can use `.Article` as the `prose` equivalent.

Also, some nitpicking:
- We should avoid element selectors on a component level
- Need a line between CSS declarations (should really add a scss lint rule 🤔 )

app/assets/stylesheets/components/_footer.scss
```diff
@@ -4,6 +4,9 @@
...
+    margin-inline: auto;
```
olyism, line 8: TIL `margin-inline` 💡

olyism (changes requested):

app/assets/stylesheets/components/_footer.scss
BrettJay: Turns out `max-width: map-get($max-w, "prose");` is already in place on the containing `.Page__content` element. Looks like this issue can be resolved simply by removing the rules under the media query as suggested.

olyism (approved):
Thanks for the fix. LGTM 🚀

PR #2447, opened 2023-09-01
🐛 GraphQL API schema pages not rendering correct attribute type

olyism:
The GraphQL API schema Docs have not been generating the attribute types correctly:
- It's not showing array types
- Not null-ish is showing as null-ish, and vice versa

# Screenshots

Take `OBJECT` / `JobTypeCommand` as an example, where:

```graphql
"""
A type of job that runs a command on an agent
"""
type JobTypeCommand implements JobInterface & Node {
...
  """
  Environment variables for this job
  """
  env: [String!]

  """
  Job events
  """
  events(
    """
    Returns the elements in the list that come after the specified cursor.
    """
    after: String

    """
    Returns the elements in the list that come before the specified cursor.
    """
    before: String

    """
    Returns the first _n_ elements from the list.
    """
    first: Int

    """
    Returns the last _n_ elements from the list.
    """
    last: Int
  ): JobEventConnection!
...
```

## Before

It's currently incorrectly rendered:

Screenshot 2023-09-01 at 2 56 03 PM

## After

After applying the fix:

Screenshot 2023-09-01 at 2 56 56 PM


scripts/graphql_api_content/render_helpers.rb
```diff
@@ -10,21 +10,24 @@ def render_html(markdown)
     markdown ? CommonMarker.render_html(markdown) : nil
   end
 
-  def render_of_type(type, nullable = true, size = "medium")
+  def render_of_type(type, is_list = false, is_non_nullable = false, size = "medium")
```
olyism, line 13: Fixed by correctly discerning type attributes, i.e. non-nullable & list/array, by correctly reading the nested JSON version of the schema. For example:

```json
{
              "name": "agentQueryRules",
              "description": null,
              "type": {
                "kind": "LIST",
                "name": null,
                "ofType": {
                  "kind": "NON_NULL",
                  "name": null,
                  "ofType": {
                    "kind": "SCALAR",
                    "name": "String",
                    "ofType": null
                  }
                }
              },
              "defaultValue": null,
              "isDeprecated": false,
              "deprecationReason": null
            },
```

The `render_of_type` method should render `agentQueryRules` as `[String]!` because it's nested within "LIST" and "NON_NULL" in the as specified in the JSON.

olyism:
I originally built the schema pages as a prototype. Would it be worth doing this properly, like fetching the schema file from the source and render the nav and page, rather then running a script? 🤔 

@dannymidnight @mbelton-buildkite

mbelton-buildkite (approved):

dannymidnight:
> I originally built the schema pages as a prototype. Would it be worth doing this properly, like refactor using ERBs, rather then HEREDOCs in the script? 🤔
> 
> @dannymidnight

I think we can think about fighting that fight another time if we're still in rails land.

PR #2452, opened 2023-09-04
🔍 Canonical URL

olyism:
Specify canonical URL on every page as `https://buildkite.com/docs/*/**` (without www).

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -48,4 +48,13 @@ RSpec.describe ApplicationHelper do
       end
     end
   end
+
+  describe "#seo_canonical_url" do
+    it "returns the correct canonical URL" do
+      # Mocking the request object's path method
+      allow(helper).to receive_message_chain(:request, :path).and_return("/docs/agent/v3")
+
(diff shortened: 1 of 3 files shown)


olyism:
Canonical URL isn't rendered on first page load, possibly due to caching? @dannymidnight I don't how caching has been set up and will need some help here, thanks 🙏

spec/helpers/application_helper_spec.rb
```diff
@@ -48,4 +48,13 @@
       end
     end
   end
+
+  describe "#seo_canonical_url" do
```
dannymidnight, line 52: Rather than I helper spec, it's probably more valuable writing a feature spec to assert the tag is rendered in the markup.

dannymidnight (approved):
Just saw the google indexing issue. I think this is good enough to ship as is.

spec/helpers/application_helper_spec.rb
olyism: cool thanks, since we're having the google indexing issue I'll merge this now and revisit the spec in a separate PR.

PR #2453, opened 2023-09-04
✨ Specify content keyword in front matter

olyism:
Define page focus keywords with the `keywords` front matter meta data.

Render keywords in `data-content-keywords` in the `body` tag to highlight the focus keywords of each page for content authors. This helps the content team to quickly inspect to see the types of content we're providing across different channels.

Use graceful fallback when no keywords have been provided — comma separated URL path segments.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -35,12 +35,4 @@ class PagesController < ApplicationController
   end
   helper_method :beta?
 
-  # Renders keywords to the page for content writers to inspect and use as a guide
-  # Gracefully falls back to the page's path if no keywords are specified
-  # to help reduce content workload
-  def content_keywords
-    @page.keywords || request.path.split("/").reject(&:empty?).map { |segment| segment.gsub("-", " ") }.join(", ")
-  end
-  helper_method :content_keywords
-
 end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -35,4 +35,12 @@ class PagesController < ApplicationController
   end
   helper_method :beta?
 
+  # Renders keywords to the page for content writers to inspect and use as a guide
+  # Gracefully falls back to the page's path if no keywords are specified
+  # to help reduce content workload
+  def content_keywords
+    @page.keywords || request.path.split("/").reject(&:empty?).map { |segment| segment.gsub("-", " ") }.join(", ")
+  end
+  helper_method :content_keywords
+
 end
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -177,6 +177,12 @@ class Page
     front_matter.fetch(:template, "show")
   end
 
+  # Returns focus keywords to guide content writers with an overview notion of the page
+  # It's not for meta keywords, which is a deprecated practice
+  def keywords
+    front_matter.fetch(:keywords)
+  end
+
   private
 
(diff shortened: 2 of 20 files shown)


app/models/page.rb
```diff
@@ -177,6 +177,12 @@ def template
     front_matter.fetch(:template, "show")
   end
 
+  # Returns focus keywords to guide content writers with an overview notion of the page
+  # It's not for meta keywords, which is a deprecated practice
+  def keywords
```
dannymidnight, line 182: Good comment 👍

app/controllers/pages_controller.rb
```diff
@@ -35,4 +35,12 @@ def beta?
   end
   helper_method :beta?
 
+  # Renders keywords to the page for content writers to inspect and use as a guide
+  # Gracefully falls back to the page's path if no keywords are specified
+  # to help reduce content workload
+  def content_keywords
+    @page.keywords || request.path.split("/").reject(&:empty?).map { |segment| segment.gsub("-", " ") }.join(", ")
```
dannymidnight, line 42: Feels like this logic should live inside the page model so that we don't need the additional view helper and the view can call `@page.keywords`.

dannymidnight (commented):
Some minor feedback but overall looks good.

app/controllers/pages_controller.rb
olyism: `request` isn't the concern of the page model so I've added the fallback logic as a helper method in the controller.

dannymidnight: I believe you should have access to the request object in the page model via `view_context`.

https://github.com/buildkite/docs/blob/a82878cda347d6de31ed6c357d53e11d81733845/app/controllers/pages_controller.rb#L12

Which means you can do something like the following in `page.rb`:

```ruby
@view.request.path
```

olyism: hi @dannymidnight thanks for the tip, I've made the change:

```ruby
def keywords
  # Gracefully falls back to the page's path if no keywords are specified to help reduce technical writer workload
  front_matter.fetch(:keywords) || keywords_from_path
end
```

Also made a few minor refactoring. Please see latest commits. Cheers :)

MelissaKaulfuss (approved):
LGTM ✨

MelissaKaulfuss:
@dannymidnight can you please deploy for me!

app/models/page.rb
```diff
@@ -177,6 +177,13 @@ def template
     front_matter.fetch(:template, "show")
   end
 
+  # Returns focus keywords to guide content writers with an overview of the page content
+  # Note: it's not for meta keywords, which is a deprecated SEO practice
+  def keywords
+    # Gracefully falls back to the page's path if no keywords are specified to help reduce technical writer workload
+    front_matter.fetch(:keywords) || keywords_from_path
```
dannymidnight, line 184: The second arg of `fetch` is the fallback value so you can write this is the following:

```suggestion
    front_matter.fetch(:keywords, keywords_from_path)
```

dannymidnight (approved):
Nice one. Looks better, only very minor feedback (non-blocking).

app/models/page.rb
olyism: ah TIL, I've committed the change 👍 

Will merge after checks pass.

PR #2464, opened 2023-09-07
Add Mel and Michelle as codeowners

mbelton-buildkite:
This adds @MelissaKaulfuss and @meiqimichelle as default reviewers for the content while I'm off. @dannymidnight, is this the right syntax? [I think it is](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax).


dannymidnight (approved):
👍

PR #2474, opened 2023-09-14
Adds a query to find UUIDs of jobs in a build

karensawrey:
Turned out, no one I asked knew how to do that from the top of their heads, so let's have it in the cookbook now that we know how to do it.


MelissaKaulfuss (approved):
LGTM! 🚢

PR #2477, opened 2023-09-15
A small addition to the meta-data page

karensawrey:
A clarification that emerged in a support ticket discussion. Having this explicitly in the docs might help someone down the line.

### DIFF
--- a/pages/pipelines/build_meta_data.md
+++ b/pages/pipelines/build_meta_data.md
@@ -4,6 +4,8 @@ In this guide, we'll walk through using the Buildkite agent's [meta-data command
 
 Meta-data is intended to store data 1 kilobyte or less in size, to be used across steps. For example, you can tag a build with the software version it deploys so that you can later identify which build deployed a particular version. For anything over 1 kb use an [artifact](/docs/pipelines/artifacts) instead.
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/build_meta_data.md
```diff
@@ -4,6 +4,8 @@ In this guide, we'll walk through using the Buildkite agent's [meta-data command
 
 Meta-data is intended to store data 1 kilobyte or less in size, to be used across steps. For example, you can tag a build with the software version it deploys so that you can later identify which build deployed a particular version. For anything over 1 kb use an [artifact](/docs/pipelines/artifacts) instead.
 
+Note that meta-data exists at the build level, so setting it in a job sets it for the build (and setting the same key in multiple jobs will result in the value from the last one to set it).
```
MelissaKaulfuss, line 7: Can I ask what this means:
> "Setting the same key in multiple jobs will result in the value from the last one to set it."

Does this mean, if there's a key in the meta-data at the build level, and the value is being set inside multiple jobs, that the final key/value pair that's stored in the metadata at the build level will be the one that was last set?

MelissaKaulfuss (commented):
Feel like line 7 can be a little clearer.

PR #2481, opened 2023-09-18
Fix typo

jcoelho93:
ran instead of run


MelissaKaulfuss:
Thanks for taking the time to make this suggestion! 😊

"Ran" can certainly be used and could also be correct, especially if we say "have installed", and "have ran". However we can shorten the phrases by dropping the have before each one. And in this specific case "run" actually works.

Thanks again, we appreciate your effort. 💚

PR #2486, opened 2023-09-19
System Banners Documentation

jameshill:
System banners is an enterprise only feature for organizations to communicate with their staff via announcement banners.
System banners is part of the Q3 release.

:basecamp: [Internal Project Docs](https://3.basecamp.com/3453178/buckets/23328660/messages/6528240759)


MelissaKaulfuss (approved):
LGTM! ✨

PR #2491, opened 2023-09-21
[GraphQL Cookbook] Added audit event query for specific users 

james2791:
Added an extra snippet to the Organization context of the GraphQL cookbook specifically around getting specific user events using an `auditEvents` query

Said query has an `actor` parameter that you can pass in, but it is the `member` ID (aka user's GraphQL ID) which needs to be obtained first to feed the parent query 🙂


MelissaKaulfuss (approved):
LGTM 👍🏻

MelissaKaulfuss (commented):
LGTM 🌟

PR #2492, opened 2023-09-21
Add docs for waterfall V2

l-suzuki:
Implements waterfall view docs (to be released  as part of Q3 event). 

**Changes:**
- introduces a new section in the pipelines docs nav called "Insights" - see [linear ticket](https://linear.app/buildkite/issue/EDGE-1294/write-up-docs-for-waterfall), which includes nested items Waterfall view and Cluster Queue metrics
- Adds docs for Waterfall view. 

DEMO: 
https://github.com/buildkite/docs/assets/10970711/c150341d-d1c9-4c0e-a6cd-07511470daf4

PR #2498, opened 2023-09-27
Bugfix: Fix cluster metrics routing

l-suzuki:
Fixes the bug picked up by @kalopilato in https://github.com/buildkite/docs/pull/2492#issuecomment-1738191712 (this was noticed just after the PR was merged)


kalopilato (approved):
I've given this a test in PR preview (so good!) and works as expected. It's a little disorientating jumping from "Clusters / Queue metrics" to "Insights / Cluster queue metrics" in the sidebar nav but I don't see a better way while we have two links to the same page (without duplicating and maintaining two copies of the same page that is)

This is at least an improvement on prior behaviour :shipit:

PR #2502, opened 2023-09-28
Update tutorial with v2.0 release of TPA

lizrabuya:


MelissaKaulfuss (approved):
Looks good 👍🏻

Have you decided to link to the `action.yml` instead of include to keep it in sync in case it changes?

PR #2503, opened 2023-09-28
Troubleshooting concurrency groups

pzeballos:
When a build is stuck because of a concurrency group, it is hard to see which jobs are in the queue.  We have a GraphQL query that is very useful in these situations, and it lists all jobs in the queue of a concurrency group and its state.

We eventually have something in the UI (we have it already, but the performance is not very good)

I wasn't sure where to add this note, so I put it as an info section


pages/pipelines/controlling_concurrency.md
```diff
@@ -43,6 +43,9 @@ For example, if you have two pipelines that each deploy to a different target bu
 
 Concurrency groups guarantee that jobs will be run in the order that they were created in. Jobs inherit the creation time of their parent. Parents of jobs can be either a build or a pipeline upload job. As pipeline uploads add more jobs to the build after it has started, the jobs that they add will inherit the creation time of the pipeline upload rather than the build.
 
+>📘 Troubleshooting
+> When a build is blocked by a concurrency group, you can run this [GraphQL query](/docs/apis/graphql/cookbooks/jobs#get-all-jobs-in-a-particular-concurrency-group) that will list all the jobs that are in the queue for that particular concurrency group.
```
mbelton-buildkite, line 47: I think this is a bit clearer:

```suggestion
> When a build is blocked by a concurrency group, you can check which jobs are in the queue and their state using this [GraphQL query](/docs/apis/graphql/cookbooks/jobs#get-all-jobs-in-a-particular-concurrency-group).
```

mbelton-buildkite (approved):
One small suggestion, but otherwise, looks good! 😊

pages/pipelines/controlling_concurrency.md
gilesgas: I've already committed this suggestion @pzeballos ; I'm assuming this wording revision is OK from your perspective?

gilesgas (changes requested):
Hi @pzeballos , just want you to confirm if this word modification (italicised) is correct...

> Be aware that both the `block` and `input` steps cause _these_ steps to be uploaded and scheduled at the same time, which breaks concurrency groups.

By "_these_", I'm referring to either the `block` or `input` steps themselves.

Is this correct?

pzeballos:
> Hi @pzeballos , just want you to confirm if this word modification (italicised) is correct...
> 
> > Be aware that both the `block` and `input` steps cause _these_ steps to be uploaded and scheduled at the same time, which breaks concurrency groups.
> 
> By "_these_", I'm referring to either the `block` or `input` steps themselves.
> 
> Is this correct?

That is correct @gilesgas !

gilesgas (approved):

PR #2479, opened 2023-09-18
[EDGE-1520] Add pipeline templates rest & graphql api documentation

jonathanly:
Adds pipeline templates REST API and GraphQL documentation for the following PRs:
- https://github.com/buildkite/buildkite/pull/13771
- https://github.com/buildkite/buildkite/pull/13886

~Note: graphql doc updates include bits around organization system banners, it's a side effect of generating the documentation via a schema file. I imagine this will resolve in the diff when https://github.com/buildkite/docs/pull/2486 is merged.~ **Fixed conflicts via a rebase ✨**

PR #2509, opened 2023-09-29
Update README for generating GraphQL docs

dannymidnight:
Based on https://github.com/buildkite/docs/pull/2479, I noted we're missing up-to-date documentation on the process for generating GraphQL API docs.

Hopefully this is a little easier to follow @jonathanly.


jonathanly:
That's super interesting, didn't know you could fetch the graphql schema that way either. Thanks for updating these docs @dannymidnight ❤️!

PR #2517, opened 2023-10-08
Upgrade from debian 11 (bullseye) to debian 12 (bookworm)

yob:
[Debian 12 was released in June 2023](https://wiki.debian.org/DebianBookworm) and has a range of package updates. The vast majority will not be significant in the context of the docs app and should have no impact on the app behaviour.

Updating is useful though, mainly just so we don't fall behind. It will also avoid some CVEs being detected by scanning software.

The critical CVE doesn't impact us, but here's the details: [CVE-2023-38408](https://security-tracker.debian.org/tracker/CVE-2023-38408).

The CVE is also fixed in bullseye so this upgrade isn't technically required, we could also install security packages . I reckon the OS bump is worth or for other reasons though. 

/cc @buildkite/platform


ellsclytn (approved):

dannymidnight:
Thanks @yob :)

PR #2518, opened 2023-10-09
Fetch security upgrades from apt [PLT-1607]

yob:
In #2517 I upgrade docs frm debian 11 to 12, but I was surprised to see that CVE-2023-38408 is sitll being detected post-upgrade.

I think it's because the patched openssh is available in the debian security repositories. If so, adding this upgrade should fetch them.

/cc @buildkite/platform

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,8 @@ RUN echo "--- :package: Installing system deps" \
     # Install all the things
     && apt-get update \
     && apt-get install -y nodejs gh jq \
+    ## Pull down security updates
+    && apt-get upgrade -y \
     # Upgrade rubygems and bundler
     && gem update --system \
     && gem install bundler \


yob:
Yup, this updated openssh (and others)

```
Calculating upgrade...
The following packages will be upgraded:
  base-files curl debian-archive-keyring debianutils krb5-multidev libc-bin
  libc-dev-bin libc6 libc6-dev libcurl3-gnutls libcurl4 libcurl4-openssl-dev
  libgssapi-krb5-2 libgssrpc4 libk5crypto3 libkadm5clnt-mit12
  libkadm5srv-mit12 libkdb5-10 libkrb5-3 libkrb5-dev libkrb5support0
  libmariadb-dev libmariadb-dev-compat libmariadb3 libpam-modules
  libpam-modules-bin libpam-runtime libpam0g libssl-dev libssl3 libsystemd0
  libudev1 libx11-6 libx11-data libx11-dev linux-libc-dev mariadb-common
  openssh-client openssl
39 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Need to get 20.3 MB of archives.
After this operation, 103 kB of additional disk space will be used.
```

PR #2519, opened 2023-10-09
Add ecr scan step

danstn:
This will introduce a soft-fail step that will add ECR scan results to build annotation.

### DIFF
--- a/.buildkite/pipeline.deploy.yml
+++ b/.buildkite/pipeline.deploy.yml
@@ -12,6 +12,13 @@ steps:
         login: true
         account-ids: ${ECR_ACCOUNT_ID}
 
+  - name: ":ecr: ECR Vulnerabilities Scan"
+    command: "true"
+    soft_fail: true
+    plugins:
+      - buildkite/ecr-scan-results#v1.2.0:
+          image-name: "${ECR_REPO}:${BUILDKITE_BUILD_NUMBER}"
+
   # If the current user is part of the deploy team, then wait for everything to
   # finish before deploying
   - wait: ~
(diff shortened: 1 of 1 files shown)


dannymidnight:
Nice one. Thanks @danstn

dannymidnight (approved):

.buildkite/pipeline.deploy.yml
```diff
@@ -12,6 +12,13 @@ steps:
         login: true
         account-ids: ${ECR_ACCOUNT_ID}
 
+  - name: ":ecr: ECR Vulnerabilities Scan"
+    command: "true"
+    soft_fail: true
+    plugins:
```
yob, line 18: does this need a `depends_on` to ensure it runs after the image is pushed?

.buildkite/pipeline.deploy.yml
```diff
@@ -12,6 +12,13 @@ steps:
...
+      - buildkite/ecr-scan-results#v1.2.0:
+          image-name: "${ECR_REPO}:${BUILDKITE_BUILD_NUMBER}"
+
   # If the current user is part of the deploy team, then wait for everything to
   # finish before deploying
   - wait: ~
```
yob, line 24: does this wait need to depend on the ECR scan? Maybe not now I guess, while the ECR scan is soft fail?

.buildkite/pipeline.deploy.yml
danstn: Yes. Updated here https://github.com/buildkite/docs/pull/2519/commits/6ce10328bfc0350cd5cf49d91687ecf85aedeafd.

PR #2521, opened 2023-10-10
Switch to a slim debian Docker image [PLT-1610]

yob:
The slim image comes with far fewer packages pre-installed and we need to manually installed a handful that we need (like curl and build-essential), but in return we get:

* A slightly smaller final image
* Fewer debian packages installed, which means fewer (low value, noisy) CVEs will be detected

The resulting image size difference looks like this:

```
docs                      slim-bookworm                           b7b6f5d9d33a   3 days ago      1.35GB
docs                      bookwork                                8a6a4a888efe   3 days ago      1.84GB
```

I think we could probably get it a lot smaller via multistage build, by ensuring tools like nodejs, gh, svn, mercurial, make, gcc, and cpp are used to build assets and native extensions, but not included in the final image. Still, this is a good start.

/cc @danstn

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/docker/library/ruby:3.1.4-bookworm@sha256:ce07ca486ea1589cdbc3df60a554c595b10a792e9a61a0363977d42e9de20726
+FROM public.ecr.aws/docker/library/ruby:3.1.4-slim-bookworm@sha256:6c5ebf029d658391fb9243b37e970f84100eab678e2e6e77f316f847e02d5ed9
 
 ARG RAILS_ENV
 ARG DD_RUM_VERSION="unknown"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,21 +10,23 @@ ENV DD_RUM_ENABLED=true
...
+    && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+    # Setup apt for node
+    && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
+    && curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
     # Install all the things
     && apt-get update \
-    && apt-get install -y nodejs gh jq \
+    && apt-get install -y nodejs gh jq build-essential \
```
yob, line 29: build-essential is the main thing that was missing from slim. It's needed to compile and install various ruby/node native packages.

Dockerfile
```diff
@@ -10,21 +10,23 @@ ENV DD_RUM_ENABLED=true
...
-RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
-RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
-
 RUN echo "--- :package: Installing system deps" \
     # Cache apt
     rm -f /etc/apt/apt.conf.d/docker-clean \
     && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache \
+    # Install a few pre-reqs
+    && apt-get update \
+    && apt-get install -y curl gnupg \
```
yob, line 19: curl and gnupg aren't needed in prod, but we do need them in the Dockerfile steps further down

dannymidnight:
Nice. On a related note I'm still very much thinking about converting the docs into a static site with the intent to eventually remove this image from production. I imagine this image will still kick around for a while and continue being used in CI but not necessarily for serving traffic.

dannymidnight (approved):

Issue #2512, opened 2023-10-04
Artifacts doc page should mention `--include-retried-jobs` flag


dewil-official:
Hello! I've been building an automatic retry script for Cypress in our CI, but came across a very odd behavior. Which is that Buildkite successfully downloads all artifacts _**except**_ from the same job, _**if it got retried**_. But that's exactly what I tried to do in my case.

So after many days of debugging and headache, I came across [this ticket](https://github.com/buildkite/agent/issues/1708), which mentions the `--include-retried-jobs` flag and it implies that by default, retried artifacts are discarded. That's unexpected behavior to me and should've been mentioned on the ["Build artifacts" doc page](https://buildkite.com/docs/pipelines/artifacts). It would've saved myself from many hours of debugging.

mbelton-buildkite:
Hi @dewil-official! I'm sorry you had to go through that debugging, but thank you for sharing your findings here. As you suggested, I've raised a PR to include a note about this on the Artifacts page. Let me know if there's anything else you think it should include. 😊

mbelton-buildkite:
I've merged the PR addressing this suggestion in the docs

PR #2522, opened 2023-10-11
Add help for when artifacts are missing from retried jobs

mbelton-buildkite:
A customer raised #2512 to describe some frustrating debugging work required on their part to discover they needed to use  `--include-retried-jobs` to get artifacts from a retried job. This change adds a note to the troubleshooting section on the Artifacts page to help people discover that.

@pda, I saw you mentioned on a thread related to that issue so hoped you might verify this information is correct. Let me know if there's someone else I should get as a reviewer. 😊

### DIFF
--- a/pages/pipelines/artifacts.md
+++ b/pages/pipelines/artifacts.md
@@ -154,7 +154,11 @@ If you manage your own artifact storage, then you are responsible for encryption
 
 To track the actions of users with access to your artifacts, use the [API Access Audit](https://buildkite.com/organizations/~/api-access-audit).
 
-## Troubleshooting a `Multiple artifacts were found for query` error
+## Troubleshooting artifacts
+
+The following suggestions resolve common issues with using artifacts.
+
+### Multiple artifacts were found for query
 
 The `buildkite-agent artifact download` command can fail with the following error message:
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/artifacts.md
```diff
@@ -172,3 +176,7 @@ For an example, read [download an artifact from a specific step](#download-artif
 
 Alternatively, download the most recent matching file by using a glob pattern.
 For an example, read [download many artifacts](#download-artifacts-with-the-buildkite-agent-example-download-many-artifacts).
+
+### Artifacts are missing from retried jobs
+
+Artifacts from retried jobs are discarded by default, so the `buildkite-agent artifact download` command won't find them. To include artifacts from retried jobs in your search results, use `--include-retried-jobs` in the command.
```
pda, line 182:
```suggestion
Artifacts from retried jobs are excluded by default, so the `buildkite-agent artifact download` command won't find them. To include artifacts from retried jobs in your search results, use `--include-retried-jobs` in the command.
```

Perhaps “discarded” could be read as “we delete artifacts when a job is retried”, and “excluded” is a nice inverse of the “include” in the CLI flag.

pda (approved):
Nice 👍

PR #2523, opened 2023-10-12
Ignore many CVEs

ellsclytn:
These CVEs have been assessed individually and all deemed safe to ignore. Best to view each commit for the justification behind each.


yob (commented):
It wouldn't help reduce the image size, but is there value in trying to add a step in the Dockerfile that removes the linux packages from the final image? IIRC they're used for building native extensions and not used at runtime, and it would drop 7 CVEs from the list we need to ignore

ellsclytn:
There probably is value. I didn't bother right now because I'd already triaged all but one of the CVEs by the time you mentioned that idea, so I figured I'll get this PR out, and then can address that idea separately.

danstn (approved):

dannymidnight:
Thanks @ellsclytn

 :shipit: !

PR #2524, opened 2023-10-12
Signed pipelines guide docs

moskyb:
We're getting closish to releasing signed pipelines on the agent - it's a fairly complicated feature and needs comprehensive docs to be useful.

These docs explain what the feature is, its use cases, and how to set it up on your agents.

Issue: [PDP-1288](https://linear.app/buildkite/issue/PDP-1288/update-bk-documentation)

### DIFF
--- a/data/nav.yml
+++ b/data/nav.yml
@@ -84,6 +84,8 @@
           path: "agent/v3/prioritization"
         - name: "Securing"
           path: "agent/v3/securing"
+        - name: "Signed pipelines"
+          path: "agent/v3/signed-pipelines"
         - name: "Tokens"
           path: "agent/v3/tokens"
         - name: "Tracing"
--- a/pages/agent/v3/signed_pipelines.md
+++ b/pages/agent/v3/signed_pipelines.md
@@ -1,0 +1,153 @@
+# Signed pipelines
+
+> 🚧 Experimental feature
+> Signed pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature, so if you'd like to try using signed pipelines, reach out to [support](https://buildkite.com/support).
+
+Signed pipelines are a security feature where pipelines are cryptographically signed when uploaded to Buildkite. Agents then verify the signature before running the job. If an agent detects a signature mismatch, it'll refuse to run the job.
+
+Maintaining a strong security boundary is important to Buildkite and informs how we design features. It's also a key reason people choose Buildkite over other CI/CD tools. Signing pipelines improves your security posture by ensuring agents don't run jobs where a malicious actor has modified the instructions. This moves you towards zero-trust CI/CD by further isolating you from Buildkite itself being compromised.
+
+The signature guarantees the origin of jobs by asserting:
+
+- The jobs were uploaded from a trusted source.
+- The jobs haven't been modified after upload.
+
+These signatures mean that if a threat actor could modify a job in flight, the agent would refuse to run it due to mismatched signatures.
+
+<details>
+  <summary>🤔 I think I've seen this before...</summary>
+  <p>This work is inspired by the <a href="https://github.com/buildkite/buildkite-signed-pipeline"><code>buildkite-signed-pipeline</code></a> tool, which you could add to your agent instances. It had a similar idea—signing steps before they're uploaded to Buildkite, then verifying them when they're run. However, it had some limitations, including:</p>
+  <ul>
+    <li>It had to be installed on every agent instance, leading to more configuration.</li>
+    <li>It only supported symmetric HS256 signatures, meaning that every verifier could also sign uploads.</li>
+    <li>It couldn't sign <a href="/docs/pipelines/build-matrix">matrix steps</a>.</li>
+  </ul>
+  <p>This newer version of pipeline signing is built right into the agent and addresses all of these limitations. Being built into the agent, it's also easier to configure and use.</p>
+  <p>Many thanks to <a href="https://www.seek.com.au/">SEEK</a>, who we collaborated with on the older version of the tool, and whose prior art has been instrumental in the development of this newer version.</p>
+</details>
+
+## Pipeline signatures
+
+Pipeline signatures establish that important aspects of steps haven't been changed since they were uploaded.
+
+The following fields are included in the signature for each step:
+
+- **Commands.**
+- **Environment variables defined in the pipeline YAML.** Environment variables set by the agent, hooks, or the user's shell are _not_ signed, and can override the environment a step's command is started with.
+- **Plugins and plugin configuration.**
+- **Matrix configuration.** The matrix configuration is signed as a whole rather than each individual matrix job. This means the signature is the same for each job in the matrix. When signatures are verified for matrix jobs, the agent double-checks that the job it received is a valid construction of the matrix and that the signature matches the matrix configuration.
+- **The repository the commands are running in.** This prevents you from copying a signed step from one repository to another.
+
+## Enabling signed pipelines on your agents
+
+You'll need to configure your agents and update pipeline definitions to enable signed pipelines.
+
+### 1. Generate a key pair
+
+Behind the scenes, signed pipelines use [JSON Web Signing (JWS)](https://datatracker.ietf.org/doc/html/rfc7797) to generate signatures. You'll need to generate a [JSON Web Key Set (JWKS)](https://datatracker.ietf.org/doc/html/rfc7517) to sign and verify your pipelines with, then configure your agents to use those keys.
+
+Luckily, the agent has you covered! A JWKS generation tool is built into the agent, which you can use to generate a key pair. To use it, you'll need to [install the agent on your machine](/docs/agent/v3/installation), and then run:
+
+```bash
+buildkite-agent tool keygen --alg <algorithm> --key-id <key-id>
+```
+
+Replacing the following:
+
+- `<algorithm>` with the signing algorithm you want to use.
+- `<key-id>` with the key ID you want to use.
+
+For example, to generate an [EdDSA](https://en.wikipedia.org/wiki/EdDSA) key pair with a key ID of `my-key-id`, you'd run:
+
+```bash
+buildkite-agent tool keygen --alg EdDSA --key-id my-key-id
+```
+
+The agent generates a JWKS key pair in your current directory: one private and one public. You can then use these keys to sign and verify your pipelines.
+
+If no key ID is provided, the agent will generate a random one for you.
+
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
+
+<details>
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  <p>In short, RSASSA-PKCS1 v1.5 signatures are less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are more secure than RSA signatures.</p>
+</details>
+
+#### Algorithm options
+
+Signed pipelines support many different signing algorithms, which you can simplify into two categories:
+
+- **Symmetric**—where signing and verification use the same keys.
+- **Asymmetric**—where signing and verification use different keys.
+
+We recommend using an asymmetric signing algorithm, as it creates a permissions boundary between the agents that upload and sign pipelines, and agents that run other jobs and can verify signatures. This means that if an attacker compromises an agent with verification keys, they can't modify the steps without detection.
+
+When using asymmetric signing algorithms, we recommend having multiple disjoint pools of agents, each using a different [queue](/docs/agent/v3/queues). One pool should be the _uploaders_ and have access to the private keys. Another pool should be the _runners_ and have access to the public keys. This creates a security boundary between the agents that upload and sign pipelines and the agents that run jobs and verify signatures.
+
+Regarding your specific algorithm choice, any of the supported asymmetric signing algorithms are fine and will be secure. If you're not sure which one to use, `EdDSA` is proven to be secure, has a modern design, wasn't designed by a Nation State Actor, and produces nice short signatures.
+
+For more information on what algorithms the agent supports, run:
+
+```sh
+buildkite-agent tool keygen --help
+```
+
+### 2. Configure the agents
+
+Next, you need to configure your agents to use the keys you generated. On agents that upload pipelines, add the following to the agent's config file:
+
+```ini
+signing-jwks-file=<path to private key set>
+signing-jwks-key-id=<the key id you generated earlier>
+verification-jwks-file=<path to public key set>
+```
+
+This ensures that whenever those agents upload steps to Buildkite, they'll generate signatures using the private key you generated earlier. It also ensures that those agents verify the signatures of any steps they run, using the public key.
+
+On instances that verify jobs, add:
+
+```ini
+verification-jwks-file=<path to verification keys>
+```
+
+### 3. Sign all steps
+
+So far, you've configured agents to sign and verify any steps they upload and run. However, you also define steps in a pipeline's settings through the Buildkite dashboard. For example, teams commonly use a single step in the Buildkite dashboard to upload a pipeline definition from [a YAML file in the repository](/docs/pipelines/defining-steps#step-defaults-pipeline-dot-yml-file). These steps should also be signed.
+
+> 🚧 Non-YAML steps
+> You must use YAML to sign steps configured in the Buildkite dashboard. If you don't use YAML, you'll need to [migrate to YAML steps](/docs/tutorials/pipeline_upgrade) before continuing.
+
+To sign steps configured in the Buildkite dashboard, you need to add static signatures to the YAML. To do this, run:
+
+```sh
+buildkite-agent tool sign \
+  --graphql-token <token> \
+  --jwks-file <path to signing jwks> \
+  --jwks-key-id <signing key id> \
+  --organization-slug <org slug> \
+  --pipeline-slug <pipeline slug> \
+  --update
+```
+
+Replacing the following:
+
+- `<token>` with a Buildkite GraphQL token that has the `write_pipelines` scope.
+- `<path to signing jwks>` with the path to the private key set you generated earlier.
+- `<signing key id>` with the key ID from earlier.
+- `<org slug>` with the slug of the organization the pipeline is in.
+- `<pipeline slug>` with the slug of the pipeline you want to sign.
+
+This will download the pipeline definition using the Buildkite GraphQL API, sign all steps, and upload the signed pipeline definition back to Buildkite.
+
+## Rotating signing keys
+
+Regularly rotating signing and verification keys is good security practice, as it reduces the impact of a compromised key. Because signed pipelines use JWKS as their key format, rotating keys is easy.
+
+To rotate your keys:
+
+1. [Generate a new key pair](#enabling-signed-pipelines-on-your-agents-1-generate-a-key-pair).
+1. Add the new keys to your existing key sets. Be careful not to mix public and private keys.
+1. Update the `signing-key-id` on your signing agents to use the new key ID.
+
+The verifying agents will automatically use the public key with the matching key ID, if it's present.
--- a/pages/agent/v3/signed_pipelines.md
+++ b/pages/agent/v3/signed_pipelines.md
@@ -1,8 +1,5 @@
 # Signed pipelines
 
-> 🚧 Experimental feature
-> Signed pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature, so if you'd like to try using signed pipelines, reach out to [support](https://buildkite.com/support).
-
 Signed pipelines are a security feature where pipelines are cryptographically signed when uploaded to Buildkite. Agents then verify the signature before running the job. If an agent detects a signature mismatch, it'll refuse to run the job.
 
 Maintaining a strong security boundary is important to Buildkite and informs how we design features. It's also a key reason people choose Buildkite over other CI/CD tools. Signing pipelines improves your security posture by ensuring agents don't run jobs where a malicious actor has modified the instructions. This moves you towards zero-trust CI/CD by further isolating you from Buildkite itself being compromised.
--- a/pages/agent/v3/signed_pipelines.md
+++ b/pages/agent/v3/signed_pipelines.md
@@ -16,7 +16,7 @@ These signatures mean that if a threat actor could modify a job in flight, the a
   <p>This work is inspired by the <a href="https://github.com/buildkite/buildkite-signed-pipeline"><code>buildkite-signed-pipeline</code></a> tool, which you could add to your agent instances. It had a similar idea—signing steps before they're uploaded to Buildkite, then verifying them when they're run. However, it had some limitations, including:</p>
   <ul>
     <li>It had to be installed on every agent instance, leading to more configuration.</li>
-    <li>It only supported symmetric HS256 signatures, meaning that every verifier could also sign uploads.</li>
+    <li>It only supported symmetric signatures (using HMAC-SHA256), meaning that every verifier could also sign uploads.</li>
     <li>It couldn't sign <a href="/docs/pipelines/build-matrix">matrix steps</a>.</li>
   </ul>
   <p>This newer version of pipeline signing is built right into the agent and addresses all of these limitations. Being built into the agent, it's also easier to configure and use.</p>
@@ -54,6 +54,8 @@ Replacing the following:
 - `<algorithm>` with the signing algorithm you want to use.
 - `<key-id>` with the key ID you want to use.
 
+Note that both the algorithm and key ID are optional - if `alg` isn't provided, the agent will default to `EdDSA`. If `key-id` isn't provided, the agent will generate a random one for you.
+
 For example, to generate an [EdDSA](https://en.wikipedia.org/wiki/EdDSA) key pair with a key ID of `my-key-id`, you'd run:
 
 ```bash
@@ -62,9 +64,16 @@ buildkite-agent tool keygen --alg EdDSA --key-id my-key-id
 
 The agent generates a JWKS key pair in your current directory: one private and one public. You can then use these keys to sign and verify your pipelines.
 
-If no key ID is provided, the agent will generate a random one for you.
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3), and that the agent does not support all JWA signing algorithms. At the time of writing, the agent supports:
+- `EdDSA` (the default)
+- `PS512`
+- `ES512`
 
-Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
+For an up-to-date list of supported algorithms, run:
+
+```sh
+buildkite-agent tool keygen --help
+```
 
 <details>
   <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
@@ -73,22 +82,9 @@ Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](http
 
 #### Algorithm options
 
-Signed pipelines support many different signing algorithms, which you can simplify into two categories:
+When using signed pipelines, we recommend having multiple disjoint pools of agents, each using a different [queue](/docs/agent/v3/queues). One pool should be the _uploaders_ and have access to the private keys. Another pool should be the _runners_ and have access to the public keys. This creates a security boundary between the agents that upload and sign pipelines and the agents that run jobs and verify signatures.
 
-- **Symmetric**—where signing and verification use the same keys.
-- **Asymmetric**—where signing and verification use different keys.
-
-We recommend using an asymmetric signing algorithm, as it creates a permissions boundary between the agents that upload and sign pipelines, and agents that run other jobs and can verify signatures. This means that if an attacker compromises an agent with verification keys, they can't modify the steps without detection.
-
-When using asymmetric signing algorithms, we recommend having multiple disjoint pools of agents, each using a different [queue](/docs/agent/v3/queues). One pool should be the _uploaders_ and have access to the private keys. Another pool should be the _runners_ and have access to the public keys. This creates a security boundary between the agents that upload and sign pipelines and the agents that run jobs and verify signatures.
-
-Regarding your specific algorithm choice, any of the supported asymmetric signing algorithms are fine and will be secure. If you're not sure which one to use, `EdDSA` is proven to be secure, has a modern design, wasn't designed by a Nation State Actor, and produces nice short signatures.
-
-For more information on what algorithms the agent supports, run:
-
-```sh
-buildkite-agent tool keygen --help
-```
+Regarding your specific algorithm choice, any of the supported signing algorithms are fine and will be secure. If you're not sure which one to use, `EdDSA` is proven to be secure, has a modern design, wasn't designed by a Nation State Actor, and produces nice short signatures. It's also the default when running `buildkite-agent tool keygen`.
 
 ### 2. Configure the agents
 
--- a/pages/agent/v3/securing.md
+++ b/pages/agent/v3/securing.md
@@ -120,7 +120,7 @@ But also remember that some [environment variables may be essential](/docs/pipel
 
 ## Signing pipelines
 
-If using plugins is crucial to your workflow and you would still like to preserve strong security guarantees, take a look at the [buildkite-signed-pipeline](https://github.com/buildkite/buildkite-signed-pipeline) tool. This tool allows uploaded steps to be signed with a secret shared by all agents, so that plugins can run without concerns of tampering by third parties.
+You can sign the steps your pipeline runs for extra security. This allows the agent to verify that the steps it runs haven't been tampered with or smuggled from one pipeline to another. For more information, see [Signed pipelines](/docs/agent/v3/signed-pipelines).
 
 ## Allowing a list of plugins
 
--- a/pages/apis/rest_api/builds.md
+++ b/pages/apis/rest_api/builds.md
@@ -92,8 +92,8 @@ curl "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pipeline.
         "step": {
           "id": "018c0f56-c87c-47e9-95ee-aa47397b4496",
           "signature": {
-            "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..gcmV8TgY6MfAXRbD_7kyuB3JAQMH-0th5E1MmQn5ND34bEs-zNBZ3tLi6w2e_gvfqL6C1kxRhOQzAwEZMQlo5w",
-            "algorithm": "HS512",
+            "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..m9LBvNgbzmO5JuZ4Bwoheyn7uqLf3TN1EdFwv_l_nMT2qh0_2EVs30SAEc-Ajjkq18MQk3cgU36AodLPl3_hBg",
+            "algorithm": "EdDSA",
             "signed_fields": [
               "command",
               "env",
@@ -227,8 +227,8 @@ curl "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pipeline.
       "step": {
         "id": "018c0f56-c87c-47e9-95ee-aa47397b4496",
         "signature": {
-          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..gcmV8TgY6MfAXRbD_7kyuB3JAQMH-0th5E1MmQn5ND34bEs-zNBZ3tLi6w2e_gvfqL6C1kxRhOQzAwEZMQlo5w",
-          "algorithm": "HS512",
+          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..m9LBvNgbzmO5JuZ4Bwoheyn7uqLf3TN1EdFwv_l_nMT2qh0_2EVs30SAEc-Ajjkq18MQk3cgU36AodLPl3_hBg",
+          "algorithm": "EdDSA",
           "signed_fields": [
             "command",
             "env",
@@ -388,8 +388,8 @@ curl -X POST "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{p
       "step": {
         "id": "018c0f56-c87c-47e9-95ee-aa47397b4496",
         "signature": {
-          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..gcmV8TgY6MfAXRbD_7kyuB3JAQMH-0th5E1MmQn5ND34bEs-zNBZ3tLi6w2e_gvfqL6C1kxRhOQzAwEZMQlo5w",
-          "algorithm": "HS512",
+          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..m9LBvNgbzmO5JuZ4Bwoheyn7uqLf3TN1EdFwv_l_nMT2qh0_2EVs30SAEc-Ajjkq18MQk3cgU36AodLPl3_hBg",
+          "algorithm": "EdDSA",
           "signed_fields": [
             "command",
             "env",
@@ -553,8 +553,8 @@ curl -X PUT "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pi
       "step": {
         "id": "018c0f56-c87c-47e9-95ee-aa47397b4496",
         "signature": {
-          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..gcmV8TgY6MfAXRbD_7kyuB3JAQMH-0th5E1MmQn5ND34bEs-zNBZ3tLi6w2e_gvfqL6C1kxRhOQzAwEZMQlo5w",
-          "algorithm": "HS512",
+          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..m9LBvNgbzmO5JuZ4Bwoheyn7uqLf3TN1EdFwv_l_nMT2qh0_2EVs30SAEc-Ajjkq18MQk3cgU36AodLPl3_hBg",
+          "algorithm": "EdDSA",
           "signed_fields": [
             "command",
             "env",
@@ -692,8 +692,8 @@ curl -X PUT "https://api.buildkite.com/v2/organizations/{org.slug}/pipelines/{pi
       "step": {
         "id": "018c0f56-c87c-47e9-95ee-aa47397b4496",
         "signature": {
-          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..gcmV8TgY6MfAXRbD_7kyuB3JAQMH-0th5E1MmQn5ND34bEs-zNBZ3tLi6w2e_gvfqL6C1kxRhOQzAwEZMQlo5w",
-          "algorithm": "HS512",
+          "value": "eyJhbGciOiJFUzI1NiIsImtpZCI6InlvdSBzbHkgZG9nISB5b3UgY2F1Z2h0IG1lIG1vbm9sb2d1aW5nISJ9..m9LBvNgbzmO5JuZ4Bwoheyn7uqLf3TN1EdFwv_l_nMT2qh0_2EVs30SAEc-Ajjkq18MQk3cgU36AodLPl3_hBg",
+          "algorithm": "EdDSA",
           "signed_fields": [
             "command",
             "env",
--- a/vale/styles/Buildkite/h1-h6_sentence_case.yml
+++ b/vale/styles/Buildkite/h1-h6_sentence_case.yml
@@ -10,6 +10,7 @@ scope:
   - heading.h6
 match: $sentence
 exceptions:
+  - '^(\d+\.) [A-Z]\w+' # Allow numbered headings. Example: "1. Install the thing"
   - '^(Agent|Build|Job) ([A-Z]\w+(ed|ing)|Lost)$' # Exceptions for /pages/integrations/amazon_eventbridge.md
   - '^(Step|Method) \d: [A-Z]\w+' # Allow caps after colon. Examples: "Step 3: Do the thing", "Method 2: Deploy keys"
   - ADFS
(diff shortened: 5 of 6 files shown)


pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+No
+
+### Signing the pipeline's stored configuration
+
+[ed: this section feels clunky to me, and i'm not sure the best way to explain it - we need to sign the initial steps of a pipeline as well as the steps that get uploaded by an agent, but it's difficult to concisely convey all the concepts involved]
+
+In Buildkite, pipelines have two YAML definitions - one that's stored Buildkite, which you can change on the settings page for a given pipeline, and other(s) that are uploaded by agents. In this guide, we'll refer to the former as the "stored" configuration, and the latter as "uploaded" configuration. The stored configuration generally contains the initial steps needed to run a pipeline, and almost always contains a step that uploads a pipeline to Buildkite.
+
+> Some older Buildkite pipelines don't use YAML for their stored pipeline configuration - these aren't supported by signed pipelines, and will have to be migrated to use the new format.
```
moskyb, line 96:
```suggestion
> Some older Buildkite pipelines don't use YAML for their stored pipeline configuration - these can't be signed, and will have to be migrated to use the new format to be used with signed pipelines
```

data/nav.yml
```diff
@@ -84,6 +84,8 @@
           path: "agent/v3/prioritization"
         - name: "Securing"
           path: "agent/v3/securing"
```
dabarrell, line 86: Note for later - worth thinking about where we want to link to the signed pipelines docs from elsewhere. I imagine it'd make sense to mention on "Securing" at a minimum

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
+# Signed pipelines
+
+## What are signed pipelines?
+
+Signed Pipelines is a security feature on the Buildkite Agent that allows you to cryptographically sign steps that are uploaded Buildkite, and then verify that signature on the Buildkite Agent before running the step. This can prevent a malicious actor from modifying the steps that are run on your agents, especially in the (hopefully unlikely) event that the Buildkite backend is compromised.
```
dabarrell, line 5: > especially in the (hopefully unlikely) event that the Buildkite backend is compromised.

hmmmmmmmmm. Not sure the best way to talk about this. What this really allows is a zero-trust model. Perhaps something like "This allows you to assert that the steps the agent is being instructed to run were uploaded by your own agents, removing the necessity to trust Buildkite." ? Along those lines, not necessarily those words

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+These signatures mean that while a threat actor could modify jobs in flight, when the Agent runs those steps, when it tries to verify the signature, it will fail, and the step will not be run.
```
dabarrell, line 7:
```suggestion
These signatures mean that if a threat actor were able to modify a job in flight, the Agent would refuse to run it due to mismatched signatures.
```

Simpler, and changes from "while something bad can happen" to "if it were somehow possible" - slightly different language

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+  <ul>
+    <li>It had to be installed on every agent instance, leading to more configuration</li>
+    <li>It only supported symmetric HS256 signatures, meaning that every verifier could also sign uploads</li>
+    <li>It couldn't sign matrix steps</li>
+  </ul>
+
+  This newer version of pipeline signing is built right into the agent, and addresses all of these limitations. Being built into the agent, it's easier to configure and use.
+
+  Many thanks to Seek.com.au, who we collaborated with on the older version of the tool, and who have been instrumental in the development of this newer version.
+</details>
```
dabarrell, line 21: This'll need a pass from someone with a better sense than I of the tone we use in our docs

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Enabling signed pipelines on your pipelines
```
dabarrell, line 23:
```suggestion
## Enabling signed pipelines on your agents
```
Less repetition?

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+### How do I generate a key pair?
+
+Luckily, the Buildkite Agent has you covered! There's a JWKS generation tool built into the agent, which you can use to generate a key pair for you. To run it, you'll need to [install the agent on your machine](/docs/agent/v3/installation), and then run:
+```bash
+buildkite-agent tool keygen --alg <algorithm> --key-id <key-id>
+```
+
+replacing `<algorithm>` and `<key-id>` with the algorithm you want to use, and the key ID you want to use. For example, to generate an RSA-PSS key pair with a key ID of `my-key-id`, you'd run:
+```bash
+buildkite-agent tool keygen --alg PS256 --key-id my-key-id
```
dabarrell, line 36: If we're recommending `EdDSA`, then I'd suggest using that here. People will just copy and paste

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+The agent will then generate two JSON Web Key Sets for you, one public and one private, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines.
```
dabarrell, line 39:
```suggestion
The agent will then generate two JSON Web Key Sets for you, one private and one public, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines, respectively.
```

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+<details>
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  In short, it's because RSASSA-PKCS1 v1.5 signatures are generally regarded to be less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still largely known to be relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are generally regarded to be more secure than RSA signatures.
+</details>
+
+### What algorithm should I use?
+
+Signed Pipelines supports a number of different signing algorithms, which can largely be broken down into two categories: Symmetric Signing Algorithms (where signing and verification use the same keys), and Asymmetric Signing Algorithms (where signing and verification use different keys).
```
dabarrell, line 52: Are Symmetric Signing Algorithms / Asymmetric Signing Algorithms proper terms? I.e. should they be capitalised? or
```suggestion
Signed Pipelines supports a number of different signing algorithms, which can largely be broken down into two categories: symmetric algorithms (where signing and verification use the same keys), and asymmetric algorithms (where signing and verification use different keys).
```

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+- `ES512`
+- `EdDSA`
+
+### Making your agents sign steps they upload 
+
+Now that your keys are generated, we need to configure your agents to use them. On agents that will be uploading pipelines, add the following to your agent's config file:
+```ini
+job-signing-jwks-path=<path to signing keys>
+signing-key-id=<the key id you generated earlier>
+job-verification-jwks-path=<path to verification keys>
```
dabarrell, line 80: Does the signing agent require the verification key? If so, I'd note here why - "the signing agent requires the verification key as well in order to validate the signature" or whatever the reason is.

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+You may have noticed that when we configured the signing instances, we gave them the capability to both sign and verify jobs. For maximum security, the stored pipeline will need to be
+
+The solution here is to add static signatures to the steps in your pipeline's stored configuration.
+
+[ed: explain how to do this - buildkite-agent pipeline upload --jwks-file-path /path/to/keys --signing-key-id blah --dry-run and copy and paste into buildkite's UI]
+
+### Hard mode: static signatures
+
+[ed: if you're really paranoid, you can sign a pipeline once, and hardcode the signatures into your pipeline yaml. this means that if anyone changes anything on the signed steps, your jobs will fail]
```
dabarrell, line 106: What's the advantage to doing this? Is it so you don't have to trust agent updates that might introduce vulnerabilities/attacks via the signing process?

dabarrell: Just wondering if it's worth even explaining here. You already explain signing stored configuration above - not that hard to connect the dots

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Rotating signing keys
+
+[ed: generate new keys, and add them to the existing key sets you've already generated. switch over the --signing-key-id and bob's your uncle]
+
+## General guidance for running signed pipelines
+
+[ed: something about using different queues for pipeline uploads vs everything else, then give uploaders the singing key and runners the verification key]
```
dabarrell, line 114: I'd bump this section up below the stored config part

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,114 @@
...
+job-verification-jwks-path=<path to verification keys>
+```
+
+No
+
+### Signing the pipeline's stored configuration
+
+[ed: this section feels clunky to me, and i'm not sure the best way to explain it - we need to sign the initial steps of a pipeline as well as the steps that get uploaded by an agent, but it's difficult to concisely convey all the concepts involved]
+
+In Buildkite, pipelines have two YAML definitions - one that's stored Buildkite, which you can change on the settings page for a given pipeline, and other(s) that are uploaded by agents. In this guide, we'll refer to the former as the "stored" configuration, and the latter as "uploaded" configuration. The stored configuration generally contains the initial steps needed to run a pipeline, and almost always contains a step that uploads a pipeline to Buildkite.
```
dabarrell, line 94: Perhaps something like 
```suggestion
Builds in Buildkite pipelines are initialized using steps that are configured on the pipeline itself. In the course of running a job, the Buildkite Agent can [upload further steps to the pipeline](link to some info about this elsewhere in the docs). We will refer to the former as "configured steps" and the latter as "uploaded steps". It is common for the configured steps to be a single step that uploads the remainder of pipeline from [a YAML file in the repository](link to info about ./buildkite.yaml).
```

dabarrell: "To ensure that all steps in a pipeline are signed, you will need to manually sign the configured steps. To do this..." etc

dabarrell (commented):
Really good start! I've dropped some thoughts, not focusing on language, but on content

pages/agent/v3/signed-pipelines.md
moskyb: ooooh yesssssssss very good. will steal that

pages/agent/v3/signed-pipelines.md
moskyb: yep, good call - they should be lowercase

pages/agent/v3/signed-pipelines.md
moskyb: ahah, i was in the process of writing that section when i committed this. tl;dr, if we statically sign the initial steps, the uploader needs to both be able to verify the signature on the initial steps, and sign the steps it uploads from the repo

pages/agent/v3/signed-pipelines.md
moskyb: ooooooh love it

pages/agent/v3/signed-pipelines.md
dabarrell: Ah of course!

pages/agent/v3/signed-pipelines.md
moskyb: the advantage it offers over "dynamic" signing (ie signing the pipeline each time it's uploaded) is that it protects against the case of "someone got into my repo and turned all of my test scripts into bitcoin miners", which is definitely something i think at least some of our customers would appreciate. the downside is that basically every time you edit the pipeline.yml you'd have to re-sign it, which could get to be a pain

moskyb: given that we're going to be signing build invariants, this approach has a lot less utility - i'm gonna yeet it

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+You may have noticed that when we configured the signing instances, we gave them the capability to both sign and verify jobs. For maximum security, the configured steps - usually a single step that uploads (and signs!) the remainder of the pipeline - should themselves be signed.
+
+The solution here is to add static signatures to the steps in your pipeline's stored configuration.
+
+To do this, copy the YAML from the editor on the pipeline's page in Buildkite, and paste it into a file on your machine. Then, run:
+```bash
+buildkite-agent pipeline upload --jwks-file-path <path to private keys> --signing-key-id <key id> --dry-run <path to pipeline file>
+```
+
+for example, if you had a file called `pipeline.yml` in the current directory, and you wanted to sign it with the key we generated earlier, you'd run:
```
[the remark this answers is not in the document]
dabarrell: Actually - I think it'd help to differentiate a bit more. "file called `pipeline.yml` in the current directory" makes me think of the pipeline file in the repo. Perhaps

```suggestion
For example, if your pipeline's configured steps were available as `configured-steps.yaml` in the current directory, and you wanted to sign it with the key we generated earlier, you'd run:
```

(and update the command)

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+```bash
+buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run pipeline.yml
+```
+
+This will output the pipeline YAML, with the signatures added to the steps. You can then copy this YAML back into the editor on the pipeline's page in Buildkite, and save it.
+
+## Rotating signing keys
+
+Because signed pipelines use JWKS as their key format, rotating keys is easy. To rotate your keys, simply generate a new key pair, and add the new keys to your existing key sets, being careful not to mix public and private keys. Once the updated key sets are in place, you can update the `signing-key-id` on your signing agents to use the new key ID. Verifying agents will automatically the public key with the matching key ID, if it's present.
```
dabarrell, line 122: >Verifying agents will automatically [use] the public key with the matching key ID, if it's present.

Isn't it signing agents that need the key ID? and verifying agents just try all the keys to see if one works? Or is the key ID now in the signature itself, so the verifying agent can reference that?

dabarrell (changes requested):
Some more comments. Also looks like you've accidentally pushed an update to the `emojis` submodule

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+</details>
+
+### What we actually sign
+
+As mentioned above, signed pipelines establishes that important aspects of steps haven't been changed since pipeline upload. In more detail, we sign the following aspects of steps:
+- The command
+- Environment variables _as defined in the pipeline YAML_ - environment variables set by the agent or by the user's shell are not signed
+- Any plugins used by the step, and their configurations
+- The matrix configuration [^1]
+- The step's key, if present
```
DrJosh9000, line 30: This would currently be WIP

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+
+The agent will then generate two JSON Web Key Sets for you, one public and one private, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines.
+
+If no key id is provided, the agent will generate a random one for you.
+
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
+
+<details>
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  In short, it's because RSASSA-PKCS1 v1.5 signatures are generally regarded to be less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still largely known to be relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are generally regarded to be more secure than RSA signatures.
```
triarius, line 65:
```suggestion
  In short, it's because RSASSA-PKCS1 v1.5 signatures are generally regarded to be less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still largely known to be relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures.
```
I'd leave this out because it may not be accurate under quantum attacks. It looks like because ECC uses keys that are an order of magnitude shorter, they may be broken by smaller quantum computers than that required to break RSA that would be weaker classically: see https://security.stackexchange.com/questions/33069/why-is-ecc-more-vulnerable-than-rsa-in-a-post-quantum-world

Of course, no such quantum computer is known to exist.

pages/agent/v3/signed-pipelines.md
triarius, line 122:
```suggestion
Because signed pipelines use JWKS as their key format, rotating keys is easy. To rotate your keys, simply generate a new key pair, and add the new keys to your existing key sets, being careful not to mix public and private keys. Once the updated key sets are in place, you can update the `signing-key-id` on your signing agents to use the new key ID. Verifying agents will automatically use the public key with the matching key ID, if it's present.
```

triarius (commented):
Looking great! Flushing some initial comments out.

pages/agent/v3/signed-pipelines.md
dabarrell: We should get Ashwin's take on this all too

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+  This newer version of pipeline signing is built right into the agent, and addresses all of these limitations. Being built into the agent, it's easier to configure and use.
+
+  Many thanks to Seek.com.au, who we collaborated with on the older version of the tool, and whose prior art has been instrumental in the development of this newer version.
+</details>
+
+### What we actually sign
+
+As mentioned above, signed pipelines establishes that important aspects of steps haven't been changed since pipeline upload. In more detail, we sign the following aspects of steps:
+- The command
+- Environment variables _as defined in the pipeline YAML_ - environment variables set by the agent or by the user's shell are not signed
```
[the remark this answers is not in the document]
DrJosh9000: There could be a command hook that totally ignores the signed command, too. Perhaps the more general principle is that we're making some guarantees about the _initial_ state of signed and verified jobs.

pages/agent/v3/signed-pipelines.md
moskyb: the key ID is encoded in the header of the signature - here's one i pulled from the agent pipeline: `eyJhbGciOiJFUzI1NiIsImtpZCI6IjEifQ..Ue0tFd7jaZj66huix2TZ1UanAB4vHlWZn1utvXQDjSMmJDQ1wWVG8fDCdboZrUcnxzNen1MaU8NfX_vOr5T7sQ`

when you base64 decode the first segment there, it gives: `{"alg":"ES256","kid":"1"}` - our key id in the agent is imaginatively called `1`. Our library then pulls this out of the header and looks for a key with that key ID in the verification JWKS.

if we didn't include a key ID (which isn't allowed as is), the library would then try each key in sequence.

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+
+> Some older Buildkite pipelines don't use YAML for their stored pipeline configuration - these can't be signed, and will have to be migrated to use the new format to be used with signed pipelines - see [Migrating to YAML steps](/docs/tutorials/pipeline_upgrade)
+
+You may have noticed that when we configured the signing instances, we gave them the capability to both sign and verify jobs. For maximum security, the configured steps - usually a single step that uploads (and signs!) the remainder of the pipeline - should themselves be signed.
+
+The solution here is to add static signatures to the steps in your pipeline's stored configuration.
+
+To do this, copy the YAML from the editor on the pipeline's page in Buildkite, and paste it into a file on your machine. Then, run:
+```bash
+buildkite-agent pipeline upload --jwks-file-path <path to private keys> --signing-key-id <key id> --dry-run <path to pipeline file>
```
DrJosh9000, line 110:
```suggestion
buildkite-agent pipeline upload --jwks-file-path <path to private keys> --signing-key-id <key id> --dry-run --format yaml <path to pipeline file>
```

pages/agent/v3/signed-pipelines.md
dabarrell: Gotcha - thanks! In that case:
```suggestion
Because signed pipelines use JWKS as their key format, rotating keys is easy. To rotate your keys, simply generate a new key pair, and add the new keys to your existing key sets, being careful not to mix public and private keys. Once the updated key sets are in place, you can update the `signing-key-id` on your signing agents to use the new key ID. Verifying agents will automatically use the public key with the matching key ID, if it's present.
```

pages/agent/v3/signed-pipelines.md
moskyb:
```suggestion
- The step's key, if present [WIP]
```

pages/agent/v3/signed-pipelines.md
moskyb: oh whoops! i'll add it to this batch of edits.

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+for example, if you had a file called `pipeline.yml` in the current directory, and you wanted to sign it with the key we generated earlier, you'd run:
+```bash
+buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run pipeline.yml
```
moskyb, line 115:
```suggestion
buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run --format yaml pipeline.yml
```

DrJosh9000: Check .yml/.yaml consistency too

pages/agent/v3/signed-pipelines.md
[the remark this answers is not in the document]
moskyb: my lowercase tendencies slipped into the docs for a hot second there

also, configured-steps is a good call

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
...
+job-verification-jwks-path=<path to verification keys>
+```
+
+### 3. Signing the pipeline's stored configuration
+
+[ed: this section feels clunky to me, and i'm not sure the best way to explain it - we need to sign the initial steps of a pipeline as well as the steps that get uploaded by an agent, but it's difficult to concisely convey all the concepts involved]
+
+Builds in Buildkite pipelines are initialized using steps that are configured on the pipeline itself. In the course of running a job, the Buildkite Agent can [upload further steps to the pipeline](link to some info about this elsewhere in the docs). We will refer to the former as "configured steps" and the latter as "uploaded steps". It's common for the configured steps to be a single step that uploads the remainder of pipeline from [a YAML file in the repository](link to info about ./buildkite.yaml).
+
+> Some older Buildkite pipelines don't use YAML for their stored pipeline configuration - these can't be signed, and will have to be migrated to use the new format to be used with signed pipelines - see [Migrating to YAML steps](/docs/tutorials/pipeline_upgrade)
```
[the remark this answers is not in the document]
moskyb: yep, good call

pages/agent/v3/signed-pipelines.md
moskyb: yeah, stuff like this was why i hedged hard af whenever i talked about crypto - there's a lot of "generally regarded to be" and "thought to be" and "can be".

regardless, i don't mind this page to become a lie when quantum computing becomes available

pages/agent/v3/signed-pipelines.md
```diff
@@ -0,0 +1,122 @@
+# Signed pipelines
+
+## What are signed pipelines?
+
+Signed Pipelines is a security feature on the Buildkite Agent that allows you to cryptographically sign steps that are uploaded Buildkite, and then verify that signature on the Buildkite Agent before running the step. This signature helps agents establish the provenance of the jobs that they run - asserting that the steps were uploaded by a trusted agent, and that they haven't been modified since they were uploaded. If an agent detects a signature mismatch, by default, it'll refuse to run the job.
+
+These signatures mean that if a threat actor were able to modify a job in flight, the Agent would refuse to run it due to mismatched signatures.
```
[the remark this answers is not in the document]
moskyb: i think the pattern i fell into was `Buildkite Agent` gets Titlecase, `agent` on its own isn't a proper noun so it doesn't - that could be contra to policy though

moskyb:
@mbelton-buildkite would love a couple of eyes next time you're in the neighbourhood :)

i can fix the linting errors no worries, i'm mostly wondering about the prose

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+
+These signatures mean that if a threat actor were able to modify a job in flight, the Agent would refuse to run it due to mismatched signatures.
+
+<details>
+  <summary>I think I've seen this before...</summary>
+  This work is inspired by the older <a href="https://github.com/buildkite/buildkite-signed-pipeline"><code>buildkite-signed-pipeline</code></a> tool, which was a tool you could add to your agent instances. It had a similar idea - signing steps before they're uploaded to Buildkite, then verifying them when they're run. However, it had a number of limitations, including:
+  <ul>
+    <li>It had to be installed on every agent instance, leading to more configuration</li>
+    <li>It only supported symmetric HS256 signatures, meaning that every verifier could also sign uploads</li>
+    <li>It couldn't sign matrix steps</li>
```
DrJosh9000, line 17: Worth directly linking to matrix doc?

```suggestion
    <li>It couldn't sign [matrix](/docs/pipelines/build-matrix) steps</li>
```

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+> ⚠️ Warning: Signed Pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature - if you're keen to get your hands on it, talk to your sales rep.
+
+## What are signed pipelines?
+
+Signed Pipelines is a security feature on the Buildkite Agent that allows you to cryptographically sign steps that are uploaded to Buildkite, and then verify that signature on the Buildkite Agent before running the step. This signature helps agents establish the provenance of the jobs that they are instructed to run - asserting that the steps were uploaded by a trusted agent, and that they haven't been modified since they were uploaded. If an agent detects a signature mismatch, by default, it'll refuse to run the job.
+
+These signatures mean that if a threat actor were able to modify a job in flight, the Agent would refuse to run it due to mismatched signatures.
+
+<details>
+  <summary>I think I've seen this before...</summary>
```
DrJosh9000, line 12: Add an emoji? 🤔

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+
+The agent will then generate two JSON Web Key Sets for you, one private and one public, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines, respectively.
+
+If no key id is provided, the agent will generate a random one for you.
+
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
+
+<details>
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  In short, it's because RSASSA-PKCS1 v1.5 signatures are generally regarded to be less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still largely known to be relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are generally regarded to be more secure than RSA signatures.
```
DrJosh9000, line 71: We could add a reference to RFC 8017 (https://www.rfc-editor.org/rfc/rfc8017.html#section-8) as further justification , particularly, "Although no attacks are known against RSASSA-PKCS1-v1_5, in the interest of increased robustness, RSASSA-PSS is REQUIRED in new applications."

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
+# Signed pipelines
+
+> ⚠️ Warning: Signed Pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature - if you're keen to get your hands on it, talk to your sales rep.
```
mbelton-buildkite, line 3: Suggesting someone reach out to a sales rep is new for the docs and makes this sound very enterprise-y. Do smaller self-service customers have a sales rep? Could we handle this through support instead?

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+
+As mentioned above, Signed Pipelines establishes that important aspects of steps haven't been changed since pipeline upload. In more detail, we sign the following aspects of steps:
+
+- The command
+- Environment variables _as defined in the pipeline YAML_ - environment variables set by the agent, by hooks, or by the user's shell are not signed and can potentially override the environment that a step's command is started with.
+
+- Any plugins used by the step, and their configurations
+- The matrix configuration
+  + The matrix configuration is signed as a whole, rather than each individual matrix job. This means that the signature will be the same for each job in the matrix. When signatures are verified for matrix jobs, we double-check that the job we've received is a valid construction of the matrix, and that the signature matches the matrix configuration.
+- The step's key, if present [WIP]
```
mbelton-buildkite, line 35: What's the go with everything labelled "WIP?" Is it to communicate what work is coming next? If so, what's the timeline you're looking at and how confident are we in it?

(I would generally avoid documenting anything "coming soon," but worth considering in the context if you think it's important)

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+### 1. Generating a key pair
+Behind the scenes, Signed Pipelines uses [JSON Web Signing (JWS)](https://datatracker.ietf.org/doc/html/rfc7797) to generate its signatures. This means that you'll need to generate a [JSON Web Key Set (JWKS)](https://datatracker.ietf.org/doc/html/rfc7517) to sign and verify your pipelines with, then configure your agents to use those keys.
+
+Luckily, the Buildkite Agent has you covered! There's a JWKS generation tool built into the agent, which you can use to generate a key pair for you. To run it, you'll need to [install the agent on your machine](/docs/agent/v3/installation), and then run:
+
+```bash
+buildkite-agent tool keygen --alg <algorithm> --key-id <key-id>
+```
+
+replacing `<algorithm>` and `<key-id>` with the algorithm you want to use, and the key ID you want to use. For example, to generate an EdDSA key pair with a key ID of `my-key-id`, you'd run:
```
mbelton-buildkite, line 57: Is there a link we could include for EdDSA? Maybe just https://en.wikipedia.org/wiki/EdDSA ?

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
...
+
+```bash
+buildkite-agent tool keygen --alg EdDSA --key-id my-key-id
+```
+
+The agent will then generate two JSON Web Key Sets for you, one private and one public, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines, respectively.
+
+If no key id is provided, the agent will generate a random one for you.
+
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
```
mbelton-buildkite, line 67: Is there a link we can include for RSASSA-PKCS1 v1.5?

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,130 @@
+# Signed pipelines
+
+> ⚠️ Warning: Signed Pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature - if you're keen to get your hands on it, talk to your sales rep.
+
+## What are signed pipelines?
+
+Signed Pipelines is a security feature on the Buildkite Agent that allows you to cryptographically sign steps that are uploaded to Buildkite, and then verify that signature on the Buildkite Agent before running the step. This signature helps agents establish the provenance of the jobs that they are instructed to run - asserting that the steps were uploaded by a trusted agent, and that they haven't been modified since they were uploaded. If an agent detects a signature mismatch, by default, it'll refuse to run the job.
```
mbelton-buildkite, line 7: Is it important to mention "by default" here? I'm not sure it's adding much and there's no further mention of changing defaults on the rest of the page.

moskyb:
> there are a bunch of small tweaks to language and formatting that I'd like to make [...] Would it be okay if I push a commit to your branch?

please do!

pages/agent/v3/signed_pipelines.md
moskyb: hmm, we do offer an option to warn instead of block, but you're right, it's never mentioned here. i think removing the "by default" is totally valid

pages/agent/v3/signed_pipelines.md
mbelton-buildkite, line 71: This language is hedging our bets a bunch. Could we be more certain, or is it important to keep it loose?

Maybe:

> In short, RSASSA-PKCS1 v1.5 signatures are less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are more secure than RSA signatures.

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+
+Pipeline signatures establish that important aspects of steps haven't been changed since they were uploaded.
+
+The following fields are included in the signature for each step:
+
+- **Commands.**
+- **Environment variables defined in the pipeline YAML.** Environment variables set by the agent, hooks, or the user's shell are not signed and can override the environment a step's command is started with.
+- **Plugins and plugin configuration.**
+- **Matrix configuration.** The matrix configuration is signed as a whole rather than each individual matrix job. This means the signature is the same for each job in the matrix. When signatures are verified for matrix jobs, the agent double-checks that the job it received is a valid construction of the matrix and that the signature matches the matrix configuration.
+- **The step key.** If there is a step key present. [WIP]
```
mbelton-buildkite, line 37: Still want to check what WIP means here

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+
+The agent generates two JWKS in your current directory: one private and one public. You can then use these keys to sign and verify your pipelines.
+
+If no key ID is provided, the agent will generate a random one for you.
+
+Note that the value of `--alg` must be a valid [JSON Web Signing Algorithm](https://datatracker.ietf.org/doc/html/rfc7518#section-3). The agent does not support the `none` algorithm, or any signature algorithms in the `RS` series (any signing algorithms that use RSASSA-PKCS1 v1.5 signatures, `RS256` for example).
+
+<details>
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  <p>In short, RSASSA-PKCS1 v1.5 signatures are less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are more secure than RSA signatures.</p>
```
mbelton-buildkite, line 80: I went with the stronger language here, but happy to go back

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+
+This ensures that whenever those agents upload steps to Buildkite, they'll generate signatures using the private key you generated earlier. It also ensures that those agents verify the signatures of any steps they run, using the public key.
+
+On instances that verify jobs, add:
+
+```ini
+job-verification-jwks-path=<path to verification keys>
+```
+
+### 3. Sign all steps
```
mbelton-buildkite, line 120: I tried to avoid introducing new terms with "configured" and "uploaded" steps. Hopefully this works

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+```bash
+buildkite-agent tool keygen --alg <algorithm> --key-id <key-id>
+```
+
+Replacing the following:
+
+- `<algorithm>` with the signing algorithm you want to use.
+- `<key-id>` with the key ID you want to use.
+
+For example, to generate an EdDSA key pair with a key ID of `my-key-id`, you'd run:
```
mbelton-buildkite, line 66: I still think it would be good to add a link on "EdDSA"

mbelton-buildkite (commented):
Okay, so I went a bit further than I planned, so please check it's still correct. 😅

pages/agent/v3/signed_pipelines.md
```diff
@@ -45,86 +47,117 @@ Additionally, we also sign the following aspects of the build for each job:
...
 
 ```bash
 buildkite-agent tool keygen --alg EdDSA --key-id my-key-id
 ```
 
-The agent will then generate two JSON Web Key Sets for you, one private and one public, and output them to files in the current directory. You can then use these keys to sign and verify your pipelines, respectively.
+The agent generates two JWKS in your current directory: one private and one public. You can then use these keys to sign and verify your pipelines.
```
moskyb, line 72: the reason i didn't use the acronym here is that JWKS isn't plural (it stands for JSON Web Key Set), and we're generating two JWKSes, and it's a difficult word to pluralize 😅

pages/agent/v3/signed_pipelines.md
moskyb: in this case it's "we're working on it and we expect to merge it prior to this PR, but don't want to forget to add it to the docs"

pages/agent/v3/signed_pipelines.md
moskyb: https://en.wikipedia.org/wiki/EdDSA should be fine i think, though we go into a bit more detail about EdDSA in particular a bit further down

pages/agent/v3/signed_pipelines.md
moskyb: this is fine, i think - the reason i hedged so much is because i wanted to avoid cryptography nerds telling me i was wrong, but i'm _pretty sure_ everything written in there is either objectively correct, or a matter of interpretation

pages/agent/v3/signed_pipelines.md
```diff
@@ -45,86 +47,117 @@ Additionally, we also sign the following aspects of the build for each job:
...
 
 ```ini
 job-verification-jwks-path=<path to verification keys>
 ```
 
-### 3. Signing the pipeline's configured steps
+### 3. Sign all steps
 
-Builds in Buildkite pipelines are initialized using steps that are configured on the pipeline itself. In the course of running a job, the Buildkite Agent can upload further steps to the pipeline. We will refer to the former as "configured steps" and the latter as "uploaded steps". It's common for the configured steps to be a single step that uploads the remainder of pipeline from [a YAML file in the repository](/docs/pipelines/defining-steps#step-defaults-pipeline-dot-yml-file).
+So far, you've configured agents to sign and verify any steps they upload and run. But there's another source of steps: the Buildkite dashboard. Rather than defining steps in your repository, you can also define steps in a pipeline's settings. The most common use for this is to have a single step that uploads a pipeline definition from [a YAML file in the repository](/docs/pipelines/defining-steps#step-defaults-pipeline-dot-yml-file). These steps should also be signed.
```
moskyb, line 122: > Rather than defining steps in your repository, you can also define steps in a pipeline's settings.

this makes it seem a bit like it's an either or, but in 99.99% of cases, it's going to be both, which is why i introduced the terms... they're both of equal importance when it comes to pipelines, both signed and not. 

maybe replacing this paragraph with something like:

```suggestion
So far, you've configured agents to sign and verify any steps they upload and run. However, in Buildkite, steps are defined both in your repo, and in the Buildkite interface itself. The most common use for the steps in buildkite is a single step that uploads a pipeline definition from [a YAML file in the repository](/docs/pipelines/defining-steps#step-defaults-pipeline-dot-yml-file). The steps stored in buildkite will also need to be signed.
```

.... or something, i think there's still some magic that someone better at writing could put on that para 😅

pages/agent/v3/signed_pipelines.md
```diff
@@ -45,86 +47,117 @@ Additionally, we also sign the following aspects of the build for each job:
...
+
+1. Copy the updated step definitions back into the editor in the Buildkite dashboard.
+1. Save the new definition.
 
-For example, if you had a file called `pipeline.yaml` in the current directory, and you wanted to sign it with the key we generated earlier, you'd run:
+For example, if you had a file called `pipeline.yaml` in the current directory, and you wanted to sign it with the key generated earlier, you'd run:
 
 ```bash
-buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run --format yaml configured-steps.yaml
+buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run --format yaml pipeline.yaml
```
moskyb: potentially we should switch it so that they're both the same and _not_ `pipeline.yaml`, eg

```suggestion
For example, if you had a file called `configured-steps.yaml` in the current directory, and you wanted to sign it with the key generated earlier, you'd run:
```bash
buildkite-agent pipeline upload --jwks-file-path /path/to/private/key.json --signing-key-id my-key-id --dry-run --format yaml configured-steps.yaml
```

to avoid confusion with the _actual_ pipeline.yaml, which is getting signed when it gets uploaded 😅 there's a lot of overlapping nomenclature here...

potentially `configured-steps.yaml` no longer makes sense as we're not introducing that term anymore, but i think it should be something that isn't `pipeline.yaml`

moskyb:
ty for your changes!!! they're great

mbelton-buildkite:
Also FYI, I've flagged the Vale failures in the build with @dannymidnight who will take a look at how we can allow headings like "1. Heading."

data/nav.yml
dabarrell: @moskyb just reminding you of this one ^^

moskyb: i've added it to the file below: https://github.com/buildkite/docs/pull/2524/files#diff-b27d77ac575fad5917c6b6719e5898132289acc351e73db4dd27987dda9aef4e

dabarrell: Ah, apologies - missed it. Resolving :)

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
+# Signed pipelines
+
+> 🚧 Experimental feature
+> Signed pipelines are experimental, and these docs (and the commands and arguments referenced in them) are subject to change. This is also an opt-in feature, so if you'd like to try using signed pipelines, reach out to [support](https://buildkite.com/support).
+
+Signed pipelines are a security feature where pipelines are cryptographically signed when uploaded to Buildkite. Agents then verify the signature before running the job. If an agent detects a signature mismatch, it'll refuse to run the job.
```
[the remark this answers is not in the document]
123sarahj123: Secuirty->Overview
image

Might be worth checking this area of the docs also to see if you need to make additions for Signed Pipelines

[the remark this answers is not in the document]
123sarahj123: I had these sorts of statements in my old PR. Apologies if you have already covered it: 

> We know that builds and deploys can be run on highly privileged machines, and that an attacker convincing that machine to run a malicious command could compromise production infrastructure.
> As customer secuirty is paramount to Buildkite, and we wish to ensure that all customers are immune by default from our control plane being compromised.

> Signed pipelines allow users to sign the pipelines they upload, which means that agents can refuse to run jobs that are not signed by a trusted public key. This makes it impossible for Buildkite (or an attacker thereof) to tamper with the workload that Buildkite orchestrates onto customer infrastructure.

> Sensitive data, such as source code and secrets, remain within your own environment and are not seen by Buildkite. We are aware that many customers invest heavily in configuring their agents to reject jobs that don't match some set of expectations, such as pipeline/repo/branch filtering, limiting to in-repo scripts not arbitrary commands etc.

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+  <summary>Why doesn't the agent support RSASSA-PKCS1 v1.5 signatures?</summary>
+  <p>In short, RSASSA-PKCS1 v1.5 signatures are less secure than the newer RSA-PSS signatures. While RSASSA-PKCS1 v1.5 signatures are still relatively secure, we want to encourage our users to use the most secure algorithms possible, so when using RSA keys, we only support RSA-PSS signatures. We also recommend looking into ECDSA and EdDSA signatures, which are more secure than RSA signatures.</p>
+</details>
+
+#### Algorithm options
+
+Signed pipelines support many different signing algorithms, which you can simplify into two categories:
+
+- **Symmetric**—where signing and verification use the same keys.
+- **Asymmetric**—where signing and verification use different keys.
```
123sarahj123, line 88: loooove it

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+
+We recommend using an asymmetric signing algorithm, as it creates a permissions boundary between the agents that upload and sign pipelines, and agents that run other jobs and can verify signatures. This means that if an attacker compromises an agent with verification keys, they can't modify the steps without detection.
```
123sarahj123, line 90: 🔥🔥🔥🔥🔥🔥

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+## Rotating signing keys
+
+Because signed pipelines use JWKS as their key format, rotating keys is easy.
+
+To rotate your keys:
+
+1. Generate a new key pair.
+1. Add the new keys to your existing key sets. Be careful not to mix public and private keys.
```
123sarahj123, line 160:
```suggestion
2. Add the new keys to your existing key sets. Be careful not to mix public and private keys.
```

mbelton-buildkite: `1.` is correct for both list items. It displays as the right number when rendered and helps make future refactoring simpler. 😊

CleanShot 2023-11-21 at 09 59 16

123sarahj123: awesome thank you for the clarity :) I'll admit that I did not have it rendered when making this suggestion!

mbelton-buildkite (commented):
A few more suggestions, but I think this is very close! I'll reply to more of the existing comments, too

pages/agent/v3/signed_pipelines.md
moskyb: perfect!

pages/agent/v3/signed_pipelines.md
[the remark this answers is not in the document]
mbelton-buildkite: Maybe adding something like:

```suggestion
Signed pipelines are a security feature where pipelines are cryptographically signed when uploaded to Buildkite. Agents then verify the signature before running the job. If an agent detects a signature mismatch, it'll refuse to run the job.

Maintaining a strong security boundary is important to Buildkite and informs how we design features. It's also a key reason people choose Buildkite over other CI/CD tools. Signing pipelines improves your security posture by ensuring agents don't run jobs where a malicious actor has modified the instructions. This moves you towards zero-trust CI/CD by further isolating you from Buildkite itself being compromised.
```

pages/agent/v3/signed_pipelines.md
mbelton-buildkite: Good call. How about this version?

```suggestion
So far, you've configured agents to sign and verify any steps they upload and run. However, you also define steps in a pipeline's settings through the Buildkite dashboard. For example, teams commonly use a single step in the Buildkite dashboard to upload a pipeline definition from [a YAML file in the repository](/docs/pipelines/defining-steps#step-defaults-pipeline-dot-yml-file). These steps should also be signed.
```

pages/agent/v3/signed_pipelines.md
```diff
@@ -0,0 +1,147 @@
...
+```sh
+buildkite-agent tool keygen --help
+```
+
+### 2. Configure the agents
+
+Next, you need to configure your agents to use the keys you generated. On agents that upload pipelines, add the following to the agent's config file:
+
+```ini
+signing-jwks-file=<path to private keyset>
```
mbelton-buildkite, line 99: The linter is picking this up as an error since you're using `keyset` as one word. A quick Google makes me think it's right—you normally see `keySet()` as a method name for a "set of keys." I think we should make it `key set` in each place it's used. Thoughts?

moskyb: yeah, sweet as - will put in a patch

mbelton-buildkite (approved):
Nice work! I appreciate how early you raised the PR for this. It gave us a good amount of time to refine the content. 💚

dabarrell:
@moskyb I'll hold my approval until you make changes on the back of the security review points :)

dabarrell (approved):
Nice work, Benno and others! this reads v well

PR #2528, opened 2023-10-16
Fix typo when replacing go test with gotestsum

sc-lachlanrobertson:
Very minor change to update the `go testsum` command to `gotestsum` (no space). Without it Go will complain about the `testsum` command not being found.

This is correct in the example underneath, so this is just a consistency change. Caught me off-guard, might catch others too.


mbelton-buildkite (approved):
Thanks for catching that and fixing it @sc-lachlanrobertson! Sorry it's taken us so long to get to, but I'll run the linting needed and then merge for you. 😊

PR #2537, opened 2023-10-26
Added `id` to cluster Update / PUT endpoint

james2791:
The `id` (UUID) of a Cluster is needed in the request URL for a PUT to update it using the REST API - missing in the current docs release.

Was noticed by a few folks - and have added it in 👍


mbelton-buildkite (approved):
Good catch! Thanks for fixing that 😊

mbelton-buildkite:
(I'm going to go ahead and merge since it's a clear error we're fixing)

PR #2539, opened 2023-10-29
Improve padding on responsive tables with faux table headings

BrettJay:
This PR fixes an issue occurring where we were removing padding on adjacent `th` and `td` elements when there was a faux  table header in place.

**Desktop:** Before / after
desktop @2x

When reviewing the tablet/mobile designs, I noticed that the padding between elements was not well resolved; leading to large amounts of negative space and making it a bit difficult to infer the proximity and relationship between elements. These updates lead to a more comfortable reading experience, with greater information density.

**Tablet** Before / after
tablet @2x

**Mobile** Before / after
mobile @2x

### DIFF
--- a/app/frontend/css/components/_responsive_table.scss
+++ b/app/frontend/css/components/_responsive_table.scss
@@ -10,13 +10,13 @@
 
   th {
     overflow-wrap: anywhere;
-    padding: 16px;
+    padding: 4px;
   }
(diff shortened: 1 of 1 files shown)


dannymidnight (approved):
Nice. Sorry, I totally missed this 3 week old PR!

PR #2543, opened 2023-11-02
Ruby v3.2.2

dannymidnight:
Bump to ruby `v3.2.2`


Gemfile.lock
```diff
@@ -106,7 +106,7 @@ GEM
       rake (>= 12.2)
       thor (~> 1.0)
     rake (13.0.6)
-    redcarpet (3.5.1)
+    redcarpet (3.6.0)
```
dannymidnight, line 109: I noticed a warning coming through so ended up also bumping this gem for latest patch.

PR #2544, opened 2023-11-02
Mount migration tool

dannymidnight:
I had a think about https://github.com/buildkite/buildkite/pull/14388 and I thought perhaps our docs is good place for hosting the migration tool. 

This PR renders a hosted version of the https://github.com/buildkite/migration under `/docs/migrate`. 

It's a _little_ hacky using a submodule, since there's some dependencies required, but it also seems totally fine as an interim solution for giving this tool a home. Long term, I wouldn't expect the API to live in the docs.

I replicated all the markup into docs primarily to control the form endpoint, but it now gives us full control over the styling etc if we want to play around with this.

Screenshot 2023-11-02 at 12 29 18 pm

Thoughts @MelissaKaulfuss @mbelton-buildkite?!

Worth noting that for this PR the static preview will render the page but the transform won't work since that requires a backend.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "vendor/emojis"]
 	path = vendor/emojis
 	url = https://github.com/buildkite/emojis.git
+[submodule "vendor/migration"]
+	path = vendor/migration
+	url = https://github.com/buildkite/migration
--- a/Gemfile
+++ b/Gemfile
@@ -79,3 +79,5 @@ end
 group :test do
   gem "buildkite-test_collector"
 end
+
+gem "parslet"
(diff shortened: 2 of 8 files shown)


dannymidnight:
> Worth noting that for this PR the static preview will render the page but the transform won't work since that requires a backend.

I tell a lie. The preview won't render this route since it's not currently linked anywhere and the static site is generated from being crawled.

mbelton-buildkite:
I think adding this to the docs is a great idea. Thoughts for making it real:

- Add it as a page at `/pipelines/translator` with title "Pipeline translator." We might want to rethink those, but seems like a good place to start. Key thing for me is that it should display in the nav alongside the relevant content.
- Slap a "Beta" label on the page to give us some wiggle room.
- Can we add some some content to the top of the page to explain what someone is looking at? Like somewhere to add normal Markdown or HTML. It would be good to update the design, but adding a disclaimer about what it is and what it can do at the top is probably a good enough place to start.

PR #2549, opened 2023-11-06
Ignore CVE-2023-5717

dannymidnight:
Pretty sure this one is safe to ignore. It affects the Linux Profiler `perf` tool which we don't use.

### DIFF
--- a/.buildkite/pipeline.deploy.yml
+++ b/.buildkite/pipeline.deploy.yml
@@ -34,6 +34,7 @@ steps:
             - CVE-2023-24329 # python3.11 3.11.2-6
             - CVE-2023-3640 # linux 6.1.55-1
             - CVE-2023-45853 # zlib 1:1.2.13.dfsg-1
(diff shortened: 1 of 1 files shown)


yob (approved):
I agree this is very likely fine to ignore.

If we were keen, at some point we could refactor the `Dockerfile` to be multistage, with the gems compiled in one target then  copied across into a production target that doesn't have all the header files (including `linux-libc-dev` which I suspect is what all the `linux` CVE ignores are for).

dannymidnight:
> If we were keen, at some point we could refactor the Dockerfile to be multistage, with the gems compiled in one target then copied across into a production target that doesn't have all the header files (including linux-libc-dev which I suspect is what all the linux CVE ignores are for).

Good call, I'm keen. It was already worth doing for optimised images and now that there's more visibility into CVEs it seems even more worthwhile.

I'll pencil something in the backlog and hit you up when I'm looking to tackling it.

yob:
🎉

PR #2553, opened 2023-11-13
Multi stage build

dannymidnight:
Switch our Dockerfile to use a multi-stage build in an effort to reduce image size, maybe improve builds times, and hopefully reduce the potential number of CVEs. 

Image size has shrunk from ~1.56GB 👉 700MB.

### DIFF
--- a/.buildkite/push-image.sh
+++ b/.buildkite/push-image.sh
@@ -6,8 +6,11 @@ echo "--- :docker: Building docker image"
 
 TAG="${BUILDKITE_BUILD_NUMBER}"
 
-docker build -t "$ECR_REPO:$TAG" --build-arg="DD_RUM_VERSION=$BUILDKITE_BUILD_NUMBER" --build-arg="DD_RUM_ENV=production" .
+docker build -t "$ECR_REPO:$TAG" \
+  --build-arg="RAILS_ENV=production" \
+  --build-arg="DD_RUM_VERSION=$BUILDKITE_BUILD_NUMBER" \
+  --build-arg="DD_RUM_ENV=production" \
+  .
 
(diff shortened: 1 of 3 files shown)


yob (approved):
I tested this locally by building it:

```
$ docker build -t docs-multistage:foo --build-arg="RAILS_ENV=production" --build-arg="DD_RUM_VERSION=123" --build-arg="DD_RUM_ENV=production" .
```
Then booting it:
```
$ docker run -it -p 3000:3000 docs-multistage:foo
```

loading http://127.0.0.1:3000/ in my browser shows the docs site as expected 🎉

I haven't looked at the full diff too closely, but given this is a static site I'm pretty confident this will Just Work.

One possible improvement from a vulnerability scanning perspective would be to add `apt purge --assume-yes linux-libc-dev` in the final layer. It won't save space, but it will avoid a buncha useless CVE  warnings. I reckon that could come later though, I'm curious to see how this goes first.

/cc @steveh and @mattborden who might find this interesting

yob:
Does the deploy-preview step require us to include curl and jq in the final layer? Not really a bug deal if so

dannymidnight:
> Does the deploy-preview step require us to include curl and jq in the final layer? Not really a bug deal if so

Yeah it does. I'm just now taking a look at adding a separate layer to target that step.

dannymidnight:
> One possible improvement from a vulnerability scanning perspective would be to add apt purge --assume-yes linux-libc-dev in the final layer. It won't save space, but it will avoid a buncha useless CVE warnings.

Nice one. I'm definitely keen to wipe out CVE warnings so for sure I'll follow up with this.

mbelton-buildkite (approved):

PR #2556, opened 2023-11-15
Document run `state` and `result`

sj26:
Add brief docs for the properties introduced in https://github.com/buildkite/buildkite/pull/14616.


pages/apis/rest_api/analytics/runs.md
```diff
@@ -45,3 +49,7 @@ curl "https://api.buildkite.com/v2/analytics/organizations/{org.slug}/suites/{su
 Required scope: `read_suites`
 
 Success response: `200 OK`
+
+Runs are created with a `state` of `running` and proceed to `finished` when all uploads have been processed. The run may return to `running` if additional results are uploaded.
+
+`result` starts as `pending` and may proceed to `passed` or `failed` once finished. If a run does not complete within a reasonable time period it may also proceed to `stale`. `result` may change back to `pending` or change between `passed` and `failed` if additional results are uploaded.
```
niceking, line 55:
```suggestion
Run `result` starts as `pending` and will proceed to `passed` or `failed` when at least one test result has been processed.  The presence of a `passed` or `failed` result does not indicate that the run has finished processing. `result` may change from `passed` to `failed` if additional results are uploaded. The `result` is `failed` when there is at least one failing test in the run, and it is not possible for `result` to change from `failed`.
```

pages/apis/rest_api/analytics/runs.md
```diff
@@ -45,3 +49,7 @@ curl "https://api.buildkite.com/v2/analytics/organizations/{org.slug}/suites/{su
 Required scope: `read_suites`
 
 Success response: `200 OK`
+
+Runs are created with a `state` of `running` and proceed to `finished` when all uploads have been processed. The run may return to `running` if additional results are uploaded.
```
niceking, line 53:
```suggestion
Runs are created with a `state` of `running` and proceed to `finished` when all uploads have been processed. The run may return to `running` if additional results are uploaded. If a run receives no results within a reasonable time period its `state` will proceed to `stale`.
```

niceking: FYI our default scope on runs excludes runs that are `stale`, so I think the behaviour might be that the API will returning the `running` run if requested, but once the run becomes `stale` the same API request will probably return a not found or similar?

niceking: This would happen now with the API as well so I'm not suggesting a change in behaviour, I was just wondering maybe if we shouldn't mention the `stale` state at all because it wouldn't ever be returned to the customer in the API

sj26: I don't think this is correct. `result` can become `stale`, but `state` can only be `running` or `finished`.

niceking: Oh, I totally thought `stale` was a `state` and not a `result`, but yeah that's not what the code says!

niceking (approved):

PR #2557, opened 2023-11-15
Add detailed instructions for plugin git repositories

deniz-lktr:
When developing a plugin, we can either use a public git repository or a private one. We can also use any git host.

Clarify how this is possible by giving an absolute URL and the required format.

Thanks to Tom Watt for the pointer, this commit should help others developing plugins and hitting similar pitfalls.

### DIFF
--- a/pages/plugins/writing.md
+++ b/pages/plugins/writing.md
@@ -251,6 +251,17 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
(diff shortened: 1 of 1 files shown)


mbelton-buildkite:
Hi @deniz-lktr, thanks for raising this! It looks like this information is covered under [Plugin sources on the Using plugins page](https://buildkite.com/docs/plugins/using#plugin-sources). Do you think that's enough information, or is there more you think we should include there? 😊

PR #2559, opened 2023-11-16
Update cli reference for agent v3.59.0

triarius:
The `tool` subcommands were released in the last few minor versions, but they had not been documented yet. We may move these sub commands out of the agent in the future. If so, we will just update the docs with where they have been moved to.

So there is a scary "Experimental" warning around them that will likely up for a while.


DrJosh9000 (approved):
☘️

PR #2560, opened 2023-11-16
Add example of GraphQL cursor-based pagination

tomowatt:
Based on feedback received and a lack of an example where it mentions utilizing Pagination


mbelton-buildkite (commented):
This is good information to include, but I feel like it's not in the right spot. The page it's on talks about limits, and while this is a technique for dealing with limits, it's useful information across the GraphQL API in general.

I think it would be good to move it to the Overview page, and rework the [Relay compatibility](https://buildkite.com/docs/apis/graphql-api#relay-compatibility) section to be the "Pagination" section. It would mention complying with the Relay specification, then use your content here as an example. We could then link to it from this page where it mentions "pagination."

What do you think? I'm happy to help make it work there if you need. 😊

PR #2582, opened 2023-12-05
Noted 70 character limit for adding a custom reason string for `skip` step attributes

james2791:
A limit of 70 characters is present when passing in a reason for a `skip` attribute of a command step: this PR adds this as a small addendum to the command steps' optional attribute row in said table 🙂


pages/pipelines/command_step.md
```diff
@@ -178,7 +178,7 @@ _Optional attributes:_
   <tr>
     <td><code>skip</code></td>
     <td>
-      Whether to skip this step or not. Passing a string provides a reason for skipping this command. Passing an empty string is equivalent to <code>false</code>.
+      Whether to skip this step or not. Passing a string (with a 70 character limit) provides a reason for skipping this command. Passing an empty string is equivalent to <code>false</code>.
```
mbelton-buildkite, line 181: Missing dash:

```suggestion
      Whether to skip this step or not. Passing a string (with a 70-character limit) provides a reason for skipping this command. Passing an empty string is equivalent to <code>false</code>.
```

mbelton-buildkite (approved):
One minor nitpick, then you're all good :)

PR #2589, opened 2023-12-11
Add signal reason to job finished event

moskyb:
See https://github.com/buildkite/buildkite/pull/14888.

one annoying thing about these examples is that they don't have an example of when the signal reason is semantically useful - ie, if a job failed, it have a signal reason `agent_refused`, which is much more useful to put in there than an empty string, but if we were to add a proper signal reason in there, it wouldn't make semantic sense. happy to do either.

### DIFF
--- a/pages/integrations/amazon_eventbridge.md
+++ b/pages/integrations/amazon_eventbridge.md
@@ -429,6 +429,7 @@ AWS EventBridge has strict limits on the size of the payload as documented in [A
         "queue=default"
       ],
       "exit_status": 0,
+      "signal_reason": "",
       "passed": true,
       "soft_failed": false,
       "state": "finished",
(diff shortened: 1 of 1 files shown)


gilesgas:
Hi @moskyb , thanks very much for this docs update and PR.

Is the issue here that when a job finishes successfully, the `"signal_reason":` value is empty (i.e. `""`) but when a job fails, this field's value would have a value?

If so, then perhaps we can add a note in the docs underneath this example to explain what the different values for this field could be?

What would all the different types of values be (or at least a subset of the most important ones)?

gilesgas (changes requested):
Hi @moskyb , please answer the queries in my previous comment before this PR can be approved and merged in.

gilesgas:
Thanks very much for the update @moskyb , that's very helpful!

I've tweaked the wording a tad, such that it now looks like this:

Screenshot 2024-04-23 at 3 19 19 PM

Hope this is still OK?

gilesgas (approved):

PR #2590, opened 2023-12-12
Adding Org level job expiry for Scheduled jobs

ozdenbuildkite:
These are the doc changes for us to add:

- expired into job states
- Explain the difference between timed out & expired
- Explain how you can enforce org level settings on expired. Global limit is 30 days, but they can enforce a lower limit
- Add a page called build timeouts


mbelton-buildkite:
👀 Looking at this now and plan to push style updates directly to the branch to move quickly. Please ping me if that's an issue or if you're still making changes that might cause merge conflicts.

Also, additional context: `expired` is not a new job state -> https://buildkite.com/changelog/157-automatic-job-expiration-after-30-days

pages/pipelines/build_timeouts.md
```diff
@@ -0,0 +1,38 @@
+# Build timeouts
+
+Build timeouts are limits on the maximum time a job can wait before being picked up by an agent. If a job exceeds the time limit, the job is automatically canceled and the build fails.
```
mbelton-buildkite, line 3: Please check this. I think the page should start by saying what the thing is, so I added this part.

pages/pipelines/build_timeouts.md
```diff
@@ -0,0 +1,38 @@
+# Build timeouts
```
mbelton-buildkite, line 1: Should this actually be called "Job timeouts?" When going through the content, it seems to be about timeouts on jobs rather than builds.

pages/pipelines/defining_steps.md
```diff
@@ -143,6 +143,13 @@ Differentiating between `broken`, `skipped` and `canceled` states:
 * This is distinct from `skipped` jobs, which might happen if a newer build is started and [build skipping](/docs/apis/rest-api/pipelines#create-a-yaml-pipeline) is enabled. Broadly, jobs break because of something inside the build, and are skipped by something outside the build.
 * Jobs can be `canceled` intentionally, either using the Buildkite UI or one of the APIs.
 
+Differentiating between `timing_out`, `timed_out`, and `expired` states:
+
+* Jobs become `timing_out`, `timed_out` when a job starts running on an agent but doesn't complete within the timeout period.
+* Jobs become `expired` when they reach the scheduled job expiry timeout before being picked up by an agent.
```
mbelton-buildkite, line 149: Please check this, too. I tried to make them a bit clearer.

pages/pipelines/build_timeouts.md
ozdenbuildkite: I found myself asking the same question with you @mbelton-buildkite . It has always been called build timeouts so I assumed that was a better name. 

I can also see why they call it build timeouts also actually as I write this. The end goal is for the build to not be running and reach a terminal state that is finished/cancelled/passed. So although the timeouts are in job level they eventually let the builds get to a terminal state. 

I also worry if job timeouts will confuse customers because of job/step confusion + their end goal of achieving build timeouts. I would rather keep the title but have something at the beginning explaining why like what you have done :)

mbelton-buildkite: 👍🏻 Cool, glad I wasn't alone in that thought haha. Happy to leave it as-is.

mbelton-buildkite (approved):
Looks good to me :)

PR #2594, opened 2023-12-14
Pass Buildkite Env Variables to Rspec steps

nprizal:
### Description
We noticed that some env variables are missing from Rspec runs in [Docs TA suite](https://buildkite.com/organizations/buildkite/analytics/suites/docs-1?branch=main). 
Screenshot 2023-12-15 at 9 20 57 AM

This PR passes Buildkite environment variables from CI env to Docker container that runs Rspec, so the Test Collector can send them to Test Analytics. ~~I exclude `BUILDKITE_BUILD_URL` to prevent linking to the private pipelines.~~

### Verification
To verify, we can check if the Rspec run has all information in [Docs TA suite](https://buildkite.com/organizations/buildkite/analytics/suites/docs-1?branch=main)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -15,6 +15,12 @@ steps:
             - CI
             - RAILS_ENV
             - BUILDKITE_ANALYTICS_TOKEN
+            - BUILDKITE_BRANCH
+            - BUILDKITE_BUILD_ID
+            - BUILDKITE_BUILD_NUMBER
+            - BUILDKITE_BUILD_URL
+            - BUILDKITE_COMMIT
+            - BUILDKITE_MESSAGE
 
   - label: ":docker: Build image"
     depends_on: "linting"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -15,6 +15,7 @@ steps:
             - BUILDKITE_BRANCH
             - BUILDKITE_BUILD_ID
             - BUILDKITE_BUILD_NUMBER
+            - BUILDKITE_BUILD_URL
             - BUILDKITE_COMMIT
             - BUILDKITE_JOB_ID
             - BUILDKITE_MESSAGE
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -18,7 +18,6 @@ steps:
             - BUILDKITE_BRANCH
             - BUILDKITE_BUILD_ID
             - BUILDKITE_BUILD_NUMBER
-            - BUILDKITE_BUILD_URL
             - BUILDKITE_COMMIT
             - BUILDKITE_MESSAGE
 
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,15 +11,15 @@ steps:
           run: app
           dependencies: false
           env:
-            - BUILDKITE_JOB_ID
-            - CI
-            - RAILS_ENV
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -15,6 +15,11 @@ steps:
             - CI
             - RAILS_ENV
             - BUILDKITE_ANALYTICS_TOKEN
+            - BUILDKITE_BRANCH
+            - BUILDKITE_BUILD_ID
+            - BUILDKITE_BUILD_NUMBER
+            - BUILDKITE_COMMIT
+            - BUILDKITE_MESSAGE
```
niceking, line 22: I think you're missing BUILDKITE_BUILD_URL and BUILDKITE_JOB_ID and then everything should work :)

niceking: It might also be a good idea to alphabetise this list, I thought it was missing BUILDKITE_JOB_ID as well but its actually just at the top

nprizal: @niceking I deliberately excluded `BUILDKITE_BUILD_URL` because the run data is sent from `Docs (main)` which is a private repo. Do you think that it is safe to expose the build URL in public test suite? It should be inaccessible, but I thought that it is more secure if we don't expose the URL?

niceking: The BUILDKITE_BUILD_URL is just the build url tho, e.g. `BUILDKITE_BUILD_URL="https://buildkite.com/buildkite/docs/builds/6991"`

And docs is a public pipeline, so its just public test suite linking to public pipeline?

niceking: Oh, are you saying that the Test Analytics builds are for a private docs pipeline?

niceking: Ohh I see https://buildkite.com/buildkite/docs-main 

Okay, that's kinda weird that we've got a public Test Suite for Docs and a public Pipeline for docs, but they aren't linked up together? Why do we only have a Test Suite for the private docs pipeline? This seems like something we should fix

nprizal: We discussed this in slack. Currently, the Rspec runs inside Docs **public** pipeline's build are not being reported to Test Analytics. The Rspec runs being reported to Test Analytics are triggered from Docs **private** pipeline's build. However, since there is no sensitive information in the build URL, we can send the URL to Test Analytics public suite.

We also discussed that initially we should report Rspec runs from Docs public pipeline's build as well.

niceking (approved):

pda (approved):

mbelton-buildkite (approved):
Looks good to me if looks good to @dannymidnight (he has more to do with the docs pipeline than I do)

.buildkite/pipeline.yml
dannymidnight: took me a moment to digest that, but makes sense to me 😄 

If there's no reason to exclude `BUILDKITE_BUILD_URL`, should we use the following option from the docker plugin? 

https://github.com/buildkite-plugins/docker-buildkite-plugin?tab=readme-ov-file#propagate-environment-optional-boolean

dannymidnight (approved):

.buildkite/pipeline.yml
nprizal: That's a good point. Although there should be no risk of passing all `BUILDKITE` env vars to the docker in our private pipelines, I'd say that it is a good practice to pass only necessary information.

PR #2604, opened 2024-01-07
Removing beta badges from TA collectors

meghan-kradolfer:
Our supported languages for collectors have several "beta" badges. These were released a year and a half ago and should not be considered beta. 

This PR removes the 'beta' badges from the left hand nav and within the collector page headings for the following collectors:
- Swift
- Android
- .NET
- Rust


nprizal (approved):
Looks good to me

mbelton-buildkite (approved):
Nice :)

PR #2599, opened 2023-12-21
feat: add organization member REST API documentation

jfhinchcliffe:
ℹ️ Don't merge yet. This is documentation for some feature-flagged work which is yet to be released on Buildkite

We've recently done some work to uplift the REST API - exposing organization membera (PR: https://github.com/buildkite/buildkite/pull/15036)

This PR adds documentation for that feature and should only be merged once it has been released on Buildkite.

PR #2598, opened 2023-12-20
feat: add documentation for Team Pipelines REST functionality

jfhinchcliffe:
ℹ️  Don't merge yet. This is documentation for some feature-flagged work which is yet to be released on Buildkite

We've recently done some work to uplift the REST API - adding team pipeline assignments (PR: https://github.com/buildkite/buildkite/pull/14941)

This PR adds documentation for that feature and should only be merged once it has been released on Buildkite.

PR #2612, opened 2024-01-11
[FDN-3587] Add new Orgs / Teams / Suites REST API Documentation

jfhinchcliffe:
Consolidates all the recent teams API additions into a single PR.

🚧 Should not be merged until we've enabled the API changes on the main app.

@mbelton-buildkite this should contain all the feedback that that you kindly provided on the https://github.com/buildkite/docs/pull/2598 and https://github.com/buildkite/docs/pull/2599 PRs. I just found it easier to put all this related work in a single PR as they'll most likely be released on the main site at the same time and it saves on merge conflicts on the nav changes etc.

As part of this I pulled 'Organization' out into its own nav item under APIS / REST, as these new endpoints are nested under the organization's / teams route in our API. eg:
```
curl --request GET \
  --url https://api.buildkite.com/v2/organizations/hinchys-fondue-factory/teams/c5e09619-8648-4896-a936-9d0b8b7b3fe9/suites \
  --header 'Authorization: Bearer TOKEN'
```

Screenshot 2024-01-18 at 3 51 29 pm

❓ We could think about adding another submenu under organization for Teams

Each commit represents another resource added to the API - would be simplest to review commit-by-commit
- [x] Team Pipelines
- [x] Team Members
- [x] Team Suites
- [x] Org Members
- [x] Org Teams

Linear ticket:
https://linear.app/buildkite/issue/FDN-3587/add-docs-pr-for-new-rest-api-functionality


jameshill:
@jfhinchcliffe could you give this a rebase ❤️

jfhinchcliffe:
@jameshill rebased!
Screenshot 2024-03-05 at 9 32 18 am

jfhinchcliffe:
Just re-testing / verifying the teams endpoint:
Teams  Index ✅ 
Screenshot 2024-03-05 at 10 00 03 am

Team Show ✅ 
Screenshot 2024-03-05 at 10 00 44 am

Team Create ✅ 
Screenshot 2024-03-05 at 10 03 00 am

Team Update ✅ 
Screenshot 2024-03-05 at 10 03 45 am

Team Destroy ✅ 
Screenshot 2024-03-05 at 10 04 35 am

Team Members ✅ 
Screenshot 2024-03-05 at 10 01 33 am

TBC

gilesgas:
Thanks very much for all this work @jfhinchcliffe !

This is looking really good!

I've noticed a small issue with the left-hand nav—most likely caused by a combination of the repeated usage of the name and component in the URL slug:

nav-problem-multiple-items-selected

I think I've got an idea of how to resolve this but would like to run this by @mbelton-buildkite first

data/nav.yml
```diff
@@ -429,7 +429,30 @@
...
+            - name: "Runs"
+              path: "apis/rest-api/analytics/runs"
+            - name: "Suites"
+              path: "apis/rest-api/analytics/suites"
+            - name: "Tests"
+              path: "apis/rest-api/analytics/tests"
+        - name: "Organizations"
+          children:
+            - name: "Organizations"
+              path: "apis/rest-api/organizations/organizations"
```
gilesgas, line 451: I suspect using this repeated URL slug pattern might be causing part of the problem I described in my first comment of this PR:
`path: "apis/rest-api/organizations/organizations"`

It's been used a few other times in this PR.

app/models/beta_pages.rb
```diff
@@ -1,7 +1,16 @@
 class BetaPages
   def self.all
     [
-      'pipelines/cluster-queue-metrics'
+      'agent/clusters',
```
gilesgas, line 4: Should the clusters API be marked "beta".

Clusters went GA recently (see #2642 for more context).

app/models/beta_pages.rb
```diff
@@ -1,7 +1,16 @@
...
+      'apis/rest-api/clusters',
```
gilesgas, line 5: Ditto about clusters going GA

app/models/beta_pages.rb
```diff
@@ -1,7 +1,16 @@
...
+      'apis/rest-api/pipeline-templates',
+      'apis/rest-api/team-pipelines',
+      'pipelines/cluster-queue-metrics',
```
gilesgas, line 8: This one should stay "beta".

data/nav.yml
```diff
@@ -429,7 +429,30 @@
...
+              path: "apis/rest-api/analytics/flaky-tests"
+            - name: "Runs"
+              path: "apis/rest-api/analytics/runs"
+            - name: "Suites"
+              path: "apis/rest-api/analytics/suites"
+            - name: "Tests"
+              path: "apis/rest-api/analytics/tests"
+        - name: "Organizations"
+          children:
+            - name: "Organizations"
```
gilesgas, line 450: Also as part of addressing the problem described in my first comment, consider changing this to "Overview" to avoid the "Organizations > Organizations" issue in the page nav.

gilesgas:
Since we are substantially re-organising the page nav as part of this PR, we should also consider the following:

- Where should the [Clusters REST API endpoint docs](https://buildkite.com/docs/apis/rest-api/clusters) live? (Within the new Organizations page hierarchy or keep it in Pipelines)
-

app/models/beta_pages.rb
```diff
@@ -1,7 +1,16 @@
 class BetaPages
   def self.all
     [
-      'pipelines/cluster-queue-metrics'
+      'agent/clusters',
+      'apis/rest-api/clusters',
+      'apis/rest-api/pipeline-templates',
```
gilesgas, line 6: Not beta either.

data/nav.yml
```diff
@@ -429,8 +429,33 @@
...
+          children:
+            - name: "Overview"
+              path: "apis/rest-api/organizations"
+            - name: "Members"
+              path: "apis/rest-api/organizations/members"
+              pill: "beta"
+        - name: "Pipelines "
+          children:
+            - name: "Overview"
+              path: "apis/rest-api/pipelines"
```
gilesgas, line 458: I needed to add a space after the word Pipelines here (i.e. `"Pipelines "`) to prevent the **Pipelines** section in the 'global' nav from being highlighted too when viewing its Overview page.

This seems to be a bit of a bug in our documentation site, whereby this kind of item structure within this YAML file:

```
        - name: "Something specific"
          ...
            - name: "Something else"
              ...
```

Would could cause any element that's also called "Something specific" elsewhere on the page to become highlighted too.

Here's an example of what I mean:

multiple-elements-being-highlighted

@dannymidnight , would you know if there's anyway to fix this issue with our docs site?

gilesgas:
Hi @jfhinchcliffe ,

I've pushed through a few changes to the docs, which I can summarise as follows:

- Changed the names of the nested 'Organizations' and 'Pipelines' pages to 'Overview' to avoid the minor docs site highlighting bug we discussed before (and described above).
- Added intros to the tops of these two (now-named) 'Overview' pages to explain the new page re-organisation in the docs.
- Made all the calls consistent (and added header for `$TOKEN` to indicate which calls require access token authentication). I started noticing too many different ways of representing curl calls throughout the docs, which is not great.

pages/apis/rest_api/teams.md
```diff
@@ -57,3 +61,164 @@ Optional [query string parameters](/docs/api#query-string-parameters):
...
+</tbody>
+</table>
+
+## Delete a team
+
+Remove a team.
+
+```bash
+curl -H "Authorization: Bearer $TOKEN" \
+  -X DELETE "https://api.buildkite.com/v2/organizations/{org.slug}/teams/{team.uuid}"
```
gilesgas, line 211: On a few of these `DELETE` calls, I removed the `-H "Content-Type: application/json" \` options as I strongly suspect these are not required (by virtue of no JSON payload being required in the request).

gilesgas (approved):

PR #2617, opened 2024-01-17
Update GraphQL cookbook with Get Job info by ID

lizrabuya:
We have customers asking how to get job info if they only have a job UUID. One scenario is when debugging agent logs, usually only job UUIDs are logged, without information of the pipeline and build number. Since a few customers have asked this, I think it would be helpful to update our cookbook with an example on how this is done.


gilesgas:
Hi @lizrabuya ,
Thanks for this PR!
Just as I mentioned previously, I have a couple of tweaks to make to this PR. Please let me know if you'd be OK for me to push them up to this branch before merging this in to `main`.
These updates are more stylistic changes rather than any changes to your content.

gilesgas:
I decided to use 'UUID' instead of 'ID' in the description since (from the GraphQL query - and correct me if I'm wrong), it looks like jobs have both an ID and UUID, and this query involves getting a job's info by its UUID only.

gilesgas (approved):
All good for to merge this now @lizrabuya :-)
Thanks for this contribution!

PR #2618, opened 2024-01-18
[CORE-6] Specify that agent API is internal

dabarrell:
Explicitly state that the non-metrics endpoints in the agent API are not public endpoints, and that we don't guarantee stability. This is implicit in `The only publicly available endpoint is /metrics.`, but this makes it clearer.


SorchaAbel (approved):
perfect!

dabarrell:
CleanShot 2024-01-18 at 17 00 05

dabarrell:
👋 @mbelton-buildkite small one here - feel free to hit merge when (if) you approve

mbelton-buildkite (approved):

PR #2620, opened 2024-01-18
Concurrency group issues with block/input steps

pzeballos:
There's a known issue when using concurrency group with block/input steps.

The time-ordering of jobs that reference a concurrency is stable (roughly: the build creation time, and then the jobs ordering within that build). But only active jobs (roughly: running or eligible-to-run) are part of the queue. Jobs waiting for a block/input step aren’t active, so they’re not active in the queue, and so the concurrency group will let jobs from a newer build run. But then when you unblock that block/input step, those jobs become active in the queue, and potentially jump in front of the newer build, breaking the concurrency group ordering.


gilesgas:
Quick question, does the `concurrency` attribute specify the concurrency limit?

pzeballos:
> Quick question, does the `concurrency` attribute defined in a step specify the concurrency limit for that step?

You need to specify it. For example `concurrency: 1` will set the limit to 1. If you don't put anything, it's like doing nothing :)

pages/pipelines/controlling_concurrency.md
```diff
@@ -43,6 +43,9 @@ For example, if you have two pipelines that each deploy to a different target bu
 
 Concurrency groups guarantee that jobs will be run in the order that they were created in. Jobs inherit the creation time of their parent. Parents of jobs can be either a build or a pipeline upload job. As pipeline uploads add more jobs to the build after it has started, the jobs that they add will inherit the creation time of the pipeline upload rather than the build.
 
+>🚧 <code>concurrency_group</code> and <code>block</code>/<code>input</code> steps
+> Note that both [block](/docs/pipelines/block-step) and [input](/docs/pipelines/input-step) steps cause the steps to be uploaded and scheduled at the same time, which breaks concurrency groups: it blocks jobs being added to the concurrency group, but does not adjust ordering once unblocked. The concurrency won't be added to the queue until the block step is unblocked, and once it does, the timestamp will be from the pipeline upload step.
```
gilesgas, line 47: Hi @pzeballos , since this subject matter is still a tad new to me, I just a few queries I'd like to make to help clarify this note...

When you say:
> cause the _steps_ to be uploaded and scheduled at the same time

Do these _steps_ refer to only 'block' and 'input' steps and if so, is this only 'block' and 'input' steps in a single pipeline or those defined across many pipelines (which are defined as part of the same concurrency group)? Alternatively, do these _steps_ refer to any type of step (in a single pipeline or many pipelines), which are defined as part of the same concurrency group?

pzeballos: Sorry @gilesgas! We are still having customers raising this issue os having the note in the docs should help a but :) 

The issue is with both, block and input steps. Concurrency group groups work across pipelines; what it matters is the name of the queue (ie, having the same concurrency attribute)

gilesgas: Thanks @pzeballos , I can now revisit this PR and should be able to have it merged in soon.

To speed things along, I'll push some updates directly to this PR myself...

pages/pipelines/controlling_concurrency.md
```diff
@@ -42,6 +42,9 @@ For example, if you have two pipelines that each deploy to a different target bu
 
 Concurrency groups guarantee that jobs will be run in the order that they were created in. Jobs inherit the creation time of their parent. Parents of jobs can be either a build or a pipeline upload job. As pipeline uploads add more jobs to the build after it has started, the jobs that they add will inherit the creation time of the pipeline upload rather than the build.
 
+> 🚧 `concurrency_group` and `block` / `input` steps
+> Note that both [block](/docs/pipelines/block-step) and [input](/docs/pipelines/input-step) steps cause the steps to be uploaded and scheduled at the same time, which breaks concurrency groups. These two steps block jobs being added to the concurrency group, but do not adjust ordering once unblocked. The concurrency won't be added to the queue until the block step is unblocked, and once it does, the timestamp will be from the pipeline upload step.
```
gilesgas, line 46: @pzeballos , I've revised this sentence a tad. Please let me know if my revisions are correct. Otherwise, please clarify what's incorrect.

Also, I have a few more queries about the existing text additions in this PR:

> but do not adjust ordering once unblocked.

Do you mean that jobs will still run in the order they were created in, or does using `block` and `input` steps break this ordering too?

> The concurrency won't be added to the queue

Do you mean the _concurrency group_? If not, what do you mean by 'concurrency' on its own?

> until the block step is unblocked

Is this _only_ a `block` step, or does this cover both `block` and `input` steps?

Sorry to be pedantic here, but I'm aiming to make the docs as unambiguous as possible.

pzeballos: Thanks @gilesgas!

> Do you mean that jobs will still run in the order they were created in, or does using block and input steps break this ordering too?
They will run in the order they were created.

> Do you mean the concurrency group? If not, what do you mean by 'concurrency' on its own?
The concurrency group

> Is this only a block step, or does this cover both block and input steps?
Both :)

gilesgas: Wonderful! Thanks for these updates Paula :-)

gilesgas (approved):

PR #2623, opened 2024-01-22
Fix a typo in SSO with okta doc

JuanitoFatas:
The session is called Credentials Details with `s`.

CleanShot 2024-01-22 at 11 32 06@2x


mbelton-buildkite (approved):
💚

PR #2631, opened 2024-01-29
Ignore CVE-2024-0567

mbelton-buildkite:
See [buildkite/buildkite](https://github.com/buildkite/buildkite/pull/15366). This looks to be how we've handled this before.

FYI @dannymidnight when you're back from leave.


yob (approved):
Yup, makes sense to ignore this one. I can't see any way it's exploitable in a docs context :+1:

PR #2643, opened 2024-02-05
Remove 'beta' labels from clusters pages and docs.

gilesgas:
During a chat with me last Friday, @LiamPClancy mentioned (and please correct me if I'm wrong) that he'd like the 'beta' label removed from our clusters docs for the post-beta soft launch of clusters.


app/models/beta_pages.rb
```diff
@@ -1,8 +1,6 @@
 class BetaPages
   def self.all
     [
-      'agent/clusters',
```
gilesgas, line 4: Just an FYI, this page/path doesn't/no longer seems to exist and so I'm removing it as part of this PR.

mbelton-buildkite (approved):
Looks good to me. You might also consider removing the [beta label from Pipeline templates](https://github.com/buildkite/docs/pull/2643/files#diff-563edf1b408959f1be664c0e40280fed93adc3959e4cdb2254e900a27b71cd04L442) as Oz mentioned (or do that in a separate PR):

```
- name: "Pipeline templates"
              path: "apis/rest-api/pipeline-templates"
              pill: "beta"
```

gilesgas:
Thanks @mbelton-buildkite , just had a brief chat with @LiamPClancy and it looks like we'll be removing the beta labels from the clusters documentation around the time that the changelog entry announcing the clusters GA release (scheduled for about a week later) is made. Liam will confirm this after chatting with Oz.

gilesgas:
Hi @ozdenbuildkite and @LiamPClancy ,

As discussed earlier today, this PR also removes the 'beta' tags from the 'pipeline templates' feature, as well as adding the 'beta' tag to the page title itself "**Queue metrics in clusters**".

If you were only intending the 'beta' tag to appear next to the page heading "**Advanced Queue Metrics**" further down the page, then unfortunately I don't this functionality is built. Looks like 'beta' labels can only be added to the page heading at the top of the page.

PR #2652, opened 2024-02-15
Temporarily suppress bullet point indentation checking.

gilesgas:
Other than not being able to spell the word 'suppress' in branch names correctly, I'd like to make this suggested change to the Markdown linting rules around list indentation.

First, when building the docs site locally, these linting rule checks don't appear to run. I presume this is by design, to improve the performance of the local docs build environment.

When changes are pushed up to GitHub, Buidlkite runs a Markdown linter that checks to ensure that unordered list items (bullet points) are strictly indented by only 2 spaces. This prevents us from leveraging the full benefits of Markdown's HTML rendering in the docs.

For example, forcing 2 spaces, I'm unable to bring the code blocks in line with my 2nd level bullet point:

forced 2 spaces

By indenting bullet points by 4 spaces, I'm able to get a 3rd level of indented bullet points (a very useful feature in docs), as well as bring code blocks in line with the bullet points they're meant to be associated with:

4 spaces allowed

Since we have a lot of bullet points currently indented by 2 spaces, I'm commenting out the lint checking for indentation. Once I've had the chance to systematically indent all other bullet points in the docs by 4-space increments (for each level), uncomment this line and change it to 4.

> [!NOTE]
> Also, it appears 'markdownlint's own documentation recommends a minimum indentation level of 3:
https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md007---unordered-list-indentation

As an interim measure to encourage new contributors to use 4-space indentation increments, I'm updating the Markdown style guide guidance on this (since no checking would be done in the meantime).

PR #2654, opened 2024-02-15
Run prettier over markdown

dannymidnight:
Following on from https://github.com/buildkite/docs/pull/2652 I thought it might be nice to run Prettier over all the files to adjust indentation in one big bang.

Turns out there's an outstanding issue in Prettier that needs to be resolved before we can drop this in otherwise we keep some extra indentation. 

- https://github.com/prettier/prettier/issues/5019
- https://github.com/prettier/prettier/pull/15526


gilesgas:
Thanks @dannymidnight for the catch-up to better explain the purpose behind this PR.

While I haven't had the chance to switch across to Prettier as an IDE plugin (to assist with developing content for the Buildkite Docs yet), I've tweaked the existing plugin I use (Markdown All in One) to check my syntax before submitting updates - in particular, the my IDE now points out additional spaces at the ends of lines.

A couple of things to note about contributing to these Markdown docs which might be considered different from practices employed when contributing to Markdown files more closely associated with actual code (e.g. READMEs):

- When implementing numbered lists, each numbered list item should begin with just `1.` (and not be numbered sequentially). Numbering sequentially results in the need to re-number downstream list items (when additional steps need to be inserted part-way through), and also impairs the ability to review actual changes in PR diffs - i.e. every downstream list item from the inserted list item would indicate unnecessary changes too.
- For bulleted lists, we don't want any additional spaces added between the bullet list indicator (e.g. `-`) and the text that follows. This impairs the readability of the source. I'm not sure if this is (still) a Prettier bug or this can be tweaked in the UI.

At some point in the near future, I'd like to prepare a proper 'contributors guide', which would provide instructions on how to configure an IDE plugin (like Prettier or Markdown All in One) to check/reformat the docs such that they'd build correctly once pushed up.

PR #2655, opened 2024-02-15
Add example of using GraphQL to create Builds on Pipelines

tomowatt:
There is an example of the using the REST API for creating Builds but no examples using the GraphQL API


gilesgas (approved):
LGTM!

Once the builds finish successfully, this can be merged. Thanks @tomowatt !

PR #2656, opened 2024-02-15
Add information on agent panel metrics

mcncl:
Currently the docus don't inform that there is potential variation in the numbers shown in the agent panel and those on the graph below.

I've added an (i) box to inform the user of this.


gilesgas (approved):
Thanks for your contribution @mcncl , happy for this to be merged in.

PR #2658, opened 2024-02-18
changed the agent heartbeat timeout in the logs to accuratly reflect …

LiamPClancy:
…the setting that is being used in the code


gilesgas (approved):
Thanks for the update @LiamPClancy , approved!

PR #2642, opened 2024-02-05
Clusters GA updates to the docs

gilesgas:
This PR is for the required docs changes around the imminent generally available (GA) release of clusters in Buildkite.

Clusters has been in beta for a while and now it will soon be officially launched and become GA.

> [!WARNING]
> Do not merge this PR until clusters is generally available (GA) to all customers!

The pages which have been updated in the docs include the following:

- [Clusters overview](https://2642--bk-docs-preview.netlify.app/docs/clusters/overview):
    - Thanks to Ben C, updated the diagram near the top of this page to group agents by queue.
    - Removed much of the content around how to enable clusters and the note at the top about the clusters GA release date, since this will no longer be relevant.
    - General wording tweaks throughout this page.

- [Manage clusters](https://2642--bk-docs-preview.netlify.app/docs/clusters/manage-clusters) - there have been substantial rewrites to the following existing sections (i.e. headings) of this page (indicated by 'formerly called'):
    - **Setting up clusters** - explains a little more detail on how clusters would typically be set up.
    - **Create a new cluster** (formerly called **Create a cluster**) - now documents processes for the Buildkite interface and REST API
    - **Connect agents to cluster** - most of these processes are documented on the revised [Agent tokens](https://2642--bk-docs-preview.netlify.app/docs/agent/v3/tokens) and [Queues](https://2642--bk-docs-preview.netlify.app/docs/agent/v3/queues) pages in the 'Pipelines > Agents' area of the docs.
    - **Moving unclustered agents to a cluster** (formerly called **Migrate to clusters**, whose content was quite minimal)
    - **Restrict an agent token's access by IP address** (formerly called **Restrict access for a cluster token by IP address**, whose background content I had to re-write substantially as it wasn't clear) - now documents processes for the Buildkite interface and REST API
    - **Manage maintainers on a cluster** (formerly called **Add maintainers to a cluster**) - now also documents process for removing maintainers
    - **Move a pipeline to a specific cluster** (formerly called **Add pipelines to a cluster**) - now documents processes for the Buildkite interface and REST API

- [Manage queues](https://2642--bk-docs-preview.netlify.app/docs/clusters/manage-queues) - a completely new page and for existing content (indicated by 'formerly called'):
    - **Setting up queues** - explains a little more detail on how queues would typically be set up.
    - **Create a new queue** (formerly called **Create a queue** on the original **Manage clusters** page) - now documents processes for the Buildkite interface and REST API
[...]

PR #2664, opened 2024-02-21
Wording tweaks for Buildkite hosted agents docs

gilesgas:
Hi @LiamPClancy ,

@mbelton-buildkite and I have reviewed this content a little more and we've come up with some suggested wording updates.

A couple of points relating to the biggest changes we've made:

- We are trying to avoid using the docs as a place to publish roadmaps. The docs should be a representation of current Buildkite functionality, even of trial features. However, we shouldn't be documenting additional features that haven't yet been developed, unless they are important additional features that the main trial feature couldn't do without.

- Therefore, we have removed some roadmap features that aren't yet available, but have kept some, e.g. 'Secret management' (which sounds less like a feature instead of 'Buildkite secrets', since this is not yet a branded feature yet) and with 'under development' notes.


pages/pipelines/hosted_agents/overview.md
```diff
@@ -1,20 +1,21 @@
 # What are hosted agents
 
-Buildkite hosted agents provides a fully-managed platform on which you can run your agents.
+Buildkite hosted agents provides a fully-managed platform on which you can run your pipelines.
```
LiamPClancy, line 3: need to stay as agents, pipelines spin up jobs that run across different agents, so a single pipeline could actually be run on both hosted and non-hosted agents, it's the agents that run on the compute not the pipeline.

gilesgas:
Thanks for your review @LiamPClancy .

@mbelton-buildkite , could you please approve this PR, as Liam is now happy with the changes we've made to it.

pages/pipelines/hosted_agents/pipeline_migration.md
```diff
@@ -1,28 +1,28 @@
...
 1. Select _Save Repository_.
 
-## Public repository
+## All repositories
 
-When accessing a public repository from a Buildkite hosted agent, you must ensure the services is using `HTTPS` for checkout.
+When accessing any repository (public or private) from a Buildkite hosted agent, you must also ensure the repository is checked out using `HTTPS`.
 
 1. Navigate to your pipeline settings.
 1. Select GitHub from the left menu.
```
mbelton-buildkite, line 28: I haven't seen the UI, but guessing:

```suggestion
1. Select _GitHub_ from the left menu.  
```

mbelton-buildkite (approved):
Nice!

pages/pipelines/hosted_agents/overview.md
```diff
@@ -2,19 +2,20 @@
 
 Buildkite hosted agents provides a fully-managed platform on which you can run your agents.
```
gilesgas, line 3: Hi @LiamPClancy ,

Would you be OK with this suggested change?

I think the main issue @mbelton-buildkite had with the original wording was that this statement didn't point out the contrast with self-hosted agents (i.e. to make Buildkite hosted agents a distinguishing feature).

```suggestion
Buildkite hosted agents provides a fully-managed platform on which you can run your agents, so that you don't have to manage agents in your own self-hosted environment.
```

gilesgas: Darn, I forgot to commit this before merging, so I will make this part of the clusters GA release doc updates.

gilesgas: Unfortunately I committed this after merging this PR 🤦 , and so I'm implementing this change as part of #2642 .

PR #2674, opened 2024-02-27
Fixes to clusters overview page.

gilesgas:
The current [Clusters overview](https://buildkite.com/docs/clusters/overview#accessing-clusters-and-agents) page (near the top of **Accessing clusters and agents** section) in the Buildkite Docs site has content which appears to have documented a snapshot of clusters behaviour before clusters went GA.

The first sentence in this section now looks completely wrong! This is the sentence:

> If you only have one Cluster with one Queue selecting Agents in the global navigation will take you to your single queue.

If, however, this is still true, please do let me know

Hence, this PR fixes this Clusters overview page to remove this 'in dev' content, along with some other improvements.


gilesgas:
As discussed with @LiamPClancy , this line here:

> If you only have one Cluster with one Queue selecting Agents in the global navigation will take you to your single queue.

is still correct. Therefore, I have added this line back in a slightly re-worded form:

> If you only have one cluster with one queue, selecting Agents in the global navigation takes you directly to your single queue in this cluster. This is typically the case with newly created organizations.

wfr23 (approved):

wfr23 (approved):
I approve

LiamPClancy (approved):

PR #2677, opened 2024-02-28
Change the plan's name to the existing one

ivannalisetska:
We don't have any " Developer plan" and it is confuse our customer.


gilesgas (approved):

gilesgas:
Thanks for picking up on this @ivannalisetska ! 🎉

PR #2687, opened 2024-02-29
Update list of job states to include expired

jeremybumsted:
A possible job state is `expired` which is shown in our job state diagram on the same page, however the list did not include this state. This updates the list to include the `expired` state.

image


gilesgas (approved):
Thanks @jeremybumsted , please feel free to merge this now!

PR #2690, opened 2024-03-01
docs: fix example in cli_oidc.md

andrewjeffree:
This just removes the space in the example for the sub oidc claim so it reflects/matches the full example token at the base of the page.


gilesgas (approved):

gilesgas:
Thanks very much @andrewjeffree ! Merging this in now :-)

PR #2695, opened 2024-03-04
Ignore CVE-2024-0553 and CVE-2024-0567

gilesgas:
Following on from [buildkite/buildkite PR #15823](https://github.com/buildkite/buildkite/pull/15823), implement the same here.


gilesgas:
Hi @dannymidnight and @yob , we seem to be getting a lot of build failures on `main` and so after discussion with @mbelton-buildkite , implemented a fix similar to what he's done in the past here #2631 .

dannymidnight (approved):
👍

PR #2696, opened 2024-03-04
Move deploy to elastic-runners queue with OIDC role for cross account access

amu-g:
Historically, we have used the `deploy` queue to manage resources in the production account. These agents run inside the account, and have fairly broad permissions that are used by several pipelines.

Instead of relying on the agents to have the permissions we need, we can use an OIDC assumable role with custom permissions. The benefits being:

- The role can be tightly scoped to the permissions needed to deploy `docs`
- The role is only assumable by the `main` branch of `docs-main` pipeline
- We can run the deploy steps on lower privilege agents

This pattern was recently used in the [site repo](https://github.com/buildkite/site/pull/2898).

The [role](https://github.com/buildkite/ops/pull/2137) has permissions to interact with ECR and ECS. However, we may need to tweak the permissions to add anything that's found to be missing.

Once the pipeline is working smoothly, we can remove buildkite/docs from the allowlist on the `deploy` agents.

### DIFF
--- a/.buildkite/pipeline.deploy.yml
+++ b/.buildkite/pipeline.deploy.yml
@@ -6,8 +6,10 @@ steps:
     if: |
       build.branch == "main"
     agents:
-      queue: deploy
+      queue: elastic-runners
     plugins:
+      - aws-assume-role-with-web-identity#v1.0.0:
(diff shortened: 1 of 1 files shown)


dannymidnight (approved):
Let's give it a whirl!

PR #2700, opened 2024-03-04
fix: BUILDKIT -> BUILDKITE

HugeIRL:
There's a typo in the no-git-submodules section of the docs where it displays the value as:

BUILDKIT_DISABLE_GIT_SUBMODULES

When it should be:
BUILDKIT**E**_DISABLE_GIT_SUBMODULES


gilesgas:
Thanks very much for spotting and fixing this @HugeIRL !

gilesgas (approved):

PR #2701, opened 2024-03-05
docs: fix wrong pipeline template mutation example in cookbook

deniz-lktr:
There is no `pipelineTemplateId` field in `pipeline` but a `pipelineTemplate` field with an `id`.

The mutation without a correct format will fail with `"Field 'pipelineTemplateId' doesn't exist on type 'Pipeline'"`


gilesgas (approved):
Thanks for picking this up @deniz-lktr , I'll merge this now :-)

PR #2704, opened 2024-03-08
Add context around [skip ci] use and builds not triggering

mcncl:
`[skip ci]` is a bit of a blunt instrument and recently we've seen > 1 case of confusion caused where builds haven't triggered due to its use.

I'd like to therefore add an informative message to the page which may aid in reducing the confusion.


gilesgas (approved):
Thanks @mcncl , please merge this PR 👍

PR #2705, opened 2024-03-10
Clarify behaviour of group merging

mhaylock:
Group merging currently has a stricter criteria than the documentation suggested.

Specifically:
- The group to merge into must be the same group that the uploading step belongs to (this was in the documentation but not all that clear).
- Merging only happens if the matching group step is the first step within the uploaded pipeline YAML (not mentioned at all).

### DIFF
--- a/pages/pipelines/group_step.md
+++ b/pages/pipelines/group_step.md
@@ -174,9 +174,9 @@ steps:
 
 ## Group merging
 
-If you upload a pipeline that has a `group` or `label` that matches a `group` or `label` in the pipeline that uploaded it, those groups will be merged together in the Buildkite UI.
+If you upload a pipeline that has a `group` or `label` that matches _the group of the step that uploaded it_, those groups will be merged together in the Buildkite UI.
 
-Group merging is only possible when a group in the uploaded pipeline matches the group step that uploaded this pipeline.
+This merging behavior only applies if the group step with the matching `group` or `label` is the first step within the uploaded pipeline.
 
 Note that inside a single pipeline, groups with the same `group` or `label` will not be merged in the Buildkite UI.
 


pages/pipelines/group_step.md
```diff
@@ -174,9 +174,9 @@ steps:
 
 ## Group merging
 
-If you upload a pipeline that has a `group` or `label` that matches a `group` or `label` in the pipeline that uploaded it, those groups will be merged together in the Buildkite UI.
+If you upload a pipeline that has a `group` or `label` that matches _the group of the step that uploaded it_, those groups will be merged together in the Buildkite UI.
```
gilesgas, line 177: Thanks very much for this @mhaylock !

Since our [writing style guide](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#ui-elements) currently suggests using italics to refer to reference Buildkite UI elements in our docs, I'm trying to avoid extensive use of italics for other purposes (probably with the only exception being text that defines new terms).
The use of italics is a non-standard way for representing UI elements in documentation and at some point in the near future (when I have a little more bandwidth), I'd like to change this to bold, which is a much more common standard used by Google and Microsoft.

Therefore, I'd suggest removing italics here unless you feel it really adds value.

```suggestion
If you upload a pipeline that has a `group` or `label` that matches the group of the step that uploaded it, those groups will be merged together in the Buildkite UI.
```

gilesgas (approved):

pages/pipelines/group_step.md
mhaylock: Thanks @gilesgas, that's very reasonable - I think it'll be fine without the emphasis so I've removed the italics as suggested 👍

gilesgas: Thanks @mhaylock , I've already approved your update :-) Please feel free to merge this once the build is completed.

PR #2708, opened 2024-03-11
Specified dependencies by default

ivannalisetska:


pages/pipelines/_build_states.md
```diff
@@ -1,8 +1,7 @@
...
 You can query for `finished` builds to return builds in any of the following states: `passed`, `failed`, `blocked`, or `canceled`.
 
->🚧
-> When a <a href="/docs/pipelines/trigger-step">triggered build</a> fails, the step that triggered it will be stuck in the <code>running</code> state forever.
-
->📘
+> 🚧
+> When a [triggered build](/docs/pipelines/trigger-step) fails, the step that triggered it will be stuck in the `running` state forever.
 > When all the steps in a build are skipped (either by using skip attribute or by using `if` condition), the build state will be marked as `not_run`.
+> By default, all steps depend on the step that uploads them, and they will not run until that one is finished.
```
gilesgas, line 7: A couple of minor wording tweak suggestions...

```suggestion
> When all the steps in a build are skipped (either by using the skip attribute or by using the `if` condition), the build state will be marked as `not_run`.
> By default, all steps depend on the step that uploads them, and they will not run until the uploading step is finished.
```

When you said 'that one', were you referring to the uploading step?
If so, it'd be better to refer to 'that one' as 'the uploading step' to be clear what it is you're referring to as 'that one' is inherently ambiguous.

PR #2714, opened 2024-03-13
Adding Java to TA language nav

meghan-kradolfer:
We now have a golden path for setting up Java repos, so we should update the docs to allow customers to easily find the documentation on setup. 

At the moment a customers have to go to References -> Importing JUnit XML. It would be easier for them just to see Java in the languages and then link to this page.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -126,6 +126,7 @@ Rails.application.routes.draw do
   get "/docs/agent/upgrading",           to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/agent/upgrading-to-v3",     to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/clusters/queue-metrics", to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
+  get "/docs/test-analytics/java", to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
 
   # Old docs routes that we changed around during the development of the v3 agent docs
   get "/docs/agent/upgrading-to-v2",    to: redirect("/docs/agent/v2/upgrading-to-v2",            status: 301)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -125,8 +125,8 @@ Rails.application.routes.draw do
   get "/docs/agent/uploading-pipelines", to: redirect("/docs/agent/v3/cli-pipeline",               status: 301)
   get "/docs/agent/upgrading",           to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/agent/upgrading-to-v3",     to: redirect("/docs/agent/v3/upgrading",                  status: 301)
-  get "/docs/clusters/queue-metrics", to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
-  get "/docs/test-analytics/java", to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
+  get "/docs/clusters/queue-metrics",    to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
+  get "/docs/test-analytics/java",       to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
 
   # Old docs routes that we changed around during the development of the v3 agent docs
   get "/docs/agent/upgrading-to-v2",    to: redirect("/docs/agent/v2/upgrading-to-v2",            status: 301)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -125,8 +125,8 @@ Rails.application.routes.draw do
   get "/docs/agent/uploading-pipelines", to: redirect("/docs/agent/v3/cli-pipeline",               status: 301)
   get "/docs/agent/upgrading",           to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/agent/upgrading-to-v3",     to: redirect("/docs/agent/v3/upgrading",                  status: 301)
-  get "/docs/clusters/queue-metrics",    to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
-  get "/docs/test-analytics/java",       to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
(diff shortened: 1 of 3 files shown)


meghan-kradolfer:
Note: Because the new page goes to the same page as 'Importing JUnit XML' both nav items are highlighted. Is this ok? Or is better to create seperate views?
Screenshot 2024-03-14 at 11 09 05 AM

KatieWright26:
> Note: Because the new page goes to the same page as 'Importing JUnit XML' both nav items are highlighted. Is this ok? Or is better to create seperate views? Screenshot 2024-03-14 at 11 09 05 AM

It does seem a bit weird / the duo highlighting might be unclear as to why we're doing that. I feel like people would raise it as a bug. Does the highlighting work similarly to the `_nav.html.erb` where you can pass a param action to signify which nav element to highlight?

pda:
Interesting re. the nav highlighting.

It gets me thinking: nav highlighting aside, would people find it odd/sneaky/_something_ that multiple nav items go to a single page which is more generic than some of those nav items suggest?

I guess the ideal thing would be adding a new Java-specific page which gives some guidance on how to get JUnit XML reports out of their Java test suite, and then links to the general Importing JUnit XML page.

meghan-kradolfer:
> Does the highlighting work similarly to the `_nav.html.erb` where you can pass a param action to signify which nav element to highlight?

Nope, the nav items are added through a yaml file. So the highlighting is done programming based of the url/path

meghan-kradolfer:
> It gets me thinking: nav highlighting aside, would people find it odd/sneaky/_something_ that multiple nav items go to a single page which is more generic than some of those nav items suggest?
> 
> I guess the ideal thing would be adding a new Java-specific page which gives some guidance on how to get JUnit XML reports out of their Java test suite, and then links to the general Importing JUnit XML page.

What if we remove the 'Importing JUnit XML' nav and just have that under Java instead? It sounds like we only import JUnit XML so people would be looking for languages - Java anyway?

pda:
> What if we remove the 'Importing JUnit XML' nav and just have that under Java instead? It sounds like we only import JUnit XML so people would be looking for languages - Java anyway?

I don't think so, because JUnit XML is the common format emitted by multiple non-Java languages.
e.g. a Rust test runner: https://nexte.st/book/junit.html

meghan-kradolfer:
> > What if we remove the 'Importing JUnit XML' nav and just have that under Java instead? It sounds like we only import JUnit XML so people would be looking for languages - Java anyway?
> 
> I don't think so, because JUnit XML is the common format emitted by multiple non-Java languages. e.g. a Rust test runner: https://nexte.st/book/junit.html

Ohh right, good to know! Hmm, maybe the best option is adding a new Java specific page. @gchan keen to get your reckons on this as well?

gilesgas:
It might be worthwhile considering implementing a redirect, similar to when you click on the **Clusters > Queue metrics** nav item in the **Pipelines** area of the docs and it redirects you **Cluster queue metrics** page of the docs.

Screenshot 2024-03-15 at 8 58 48 am

That way, you can avoid the [odd multiple-page highlighting issue](https://github.com/buildkite/docs/pull/2714#issuecomment-1995983095) you described above.

Take a look at how this is implemented in the source code and let me know if you need help.

I hope to document some more best practices on how to contribute to the Buildlkite Docs at some point in the near future, when time permits.

meghan-kradolfer:
Thanks @gilesgas! @pda and @gchan how do you feel about this PR being just to set up that redirect, so then it solves the initial issue of Java language not being immediately findable? 

I think if we are adding a whole new set of instructions for Java then we may also need to do this in the suite setup page in TA and I can write a couple of extra tickets to do that

config/routes.rb
```diff
@@ -126,6 +126,7 @@
   get "/docs/agent/upgrading",           to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/agent/upgrading-to-v3",     to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/clusters/queue-metrics", to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
+  get "/docs/test-analytics/java", to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
```
niceking, line 129: not a biggie but it would be nice to align the `to:` on this line and the line previous!

gilesgas (approved):
Just had a quick chat with @meghan-kradolfer and will approve this now (to unblock merging).
However, please feel free to seek approval from another colleague before merging this.

niceking (approved):
Nice! ✨ Just the one style comment but not a blocker to merging

gchan (approved):

config/routes.rb
gilesgas:
```suggestion
  get "/docs/clusters/queue-metrics",    to: redirect("/docs/pipelines/cluster-queue-metrics",     status: 301)
  get "/docs/test-analytics/java",      to: redirect("/docs/test-analytics/importing-junit-xml",   status: 301)
```

config/routes.rb
```diff
@@ -125,7 +125,8 @@
   get "/docs/agent/uploading-pipelines", to: redirect("/docs/agent/v3/cli-pipeline",               status: 301)
   get "/docs/agent/upgrading",           to: redirect("/docs/agent/v3/upgrading",                  status: 301)
   get "/docs/agent/upgrading-to-v3",     to: redirect("/docs/agent/v3/upgrading",                  status: 301)
-  get "/docs/clusters/queue-metrics", to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
+  get "/docs/clusters/queue-metrics",    to: redirect("/docs/pipelines/cluster-queue-metrics",        status: 301)
+  get "/docs/test-analytics/java",       to: redirect("/docs/test-analytics/importing-junit-xml",        status: 301)
```
gilesgas:
```suggestion
  get "/docs/clusters/queue-metrics",    to: redirect("/docs/pipelines/cluster-queue-metrics",     status: 301)
  get "/docs/test-analytics/java",       to: redirect("/docs/test-analytics/importing-junit-xml",  status: 301)
```

pda:
> how do you feel about this PR being just to set up that redirect, so then it solves the initial issue of Java language not being immediately findable?

Good pragmatic step in the right direction; maintain momentum 💪🏼

PR #2718, opened 2024-03-15
docs: Add GraphQL example of associating Pipeline with a Cluster

tomowatt:
As Clusters have entered General Availability, we do not have an example of how to use the GraphQL to associate Pipelines with Cluster which can be used to programmatically update multiple Pipelines

### DIFF
--- a/pages/apis/graphql/cookbooks/clusters.md
+++ b/pages/apis/graphql/cookbooks/clusters.md
@@ -186,7 +186,7 @@ query getClusterQueueAgent {
 
 ## Associate a pipeline with a cluster
 
-First, [get the Cluster ID](#get-a-pipelines-id) you want to associate the Pipeline with.
+First, [get the Cluster ID](#list-clusters) you want to associate the Pipeline with.
 Second, [get the Pipeline's ID](./pipelines#get-a-pipelines-id).
 Then, use the IDs to archive the pipelines:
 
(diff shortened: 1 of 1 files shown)


pages/apis/graphql/cookbooks/clusters.md
```diff
@@ -184,3 +184,21 @@ query getClusterQueueAgent {
 }
 ```
 
+## Associate a pipeline with a cluster
+
+First, [get the Cluster ID](#get-a-pipelines-id) you want to associate the Pipeline with.
```
gilesgas, line 189: Thanks again @tomowatt !

Looks like what you're linking to is on a different page...

```suggestion
First, [get the Cluster ID](/docs/apis/graphql/cookbooks/pipelines#get-a-pipelines-id) you want to associate the Pipeline with.
```

pages/apis/graphql/cookbooks/clusters.md
```diff
@@ -184,3 +184,21 @@ query getClusterQueueAgent {
...
+Second, [get the Pipeline's ID](./pipelines#get-a-pipelines-id).
```
gilesgas, line 190: Probably better to adopt [our consistent method for linking to other pages within the help docs](https://github.com/buildkite/docs/blob/main/styleguides/markdown-syntax-style.md#internal-links-to-other-pages).

That way, if we ever need to modify these in future (because content may move around), it makes it much easier to conduct a global search and replace on these links...

```suggestion
Second, [get the Pipeline's ID](/docs/apis/graphql/cookbooks/pipelines#get-a-pipelines-id).
```

pages/apis/graphql/cookbooks/clusters.md
tomowatt: Cheers for spotting @gilesgas the error threw me off when looking into the Build due to the matching header names, my bad, but sorted now!

gilesgas (approved):
Thanks @tomowatt , ready for you to merge this 😊

PR #2726, opened 2024-03-21
Remove bits that aren't quite true

sj26:
These claims were true for Packagecloud, but not Buildkite Packages. Buildkite Packages repositories can only contain one package type each. We do allow any distribution to use repositories, but we don't have "sub-repositories" for each ubuntu/debian distribution etc the way Packagecloud does. And authentication tokens are still evolving in product. So it might be easiest to remove these claims for now.

/cc @gilesgas


gilesgas (approved):
Ah, good catch @sj26 , thanks for preparing this!

Approved and feel free to merge after the builds complete successfully

PR #2728, opened 2024-03-25
Fix Algolia results 

dannymidnight:
Something isn't quite right with our Algolia doc scraper configuration. We've got lots of blank links and grouping isn't really happening.

Screenshot 2024-03-25 at 10 59 42 am

Screenshot 2024-03-25 at 10 59 27 am

After reading the [docs](https://docsearch.algolia.com/docs/legacy/config-file/) some more and reviewing some other configurations, it occurred to me that our doc search is missing the top level grouping (ie. Packages, Test Analytics, Pipelines etc.) which I think is causing the issue since that content should live in the `header` of the page.

This PR switches up the selectors to match `lvl0` to the active nav item and removes `h6` headers which aren't being used.

Before merging, I'll manually trigger a crawl from this branch to confirm it all works as expected.

### DIFF
--- a/config/algolia.json
+++ b/config/algolia.json
@@ -3,7 +3,10 @@
   "start_urls": ["https://buildkite.com/docs"],
   "stop_urls": ["agent/v2"],
   "selectors": {
-    "lvl0": ".Nav__link.Nav__link--level1.Nav__link--current",
+    "lvl0": {
+      "selector": ".SiteHeader__nav .Nav__link.Nav__link--level1.Nav__link--current",
+      "global": true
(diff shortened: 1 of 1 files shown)


gilesgas (approved):
Yes I noticed this too @dannymidnight ! Thanks for this fix. Ready for you to merge this in.

dannymidnight:
Ooooo yeah, just tested and it's now looking much better.


Screenshot 2024-03-25 at 3 25 12 pm

Screenshot 2024-03-25 at 3 25 20 pm

PR #2729, opened 2024-03-25
Improve Algolia search hits (part 2)

dannymidnight:
Following on from https://github.com/buildkite/docs/pull/2728, this PR makes a couple minor updates:

- Remove "Beta" from "Packages" search results.
 
  Screenshot 2024-03-25 at 5 43 25 pm

- Add missing headers for GraphQL search results. These hits are currently pretty confusing – this change will make them slightly more obvious.

  Screenshot 2024-03-25 at 5 44 14 pm


app/views/application/_nav_link.html.erb
```diff
@@ -23,7 +23,7 @@
     <%= name %>
 
     <% if pill %>
-      <span class="Nav__pill pill pill--<%= pill %> pill--small"><%= pill %></span>
+      <span class="Nav__pill pill pill--<%= pill %> pill--small" data-algolia-exclude><%= pill %></span>
```
dannymidnight, line 26: There's a special exclusion selector setup that means these won't be included in the DOM when Algolia does it's thing.

https://github.com/buildkite/docs/blob/main/config/algolia.json#L18

gilesgas:
Wow, was this PR generated from an automated script?

Also, it looks like it's currently causing the docs build to fail.

dannymidnight:
> Wow, was this PR generated from an automated script?

Yep, that's right! I messed around with the base template that's used for generating content and then ran the following command to rebuild all the pages using said template.

```
rake graphql:generate
```

dannymidnight:
Just coming back to confirm it's working.

Beta tag removed  ✅ 
Screenshot 2024-03-26 at 9 33 14 am

More context for GraphQL attributes ✅ 

Screenshot 2024-03-26 at 9 34 00 am

PR #2743, opened 2024-04-08
docs: Update NGINX proxy example with best practises

tomowatt:
We've had a customer raise some concerns about our current example so adding extra information for security as the current example shows loose config in terms of not showing usage of the allowing only from the Buildkite IPs and limiting access to specific URLs for the use with Buildkite

By listing the the specific and only URLs/Endpoints required, we can ensure users follow best practices.

Will update PR for review when confirmed all the URLs/Endpoints required for GitHub Enterprise Server.


gilesgas (approved):

gilesgas:
Hi @tomowatt , thanks very much for this contribution! 🎉

Based on your description, I'm not sure if this is still a work in progress or if it's ready to merge in.

Regardless, I've marked this PR as a draft, but I've already approved it. Therefore, if this is ready to merge in, please select **Ready for review** and this PR 🤞 should then be ready to merge in. If you can do this, please proceed with the merge!

I only implemented some minor wording tweaks around the link to the NGINX docs.

tomowatt:
Had a second round of testing and checking and happy with the results 🚀

PR #2757, opened 2024-04-18
SUP-2072: Add example of using GraphQL to get the Download URL for Artifacts from a Build

tomowatt:
We have it documented how to get the Download URLs for Artifacts via the REST API but missing an example of using GraphQL to get them.
In addition mentining the URLs from Buildkite Artifacts expire after 10 minutes rather than 60 seconds with the REST API.

Using GraphQL to get the multiple Download URLs is faster than the REST API.

### DIFF
--- a/data/nav_graphql.yml
+++ b/data/nav_graphql.yml
@@ -9,6 +9,8 @@
     path: apis/graphql/graphql-cookbook
   - name: Agents
     path: apis/graphql/cookbooks/agents
+  - name: Artifacts
+    path: apis/graphql/cookbooks/artifacts
   - name: Builds
     path: apis/graphql/cookbooks/builds
   - name: Clusters
--- a/pages/apis/graphql/cookbooks/artifacts.md
+++ b/pages/apis/graphql/cookbooks/artifacts.md
@@ -1,0 +1,30 @@
+# Artifacts
+
+A collection of common tasks with artifacts using the GraphQL API.
+
+## List download URLs for artifacts from a build
+
+To get the download URLs for artifacts from a build.
+If the artifact is stored on Buildkite-managed artifact storage, the download URL will be valid for only 10 minutes.
+
+```graphql
+query GetDownloadURLsForArtifactsFromBuild {
+  build(uuid: "build-uuid") {
+    jobs(first: 500) {
+      edges {
+        node {
+          ... on JobTypeCommand {
+            artifacts {
+              edges {
+                node {
+                  downloadURL
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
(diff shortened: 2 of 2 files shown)


gilesgas:
Thanks very much for your contribution @tomowatt ! 🎉 

Could I just confirm the following questions with you first:

1. Is this a new page you're adding? If so, you'll need to add an entry for it in the `nav.yml` file. The lack of an entry to this new page/file might explain the cause of the RSpec docs build errors.
2. Is this the start of new GraphQL API content whose REST API equivalent is documented on [this page](https://buildkite.com/docs/apis/rest-api/artifacts)?
3. Does this API feature work independently of [Buildkite Packages](https://buildkite.com/docs/packages)? If the answer to question 2 is yes, then the answer to this question is likely yes too I suspect.

tomowatt:
Hey @gilesgas 

> * Is this a new page you're adding? If so, you'll need to add an entry for it in the `nav.yml` file. The lack of an entry to this new page/file might explain why the RSpec jobs of the docs build are failing.

Yes is a new page, wasn't aware of the `nav.yml` and `nav_graphql.yml` files so thanks, but updated it now!

> * Is this the start of new GraphQL API content whose REST API equivalent is documented on [this page](https://buildkite.com/docs/apis/rest-api/artifacts)?

Yes it is, I felt it was separate topic from the other categories so created the new page

> * Does this API feature work independently of [Buildkite Packages](https://buildkite.com/docs/packages)? If the answer to question 2 is yes, then the answer to this question is likely yes too I suspect.

Yes, it's specifically for Buildkite Artifacts, not related to Packages

gilesgas:
Thanks heaps @tomowatt !

Just with regard to your response:

> Yes is a new page, wasn't aware of the `nav.yml `and `nav_graphql.yml` files so thanks, but updated it now!

We have a [section that's currently part of our Markdown syntax guide](https://github.com/buildkite/docs/blob/main/styleguides/markdown-syntax-style.md#adding-and-naming-new-documentation-pages) (i.e. it should actually be in its own 'contributors guide') on how to add new pages to these docs. However, you've just made me aware of the separate `nav_graphql.yml` file. This content probably needs a little tidying up and clarity to be added, which is something I have planned for the near future (hopefully).

Just going to review the remainder of your PR...

gilesgas (approved):
This is looking great now @tomowatt . Please merge this in!

I was thinking of adding callouts near the top of your new GraphQL API page for artifacts (and the top of its [REST API eqivalent](https://buildkite.com/docs/apis/rest-api/artifacts) page) to disambiguate these features from those provided by Buildkite Packages.

However, I'll leave this be for the time being and will add such callouts if we do happen to receive customer feedback about this, and/or if Buildkite Packages obtains similar API functionality to this existing Buildkite Pipelines feature.

PR #2763, opened 2024-04-22
Clarify metrics not on unclustered

mcncl:
Currently we don't call out that `Unclustered` agents are not included in the advanced queue metrics. This is because they cannot be assigned to `queues` in the same way that the new clusters set us can be.

A `queue` can be targeted but that is simply an Elastic Stack tag, rather than a `queue` in the Buildkite sense.

Turning on `advanced queue metrics` will therefore have no affect on folks who only use `Unclustered`, nor do agents in that "cluster" have an impact on the metrics.

### DIFF
--- a/pages/pipelines/cluster_queue_metrics.md
+++ b/pages/pipelines/cluster_queue_metrics.md
@@ -3,7 +3,7 @@
 Queue metrics show the most important statistics to help you optimize your agent setup and monitor a queue's performance. These statistics are updated on the page every 10 seconds.
 
 > 📘
-> `Unclustered` agents are not reported in advanced queue metrics.
+> _Unclustered agents_ are not reported in advanced queue metrics. Learn more about unclustered agents in [Unclustered agent tokens](/docs/agent/v3/unclustered-tokens).
 
 ## Metrics panels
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/cluster_queue_metrics.md
```diff
@@ -2,6 +2,9 @@
 
 Queue metrics show the most important statistics to help you optimize your agent setup and monitor a queue's performance. These statistics are updated on the page every 10 seconds.
 
+> 📘
+> `Unclustered` agents are not reported in advanced queue metrics.
```
gilesgas, line 6: Thanks very much @mcncl for your contribution! 🎉 

A couple of suggestions, 'unclustered agents' is a key term in this context and should be formatted as such in our revised [writing style guide on key terms](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#key-terms-and-emphasis).

Also, I think it'd help at this point to provide a convenient link for readers if they want to find out more about what unclustered agents (and their tokens) are.

Hence, these suggested changes...

```suggestion
> _Unclustered agents_ are not reported in advanced queue metrics. Learn more about unclustered agents in [Unclustered agent tokens](/docs/agent/v3/unclustered-tokens).
```

mcncl: Committed your suggestions @gilesgas

gilesgas (approved):
Perfect @mcncl , ready for you to merge this in 🎉

PR #2770, opened 2024-04-26
Update managing_log_output.md

stephanieatte:
This PR adds information about the minimum byte length for redacted environment variables

`Warning: Some variables have values below minimum length (6 bytes) and will not be redacted: K_SECRET 
`

### DIFF
--- a/pages/pipelines/managing_log_output.md
+++ b/pages/pipelines/managing_log_output.md
@@ -133,7 +133,7 @@ To further tighten the security in a Buildkite organization, you can use the [AP
 
 Agents can redact the values of environment variables whose names match common patterns for passwords and other secure information before the build log is uploaded to Buildkite.
 
-If the environment variable is shorter than the minimum length of 6 bytes, it will not be redacted.
+If the environment variable's value is shorter than the minimum length of 6 bytes, then this value will not be redacted.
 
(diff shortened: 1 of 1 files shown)


pages/pipelines/managing_log_output.md
```diff
@@ -133,6 +133,8 @@ To further tighten the security in a Buildkite organization, you can use the [AP
 
 Agents can redact the values of environment variables whose names match common patterns for passwords and other secure information before the build log is uploaded to Buildkite.
 
+If the environment variable is shorter than the minimum length of 6 bytes, it will not be redacted.
```
gilesgas, line 136: Thanks very much @stephanieatte , just adding some clarifications...

```suggestion
If the environment variable's value is shorter than the minimum length of 6 bytes, then this value will not be redacted.
```

gilesgas: Reasons:
- I believe it's the variable's _value_ that gets redacted and not the variable name?
- Using 'it' as the subject of a new clause isn't always 100% clear as to what that 'it' refers to.

stephanieatte: Yes @gilesgas its the value of the environment variable, will make the changes

gilesgas (approved):

PR #2771, opened 2024-04-29
Update docker push to be buildkit compatible [PLT-2470]

yob:
The agent that builds these images runs buildkit, and we can instruct buildkit to push the images directly to the repository rather than export them to the docker image store then push. This should s lightly speed up this step.

I've added --provenance=false so buildkit pushes a regular image, rather than an image manifest. Some parts of our deploy pipeline are not ready for image manifests yet.


dannymidnight (approved):
:shipit:

yob:
If there's any surprises we're not going to know until after merging and watching the deploy run. If I've accidentally broken things, I'll follow up straight away ❤️

PR #2772, opened 2024-04-29
SUP-1904: Update info on same key attribute value with groups and steps

tomowatt:
This is only specifically mentioned in the group step page, but groups and steps cannot share the same value for the `key` attribute within Build, so updating the info block.


gilesgas (approved):
Thanks @tomowatt 👍

PR #2781, opened 2024-05-02
More info on how to download artifacts from different builds

AthreyaRay:
Modified the docs to specify that we need a build id if we are trying to download artifacts uploaded in different builds


AthreyaRay:
Need to make some changes. Closing the PR

PR #2785, opened 2024-05-07
change to hosted queue

yob:
We're experimenting with a new queue and docs feels like an excellent place to experiment. There's quite a few steps, and there's a chance we'll get a speed up.

Safe to revert if there's any surprises, the goal isn't to get in the way of other work!


dannymidnight (approved):
> there's a chance we'll get a speed up

Gimme! Gimme!

PR #2789, opened 2024-05-09
feat(analytics): add utm cookie

hannahcancode:
This felt like the most rails-y way to do this.

Adds the UTM 🍪 for non-third party cookie times.


zchase (approved):
Awesome! I see the script in the `head` in the preview so this seems fine to me (I agree on it being the same issue as vercel). :shipit:

PR #2761, opened 2024-04-19
Securing your Buildkite Agent and environment variables

rianmcguire:
Add a callout about environment variables that may allow arbitrary command execution.

PR #2792, opened 2024-05-10
Add paragraph break in "Custom hooks and environment" variables callout

rianmcguire:
I was just having a peek at https://github.com/buildkite/docs/pull/2761 post-deploy, and I noticed that there's no break between the paragraphs in the rendered output:
image

Adding a hacky `<p>` between the paragraphs makes it do the right thing. If you know a nicer way to achieve the same result, let's do that instead!

image

### DIFF
--- a/pages/agent/v3/securing.md
+++ b/pages/agent/v3/securing.md
@@ -40,7 +40,7 @@ Command line evaluation can be disabled by setting [`no-command-eval`](/docs/age
 
 > 🚧 Custom hooks and environment variables
 > If you have a custom `command` hook, using `no-command-eval` will have no effect on your command execution. See [Allowing a list of plugins](#allowing-a-list-of-plugins) and [Custom bootstrap scripts](#customizing-the-bootstrap) for examples of how to completely lock down your agent from arbitrary code execution.
->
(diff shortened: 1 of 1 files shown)


pages/agent/v3/securing.md
```diff
@@ -40,7 +40,7 @@ Command line evaluation can be disabled by setting [`no-command-eval`](/docs/age
 
 > 🚧 Custom hooks and environment variables
 > If you have a custom `command` hook, using `no-command-eval` will have no effect on your command execution. See [Allowing a list of plugins](#allowing-a-list-of-plugins) and [Custom bootstrap scripts](#customizing-the-bootstrap) for examples of how to completely lock down your agent from arbitrary code execution.
->
+><p>
```
gilesgas, line 43: Thanks @rianmcguire , this makes the callout look a little clearer.

I'd only recommend adding a single space between the initial `>` and `<p>` to make it clearer that these are separate pieces of markup.

```suggestion
> <p>
```

gilesgas (approved):
Ready for you to merge this once you've address my earlier comment. Thanks!

PR #2793, opened 2024-05-12
Fix '- wait:' syntax for pipeline dependencies page (and wait step page too).

gilesgas:
There was some discussion about clarifying the code examples depicting `wait` step syntax in our docs.

This PR should also hopefully address this piece of direct page feedback we got here:

Screenshot 2024-05-13 at 9 19 07 AM

### DIFF
--- a/pages/pipelines/dependencies.md
+++ b/pages/pipelines/dependencies.md
@@ -16,7 +16,7 @@ A wait step, as in the example below, is dependent on all previous steps complet
 steps:
   - command: "one.sh"
   - command: "two.sh"
-  - wait:
+  - wait: ~
   - command: "three.sh"
   - command: "four.sh"
 ```
@@ -88,7 +88,7 @@ To ensure that a step is not dependent on any other step, add an explicit empty 
 ```yml
 steps:
   - command: "tests.sh"
-  - wait:
+  - wait: ~
   - command: "lint.sh"
     depends_on: ~
 ```
--- a/pages/pipelines/dependencies.md
+++ b/pages/pipelines/dependencies.md
@@ -16,7 +16,7 @@ A wait step, as in the example below, is dependent on all previous steps complet
 steps:
   - command: "one.sh"
   - command: "two.sh"
-  - wait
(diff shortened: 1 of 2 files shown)


tessereth (approved):
Hm, interesting. The old syntax wasn't "wrong" but the new way is easier to extend after you've copy-pasted it. So this seems reasonable to me.

gilesgas:
Thanks @tessereth , I'll merge this once the build passes :-)

pages/pipelines/dependencies.md
```diff
@@ -16,7 +16,7 @@ A wait step, as in the example below, is dependent on all previous steps complet
 steps:
   - command: "one.sh"
   - command: "two.sh"
-  - wait
+  - wait:
```
pda, line 19: I wonder if `wait: ~` is clearer than `wait:`.

I think the string "wait" is pretty clear:

```yaml
- command: "two.sh"
- wait
- command: "three.sh"
```

… but as @tessereth mentions it's harder/non-obvious to extend.

I worry that `- wait:` as an object _without any keys_ and without the `~` explicit null is a bit harder / less clear to read… the `:` seems to imply that whatever is next is part of the `wait`, but the `- command …` that is next isn't part of the wait.

```yaml
- command: "two.sh"
- wait:
- command: "three.sh"
```

So perhaps this is clearer in that particular scenario?

```yaml
- command: "two.sh"
- wait: ~
- command: "three.sh"
```

gilesgas: I feel the docs should reflect best practices that readers should follow, as well as reducing any ambiguity that causes readers to think 'am I doing the right thing' (with this syntax)?

Therefore, I'm happy to take your guidance on this, and I'll amend this PR accordingly...

pda (approved):
It's hard to reconcile “how I'd write this pipeline” vs “what's a good starting point for people to base their own pipelines on” with a dash of “YAML is weird sometimes”.

I _think_ this proposed syntax probably comes closest to solving for all of those 👍🏼

PR #2794, opened 2024-05-13
Switch to less-diverged branch of ECR scan plugin

duckalini:
## Context

Update our ECS scanning Buildkite plugin to a version that's closer to upstream:

Depends on https://github.com/buildkite/ecr-scan-results-buildkite-plugin/pull/11

- Switch to updated plugin (closer to upstream)
- Change format of ignore list, because upstream's parallel implementation is slightly different to ours
- Add fail-build-on-plugin-failure flag, to preserve our current plugin behaviour

### DIFF
--- a/.buildkite/ecr-scan-results-ignore.yml
+++ b/.buildkite/ecr-scan-results-ignore.yml
@@ -1,0 +1,21 @@
+ignores:
+  - id: CVE-2023-29007 # git 1:2.39.2-1.1
+  - id: CVE-2023-25652 # git 1:2.39.2-1.1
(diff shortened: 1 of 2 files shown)


gilesgas:
Hi @duckalini , this looks fine to, but as I'm not expert, I'll wait for someone in Platform to review this first.

Otherwise, if you have trouble getting this review, let me know—I can then approve it so you can merge this in.

yob (approved):
looks good to me. In this repo the ECR scan only happens on main builds so we won't know for sure this will work until after we merge, but the same upgrade has been made to a variety of repos and I don't expect any issues :+1:

PR #2786, opened 2024-05-07
Document new features for Buildkite hosted agents GA release

gilesgas:
These include the following new pages:

- **Terminal access** to hosted agents (preview link: https://2786--bk-docs-preview.netlify.app/docs/pipelines/hosted-agents/terminal-access)
- **Buildkite secrets** (preview link: https://2786--bk-docs-preview.netlify.app/docs/pipelines/buildkite-secrets)

As well as the following, which involve updates to existing pages:

- Merging in [this reviewed PR](https://github.com/buildkite/docs/pull/2737) on cached volumes, which has been added to the end of the **Linux hosted agents** page (preview link: https://2786--bk-docs-preview.netlify.app/docs/pipelines/hosted-agents/linux#cache-volumes)
    - Add of the **Billing model** section to this page.
- Substantial updates to add support for Buildkite hosted agents through the **Manage queues** page (https://2786--bk-docs-preview.netlify.app/docs/clusters/manage-queues), including addition of:
    - New **Agent infrastructure** section, including clarification that API methods do not support hosted agent queue creation.
    - Integration of new hosted agents steps to the **Create a queue > Using the Buildkite interface** section.

PR #2796, opened 2024-05-14
Post-Buildkite hosted agents GA release doc updates

gilesgas:
Following on from https://github.com/buildkite/docs/pull/2786, the doc updates in this PR achieve the following:

- Removes content duplication (and outdated materials) from the [**Buildkite hosted agents**](https://buildkite.com/docs/pipelines/hosted-agents/overview) overview page (see preview page [here](https://2796--bk-docs-preview.netlify.app/docs/pipelines/hosted-agents/overview)).
- Heavily modifies the [**Getting started**](https://buildkite.com/docs/tutorials/getting-started) page to incorporate support for Buildkite hosted agents (see preview page [here](https://2796--bk-docs-preview.netlify.app/docs/tutorials/getting-started)).
- Role clarifications (i.e. mentioning which procedures need to be done by a 'cluster maintainer' with links through to how to configure a cluster maintainer) through the [clusters](https://buildkite.com/docs/clusters/manage-clusters), [queues](https://buildkite.com/docs/clusters/manage-queues) and [agent tokens](https://buildkite.com/docs/agent/v3/tokens) pages (see preview pages [here](https://2796--bk-docs-preview.netlify.app/docs/clusters/manage-clusters), [here](https://2796--bk-docs-preview.netlify.app/docs/clusters/manage-queues), and [here](https://2796--bk-docs-preview.netlify.app/docs/agent/v3/tokens), respectively).
    - This requirement was discovered as part of updating the Getting started guide.

Things to investigate/add:

- Do we reference the Terminal access feature as an SSH hosted agent feature (in other products)?
- Change 'control panel' refs to 'control plane'. ✅


pages/pipelines/buildkite_secrets.md
```diff
@@ -69,11 +69,10 @@ Buildkite secrets are designed, with the following controls in place:
 - All access to the secrets are logged.
 - Employee access to secrets is strictly limited and audited.
 
-> 📘 Buildkite secrets is not a zero-knowledge system
-> Buildkite owns, stores and manages the keys used for encrypting the secrets stored in the service at rest and in transit. Buildkite recommend customers implement additional controls to manage the lifecycle of secrets stored within Buildkite secrets, in addition to any monitoring capability you may require in line with your risk appetite.
-
 ## Best practices
 
+Buildkite secrets is not a zero-knowledge system, whereby Buildkite owns, stores, and manages the keys used for encrypting the secrets stored in the service at rest and in transit. You should implement additional controls to manage the lifecycle of secrets stored within Buildkite secrets, in addition to any monitoring capability you may require in line with your risk appetite. For example:
+
```
gilesgas, line 75: FYI only @matthewborden , I hope you don't mind my moving the content from this note to the top of the **Best practices** section. It seemed that this content was better-suited to this section as it refers to the need to 'implement additional controls', which are effectively what are listed in this "Best practices" section.

While in the note, it felt as though the reader was left up in the air as to what these additional controls wer.

mbelton-buildkite (approved):
I won't do a detailed review at this stage, but happy for this to merge and address any further updates in a follow-up PR.

Initial feedback shared internally:

- **Active voice:** "A [Buildkite hosted agent](https://2796--bk-docs-preview.netlify.app/docs/pipelines/hosted-agents/overview) can be created" -> "You can create..."
- **Too many parenthesis:** "(and another queue is marked (default))" -> I suggest removing the outer ones.
- **Extra info:** I wonder if you can remove the callout in the hosted agents flow about not needing an agent token. Seems unnecessary to me so just adds complexity. It's also the first usage of that term on the page so might be confusing. I think you can just delete the callout.
- Ideally avoid the "and/or" construction.
- **Key terms:** I think you can reduce the italics used in the paragraph starting with  "Compared to fully self-hosted solutions, where you run both..."

gilesgas:
Thanks for your review and approval @mbelton-buildkite , will implement these in a subsequent commit, but I did have a couple of queries about some points:

> **Extra info:** I wonder if you can remove the callout in the hosted agents flow about not needing an agent token. Seems unnecessary to me so just adds complexity. It's also the first usage of that term on the page so might be confusing. I think you can just delete the callout.

I still think it's worthwhile keeping the content of the note, just as additional contrasting info, and so I was thinking of adding it as an additional sentence to the end of the second point in [**Install and run a self-hosted agent**](https://2796--bk-docs-preview.netlify.app/docs/tutorials/getting-started#set-up-an-agent-install-and-run-a-self-hosted-agent).

> Ideally avoid the "and/or" construction.

If we avoid 'and/or' constructs, what word or term should we use to indicate item1 or item2, or both item1 and item2? (Since 'or' could be interpreted as the exclusive 'or'?

PR #2800, opened 2024-05-15
Add container cache docs and update images of cache settings

l-suzuki:
Adds documentation for container caching to the hosted agents/linux page. Updates the screenshots of the cache settings to reflect the new UI.


gilesgas (approved):
Thanks @l-suzuki for this addition 🎉 

I just pushed up a fix to your image reference as it took me a while to understand from the breaking build output that this was the cause of the build failures.

Please proceed with merging this as this feature is now live :-)

l-suzuki:
> Thanks @l-suzuki for this addition 🎉
> 
> I just pushed up a fix to your image reference as it took me a while to understand from the breaking build output that this was the cause of the build failures.
> 
> Please proceed with merging this as this feature is now live :-)

Thanks @gilesgas - fyi the git mirror caching is live but the container caching hasn't been enabled yet (I only just merged it!) We're going to do a bit of prod testing and then turn it on for everyone :)

PR #2802, opened 2024-05-16
Update agent cli docs with better description for 'secret get'

gilesgas:
And fix overview page and alphabetise everything else, with the exception of start.


DrJosh9000 (approved):
LGTM

PR #2805, opened 2024-05-21
fixed capitalisation

LiamPClancy:
This is a very small PR to remove some confusion around team member permissions. 
It was raised through the support channel that if Team member permissions are set to not allow pipeline creation that some members of teams can still create pipelines. 

After some investigation we found that Team Maintainers are not bound by the team member pipeline creation permissions, so I'm just adding this to the documentation so it's clear that team maintainers will be able to create pipelines.

### DIFF
--- a/pages/team_management/permissions.md
+++ b/pages/team_management/permissions.md
@@ -26,6 +26,7 @@ Users who are team maintainers can:
 - Add users to existing teams, of which they are the maintainer
 - Remove users from their teams
 - Set read, write, and edit permissions for users on pipelines in their team
+- Create pipelines
 
 All users in a team have the same level of access to the pipelines in their team. If you need to have more fine grained control over the pipelines in a team, you can create more teams with different permissions.
 


gilesgas:
Thanks @LiamPClancy . I am substantially re-writing this page (as part of https://github.com/buildkite/docs/pull/2803) because its brevity is pretty confusing and they aren't written from a role-based perspective.

The change you're making as part of this PR is not entirely true. Once the teams feature is active for an org, a user just needs to be a 'team member' to be able to create a pipeline within that org (i.e. they don't need to be a 'team maintainer').

Here's a preview of this page re-write (https://2803--bk-docs-preview.netlify.app/docs/team-management/permissions).

I'll take what I believe you're wanting to convey (which I think is that 'team members' as well as 'team maintainers' who are still a 'team member' can create pipelines in an org) and integrate this into the newer version of this page ^

LiamPClancy:
Hey @gilesgas 

I've had a look at the preview and I don't think this line is actually correct


When the [teams feature is enabled](https://2803--bk-docs-preview.netlify.app/docs/team-management/permissions#manage-teams-and-permissions), any user can create a new pipeline, as long as this user is a member of at least one team within the Buildkite organization.

what I was trying to convey with my changes is that when there are specific team member permissions set to disallow the creation of pieplines these permissions only apply to team members, and team maintainers will still have the permission to create pipelines. 

team member permission can be set here. 
Organisation > settings > teams > <selected team> > settings

Screenshot 2024-05-22 at 9 01 26 AM


take the following example

an organization has a team called team A
and a user called Liam
and a user called Giles

if Liam are only a member of team A - and is a team member
and team A settings disallows the creation of pipelines
Liam will not be able to create pipelines

If Giles is only a member of team A - and is a team maintainer
Giles will be able to create pipelines 
despite members of team A being disallowed from creating pipelines. 

so I don't think 'any user can create a new pipeline, as long as this user is a member of at least one team within the Buildkite organization.' is true in the case of Liam, we have a request from a customer to make this difference between being a team member and a team maintainer clearer.

gilesgas:
Ah thanks very much for clarifying this point @LiamPClancy and sorry that I didn't initially get what you'd written.
After my own testing of the teams feature as part of revamping this page in the docs (and now its related test suites one in the test analytics docs), this is one use case I don't believe I thoroughly tested, so thanks again for the clarity!
Correct me if I'm wrong, but I'm also assuming _team maintainers_ of a given team could also create test suites (and now packages) within this team despite its team member permissions not being granted for pipelines, test suites and packages.

gilesgas (approved):
Please feel free to merge this in now, and'll ensure this information is incorporated clearly in its revamped version.

PR #2807, opened 2024-05-22
update: add triggered pipeline cancellation explanation

HugeIRL:
This PR is the initial idea and needs some love on how to explain the following scenario (example 2) when triggered pipelines have `Cancel Intermediate Builds` turned on and multiple pipelines trigger a build. We still need example 1 written in the docs in some way that makes sense.

Basically, we need to find a way to add these two examples clearly in the docs to outline behaviours of triggered pipelines and how they interact with cancellation either manual, or via `Cancel Intermediate Builds`.

**Example 1:** `Cancel Intermediate Builds` Enabled (3 pipelines, same commit builds)

- Pipeline A 
- Pipeline B
- Pipeline C 

Action: Pipeline A triggers a build on Pipeline B (first)
Action: Pipeline C triggers a build on Pipeline B (second, right after A starts to execute)

Pipeline A's build will be cancelled as it is not the `newest` build on the same commit in the chain. Pipeline C's trigger of Pipeline B will run to completion, as it is the `newest` build on the same commit in the chain.

--

**Example 2:** No `Cancel Intermediate Builds` (2 pipelines, same commit builds)

- Pipeline A does not have `Cancel Intermediate Builds` enabled
- Pipeline B does not have `Cancel Intermediate Builds` enabled

Action: Pipeline A triggers a build on Pipeline B, but Pipeline A's build is manually cancelled.

If you cancel Pipeline A before any triggering steps have occurred, then Pipeline B never fires.
If you cancel Pipeline A after any triggering steps have occurred, then Pipeline B proceeds as normal and the build goes on.

**Cancellation in Pipeline A:**
- Manually cancelled builds here will be stopped and any ongoing jobs in that build get stopped and marked as cancelled.
- Steps that have not started yet will not execute, including any trigger steps that might call Pipeline B.

**Effects on Pipeline B:**
- If the trigger to this pipeline is already executed before the cancellation in Pipeline A, Pipeline B will proceed as normal. Pipeline B is independent in this regard and does not actually know about the cancel in Pipeline A.
- If the trigger to this pipeline is part of the steps that are not yet executed in Pipeline A, (i.e, the trigger happens AFTER you've cancelled it step wise), Pipeline B does not fire as Pipeline B does not know it needs to fire from Pipeline A.


pages/pipelines/trigger_step.md
```diff
@@ -239,3 +239,7 @@ steps:
 
 ```
+
+> 📘 A note on triggered pipeline dependencies and cancellation
+> When you have multiple pipelines that trigger a build with the same commit on an additional pipeline, while also having <code>Cancel Intermediate Builds</code> enabled, the newest build will be the only build that runs. 
+> So if you have <b>Pipeline A</b> trigger a build with a commit on <b>Pipeline B</b> and <b>Pipeline C</b> trigger a build with the <b>same</b> commit on <b>Pipeline B</b>, whichever pipeline triggered the <b>first</b> build will be <b>cancelled</b> in favour of the newest triggered build.
```
gilesgas, line 245: Thanks very much for your contribution @HugeIRL !

Just a few points:

- As indicated in our [Markdown syntax style guide](https://github.com/buildkite/docs/blob/main/styleguides/markdown-syntax-style.md#callouts), it is better to use Markdown syntax within callouts as they make the source more readable.
- As indicated in the accompanying [Writing style guide](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#ui-elements), note the difference between UI elements and emphasis.
- I removed some additional spacing at the end of a line, which was causing the linter to fail.

```suggestion

> 📘 A note on triggered pipeline dependencies and cancellation
> When you have multiple pipelines that trigger a build with the same commit on an additional pipeline, while also having `Cancel Intermediate Builds` enabled, the newest build will be the only build that runs.
> So if you have **Pipeline A** trigger a build with a commit on **Pipeline B** and **Pipeline C** trigger a build with the _same_ commit on **Pipeline B**, whichever pipeline triggered the _first_ build will be _cancelled_ in favor of the newest triggered build.
```

gilesgas: I'm assuming **Cancel Intermediate Builds** (verbatim) refers to something in the Buildkite UI?

gilesgas (changes requested):
Please read/address the comments above, then I'll approve :-)

pages/pipelines/trigger_step.md
HugeIRL: Yep! That's correct. Your suggestions make sense - on that note though I don't think we should merge this PR until we get the words included for the missing example here too. I struggled with explaining the concept as part of this PR, and require some magic hands here to help us get the words going for the examples. I think maybe having this as a "note" section isn't ideal, and instead it should be it's own sub section of content in some capacity with the two examples living there, as the distinctions here are important to convey in the docs, as they are not easily conveyable in a live discussion with folks without detailed explanation, and some confusion.

gilesgas: Thanks @HugeIRL , yes that's a great point!
I'll work your examples above into this PR and will extract the information from the callout and rework this into its next revision.

HugeIRL: Hey @gilesgas just bumping this so it's not lost.

gilesgas: Sorry for my delay on this one @HugeIRL , this PR's definitely on my radar!
I'll try looking at this today and incorporate your example into this PR.

gilesgas: Hi @HugeIRL , I've finally got around to addressing this. Sorry it's taken a while.

I decided to explain everything around the first example you provided, and so I hope my interpretation is correct. Please let me know so I can approve this PR...

The revised content can be [previewed here](https://2807--bk-docs-preview.netlify.app/docs/pipelines/trigger-step#a-note-on-triggered-pipeline-dependencies-and-cancellation).

HugeIRL: Yeah I originally liked the note call out idea, but now that its getting larger, we should probably move it out.

An idea I would have is that the note/callout explains the high level overview of what we're describing, and then maybe under the callout we go into more detail?

gilesgas (approved):

gilesgas (approved):
Implemented feedback from @HugeIRL , who also agreed with the rewording.

This involved tweaking the intro a little.

PR #2814, opened 2024-05-29
Fix typo

pzeballos:
email is wrong. It should be support@buildkite.com


gilesgas (approved):
Whoops! Thanks @pzeballos !

PR #2826, opened 2024-06-06
removed the experiment highlight from the k8s implementation doc

LiamPClancy:
the usage of the k8s stack was been called out in the documentation as experimental and this is not the case. 
I have removed the experimental callout for the k8s service in the docs.


DrJosh9000 (approved):
Yep this bit is definitely wrong 🔪 🔪 🔪

DrJosh9000 (approved):
Even better!

mbelton-buildkite (approved):

gilesgas (approved):

PR #2831, opened 2024-06-12
Add details about API calls for creating pipelines requiring the appr…

gilesgas:
…opriate cluster ID value to be specified in the request.


tessereth (approved):
Thanks for adding this!

PR #2833, opened 2024-06-13
REST api docs: Update example json with correct agent web_url formats

l-suzuki:
I recently fixed a bug where we're displaying incorrect web_url or agents in the REST api. This PR updates the docs with the correct web_url format. UUIDs used were generated randomly.


zhming0 (approved):

gilesgas:
Thanks for this @l-suzuki !

PR #2839, opened 2024-06-19
Improve the layout of info on the 'Test ownership' page.

gilesgas:
Some of these include:

- Removing the FAQs and integrating their content into relevant page headings that provide instructions on how to do a task (or fulfil a use case).
- Moving the upload TESTOWNERS file to after the heading/section that explains the format of the file first.
- Moving content on how to obtain team slugs to the top of the TESTOWNERS file format heading/section.


gilesgas:
The following description of the first test file pattern in the TESTOWNERS file:

```
# The following example teams will be the test owners for all test
# location metadata (that is, test files) from your pipeline builds
# in this repository. While both these example teams own these
# tests, the first team specified in this file pattern becomes the
# default owner for all test files from your pipeline builds. Any
# file pattern matches defined later in this file take precedence
# and override any file patterns defined further up this file.
# Therefore, unless a subsequent file pattern match takes
# precedence, `team-slug-1` will have any flaky tests assigned to
# them by default.
*                     team-slug-1 team-slug-2
```

which I had modified to make a little clearer, still doesn't really explain the purpose behind specifying multiple teams for this (or any other) test file pattern.

The original FAQ:

> **Can more than one team own a test?**
> 
> More than one team may own a test, and the order of teams in your TESTOWNERS file is important. The first team listed will be the default owner, and they will be auto-assigned to the test if it flakes. Any team with suite access can override this auto-assignment.

Also doesn't really explain clearly why I would want to specify additional teams for a given test file pattern.

gilesgas:
Paraphrased from @jameshill regarding specifying multiple teams for a give test file pattern:

> Default team: this team will be automatically notified of issues with the test
> Other teams: can be identified as an owner, and therefore will appear in reports around the reliability of that test. However, these teams will not be notified

Will update the docs content accordingly.

jameshill (approved):
💚

PR #2848, opened 2024-06-21
Use IAM role to pull from public ecr

amu-g:
In order to avoid hitting rate limits. Unauthenticated pulls from public ECR are limited to 1/sec. Authenicated pulls are 10/sec. Hopfully this will be enough for us to avoid hitting the rate limit.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -106,7 +106,6 @@ steps:
           - ecr#v2.9.0:
               login: true
               account-ids: public.ecr.aws
-              region: us-east-1
           - docker#v3.7.0:
               image: "public.ecr.aws/docker/library/node:lts-alpine3.14"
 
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -101,6 +101,12 @@ steps:
         agents:
           queue: hosted
         plugins:
+          - aws-assume-role-with-web-identity#v1.0.0:
+              role-arn: arn:aws:iam::445615400570:role/pipeline-buildkite-docs
+          - ecr#v2.9.0:
+              login: true
+              account-ids: public.ecr.aws
+              region: us-east-1
```
yob, line 109: do we need to specify the region? Or will the plugin just always use `us-east-1` for the public gallery?

gilesgas (approved):
Not knowing the full details behind this PR, I'm assuming it will work (and I gather the passing pipeline is testament to that).

Therefore, I'm happy to approve this PR now although it would be great if @yob 's query could be answered on this PR too.

PR #2853, opened 2024-06-24
Add an information block for using step-level vs build-level notifications

mcncl:
## Added
A small information block on using step-level notifications and how they may differ to build-level.

Step-level notifications may act differently to how people assume, eg they can ignore branch filters set at a build-level.

### DIFF
--- a/pages/pipelines/notifications.md
+++ b/pages/pipelines/notifications.md
@@ -154,8 +154,8 @@ steps:
 ```
 {: codeblock-file="pipeline.yml"}
 
-> 📘Step-level vs build-level
-> A step-level notify step will ignore the requirements of a build-level notification. If a build-level notification condition is that it runs only on `main`, a step-level notification will run on all branches if no branch conditional is set.
+> 📘 Step-level vs build-level notifications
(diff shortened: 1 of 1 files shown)


pages/pipelines/notifications.md
```diff
@@ -154,6 +154,9 @@ steps:
 ```
 
+> 📘Step-level vs build-level
+> A step-level notify step will ignore the requirements of a build-level notification. If a build-level notification condition is that it runs only on `main`, a step-level notification will run on all branches if no branch conditional is set.
```
gilesgas, line 158: Some minor tweaks to the callout intro line (so far) ...

```suggestion
> 📘 Step-level vs build-level notifications
> A step-level notify step will ignore the requirements of a build-level notification. If a build-level notification condition is that it runs only on `main`, a step-level notification will run on all branches if no branch conditional is set.
```

> ... if no branch conditional is set

This clause sounds a little incomplete, perhaps because 'set' (in this context) is a transitive verb and hence, the clause's object is missing.

Does this mean '... if no branch conditional has been set on this step-level notification' ?

gilesgas (approved):

PR #2857, opened 2024-06-26
Add --testLocationInResults flag to Jest setup

malclocke:
When using Jest with the Javascript test collector this CLI flag is required so that test file line numbers are reported to Test Analytics.

### DIFF
--- a/pages/test_analytics/javascript_collectors.md
+++ b/pages/test_analytics/javascript_collectors.md
@@ -59,6 +59,7 @@ To configure Jest:
         "testLocationInResults": true,
     }
     ```
+    **Note:** The `"testLocationInResults": true` setting enables column and line capture for Test Analytics.
 
 ### Jasmine
 
@@ -70,7 +71,7 @@ To configure Jasmine:
     // SpecHelper.js
     var BuildkiteReporter = require("buildkite-test-collector/jasmine/reporter");
     var buildkiteReporter = new BuildkiteReporter();
-    
+
     jasmine.getEnv().addReporter(buildkiteReporter);
     ```
 
@@ -160,7 +161,7 @@ To configure Playwright:
     // playwright.config.js
     {
       "reporter": [
-        ["line"], 
+        ["line"],
         ["buildkite-test-collector/playwright/reporter"]
       ]
     }
@@ -201,4 +202,3 @@ Using the [`--forceExit`](https://jestjs.io/docs/cli#--forceexit) option when ru
 `--forceExit` could potentially terminate any ongoing processes that are attempting to send test executions to Buildkite.
 
 We recommend using [`--detectOpenHandles`](https://jestjs.io/docs/cli#--detectopenhandles) to track down open handles which are preventing Jest from exiting cleanly.
-
--- a/pages/test_analytics/javascript_collectors.md
+++ b/pages/test_analytics/javascript_collectors.md
@@ -55,7 +55,8 @@ To configure Jest:
 
     ```json
     {
-        "reporters": ["default", "buildkite-test-collector/jest/reporter"]
+        "reporters": ["default", "buildkite-test-collector/jest/reporter"],
+        "testLocationInResults": true,
     }
     ```
 


pages/test_analytics/javascript_collectors.md
```diff
@@ -58,6 +58,7 @@ To configure Jest:
         "reporters": ["default", "buildkite-test-collector/jest/reporter"]
     }
     ```
+1. Ensure that you run Jest with the `--testLocationInResults` flag.
```
gilesgas, line 61: Thanks for your contribution @malclocke 🎉 

Just a couple of suggestions ...

- Adding a space between before and after a code block, makes it more readable (as it stands out more) in the source.
- Since you're not instructing the reader to actually run Jest, but instead make a recommendation as part of running Jest, make this [a callout/note within a list](https://github.com/buildkite/docs/blob/main/styleguides/markdown-syntax-style.md#callouts-within-lists).

```suggestion

    **Note:** Ensure that you run Jest with the `--testLocationInResults` flag.
```

gilesgas: Alternatively, you could make this part of step 1:

```
1. Make sure Jest runs with:

    * Access to [CI environment variables](/docs/test-analytics/ci-environments).
    * The `--testLocationInResults` flag.

1. Add `"buildkite-test-collector/jest/reporter"` to ...
```

malclocke: @gilesgas As I've been following this up I've realised this can be set as a configuration file option on the current version of Jest.  The project I was originally having problems with was using an older version which only supported this option as a CLI flag.
In the current version (v29.7.0) of Jest this works both as a config file option and a CLI flag.

So we could recommend the following in the configuration file instead, which would be a little cleaner:

``` json
    {
        "reporters": ["default", "buildkite-test-collector/jest/reporter"],
        "testLocationInResults": true,
    }
```

We already recommend setting this up this way in the [Javascript test collector README](https://github.com/buildkite/test-collector-javascript?tab=readme-ov-file#jest) so the instructions should probably be consistent here.

However, this is not mentioned in the [Jest docs for the configuration file](https://jestjs.io/docs/configuration) (it does work though) and I'm not sure how to word the instructions for this.  WDYT?

pages/test_analytics/javascript_collectors.md
```diff
@@ -58,6 +58,7 @@ To configure Jest:
         "reporters": ["default", "buildkite-test-collector/jest/reporter"]
```
gilesgas, line 58: [Your suggestion](https://github.com/buildkite/docs/pull/2857/files#r1654093536) sounds even better then!

How about this then?

```suggestion
        "reporters": ["default", "buildkite-test-collector/jest/reporter"],
        "testLocationInResults": true
```

malclocke: OK, so we don't document the rationale for `"testLocationInResults": true`?

pages/test_analytics/javascript_collectors.md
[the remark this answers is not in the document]
malclocke: @gilesgas thank you, updated.  Back to you! :ping_pong:

pages/test_analytics/javascript_collectors.md
gilesgas: Sorry I missed seeing that in your previous comment @malclocke !

Yes I agree that that's important to mention it, and so I've committed this information in a note underneath this JSON code snippet. (I was trying to do that directly through this PR, but had trouble getting the GitHub editor to do this, so I had to commit and push up from a local change.)

pages/test_analytics/javascript_collectors.md
```diff
@@ -55,9 +55,11 @@ To configure Jest:
 
     ```json
     {
-        "reporters": ["default", "buildkite-test-collector/jest/reporter"]
+        "reporters": ["default", "buildkite-test-collector/jest/reporter"],
+        "testLocationInResults": true,
     }
     ```
+    **Note:** The `"testLocationInResults": true` setting enables column and line capture for Test Analytics.
```
gilesgas, line 62: Hopefully this note is OK?

Anyway, I'll approve this PR so that you can merge it in.

gilesgas (approved):
Ready for you to merge this in if it's OK with you @malclocke ?

PR #2858, opened 2024-06-27
Remove beta pills from Buildkite Packages docs.

gilesgas:
And remove beta callout near top of landing page.


pages/packages.md
mekenthompson, line 1: @gilesgas Can we leave the "You can enable through org settings in" but just remove reference to beta. For existing customers, they'll still have to enable it to get access, it still won't turn on by default until we resolve some fo the billing stuff in next 2-4 weeks.

gilesgas: Done!

mekenthompson (approved):
Looks great!

PR #2861, opened 2024-07-01
Bump to latest ruby 3.2

yob:
I noticed we're pinned to a version of ruby 3.2 that's a few months. It's unlikely there's changes in ruby 2.2.{3,4} that we need urgently, but bumping means we get a newer OS image with security packages preinstalled.

At some point we should try bumping to ruby 3.3 as well, but this seems like a worthwhile first step.


gilesgas (approved):
Looks good to me!

PR #2866, opened 2024-07-02
Add documentation for CLI v3

jradtilbrook:
This adds documentation for the CLI v3. It follows closely how the docs work for the agent as its a similar Go binary installed on users machines.

The information and configuration for the CLI is much smaller and mostly automated/guided with the user so has not been documented as thoroughly.

This adds a CLI section to the docs under "Agent". Only overview, install and configure for now. We'll consider something like command-line reference later, but most of the things are self discoverable and this is intended to be run by a human unlike the buildkite-agent so having that stuff in the docs seems less valuable.
Screenshot 2024-07-03 at 11 36 42


jradtilbrook:
I'm getting this error in the linting step and unsure what to do about it. I believe its saying "Configuration" should be "configuration", but that seems weird. It also isn't complaining about the similar file "Buildkite CLI Installation". @gilesgas do you know what to do here?

Screenshot 2024-07-03 at 11 34 03

gilesgas:
Thanks for this PR contribution @jradtilbrook !

First, I'm not sure why "Buildkite CLI Configuration" should be treated in Title Case, when our headings/page titles should use Sentence case (according to our [writing style guide](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#headings)). Therefore, this should be written as "Buildkite CLI configuration".

If, however, if you do have a term you want to use in a heading/page title that is a Proper Noun, and the linter throws an error like the one you see above, try adding it as an exception to [this file](https://github.com/buildkite/docs/blob/main/vale/styles/Buildkite/h1-h6_sentence_case.yml), preferably in alphabetical order.

pages/cli/v3/configuration.md
```diff
@@ -0,0 +1,16 @@
+# Buildkite CLI Configuration
```
gilesgas, line 1: Unless this needs to be Title Case, make this:

```suggestion
# Buildkite CLI configuration
```

pages/cli/v3/installation.md
```diff
@@ -0,0 +1,64 @@
+# Buildkite CLI Installation
```
gilesgas, line 1:
```suggestion
# Buildkite CLI installation
```

jradtilbrook:
Oh great no worries. My only issue was the confusion between both new pages. I assumed there was some allow list somewhere but didn't know. I'll update the headings now 👍

.github/labeler.yml
```diff
@@ -1,5 +1,7 @@
 agent:
   - pages/agent/**/*
+cli:
+  - pages/cli/**/*
```
jradtilbrook, line 4: I figured adding a corresponding label would be a good idea. But I don't know why its not adding it. Perhaps because the label doesn't exist yet?

gilesgas:
Hey @jradtilbrook , in case we haven't discussed this before, one thing we're intending to do is create a new top-level **Platform** category to the docs site (equivalent to **Pipelines**, **Test Analytics**, **Packages**) and in this **Platform** category, migrate into it most of the [Team management](https://buildkite.com/docs/team-management) content (with the exception of Pipeline permissions-related stuff) from the **Pipelines** docs.

This **Platform** docs category will cover aspects of Buildkite functionality that is Buildkite-system wide. The Buildkite CLI looks like a good candidate for this **Platform** category migration.

The **Integrations** content looks like it could be migrated into both the **Pipelines** and new **Platform** category too.

Out of interest, will it be necessary to keep docs for multiple versions of the Buildkite CLI, or could we just keep the one version in the docs and update it over time?

jradtilbrook:
Oh okay interesting @gilesgas that sounds reasonable to me 👍 
Does that mean it needs to change around now?

I think just the most recent version will be relevant to keep in the docs

gilesgas:
That's great, I'll remove the `v3` path then :-)

We only keep separate docs versions for the Buildkite Agent docs (i.e. a v3 and v2 of the Buildkite Agent), since even though the v2 Agent is deprecated, there are several customers still using it. If there haven't been many customers using previous versions of the Buildkite CLI tool (e.g. v2 and v1), and there's no need to maintain separate docs versions for them, then we can 'de-version' this part of the docs.

gilesgas:
FYI @jradtilbrook , I'm done with updating the **Configuration** page, which is the last update I wanted to do to this PR.

If you're happy with these changes, please merge in this PR, which I'll approve in the meantime.

gilesgas (approved):

jradtilbrook:
I've looked through the preview URL and it all looks good 🙌 thanks for your help @gilesgas

PR #2868, opened 2024-07-04
Added Packages to GraphQL cookbook, registries query example

james2791:
Added a new section and nav for Packages relative GraphQL queries and mutations

Currently filled in with obtaining package registries (`getOrganizationRegistries`).

Altered spec and nav_data script for ensuring this is expected in tests/generated upon running `rake graphql:generate`


pages/apis/graphql/cookbooks/packages.md
```diff
@@ -6,7 +6,7 @@ You can test out the Buildkite GraphQL API using the [Buildkite explorer](https:
 
 ## List organization registries
 
-List the first 50 package registries in the organization.
+List the first 50 registries in the organization.
```
gilesgas, line 9: Thanks very much for this contribution @james2791 ! 🎉

FYI, not all package formats supported by Buildkite Packages are called 'packages'. Hence, although the product is still called 'Buildkite Packages', we just refer to registries as 'registries' rather than 'package registries', which is a little too specific.

For example, [containers](https://buildkite.com/docs/packages/container) have 'images', and [Terraform](https://buildkite.com/docs/packages/terraform) has 'modules'.

Anyway, this looks pretty good to me and so I'll approve this now so that you can merge it in!

gilesgas (approved):

pages/apis/graphql/cookbooks/packages.md
james2791: Thats fair - has been amended 👍

PR #2880, opened 2024-07-10
Open external links in a new tab

mbelton-buildkite:
We currently style external links with an arrow by adding the the class `external-link`. This change extends that behavior to also add the target `_blank` so that external links open in a new tab.

Opening external links in a new tab helps users avoid losing the context they were in with Buildkite.

## Example

See the link on to Jenkins in the first sentence:

Before: [Current page](https://buildkite.com/docs/pipelines/migrate-from-jenkins)
After: [Preview page](https://2880--bk-docs-preview.netlify.app/docs/pipelines/migrate-from-jenkins)

### DIFF
--- a/app/models/page/renderers/external_link.rb
+++ b/app/models/page/renderers/external_link.rb
@@ -43,6 +43,7 @@ class Page::Renderers::ExternalLink
   def decorate_external_link_node
     unless node['class']
       node.set_attribute('class', 'external-link')
+      node.set_attribute('target', '_blank')
     end
   end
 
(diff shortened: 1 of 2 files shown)


gilesgas:
This looks great @mbelton-buildkite !
I've tested the external link behaviours on the before and after pages you linked to above, as well as some external links in the 'Background to packages' page of the Packages docs, and they work as expected.

gilesgas (approved):

BrettJay (approved):
Great uplift!

PR #2817, opened 2024-05-31
Add nonzero exit behavior to hook documentation

thejcannon:
From support Slack channel:

> So in general, non-zero codes in hooks are used to indicate that failure has occurred, and Buildkite respect this by failing the job or stopping further execution, depending on where the hook is used.
>
> For instance: Non-zero exit codes in pre-checkout, pre-command and pre- exit hooks can cause the assotiated job fail. 
>
> This means that if any of these hooks return a non-zero exit code, the job will not proceed to the next steps and will be marked as failed. (edited) 
>
> For the post-checkout or post-command, or post-exit hooks, typically don't affect the job success or failure directly but it can be used for logging or cleanup purposes.

So thought I'd PR putting this in the docs.

I don't like documenting the _negative_ (E.g. check mark implies job fails) so happy to hear alternative ways of succinctly documenting this.

PR #2883, opened 2024-07-11
Add description for hook failure behaviour.

gilesgas:
This PR supersedes this one: https://github.com/buildkite/docs/pull/2817


gilesgas:
Hi @thejcannon ,

Please feel free to review this PR.

pages/agent/v3/hooks.md
```diff
@@ -147,6 +147,11 @@ they are run as part of each job:
 | `pre-exit`      | <span class="add-icon-agent">Agent</span><br /><span class="add-icon-repository">Repository</span><br /><span class="add-icon-plugin">Plugin (non-vendored)</span><br /><span class="add-icon-plugin">Plugin (vendored)</span> | Runs before the job finishes. Useful for performing cleanup tasks. |
 {: class="table table--no-wrap"}
 
+> 📘 Hook failure behavior
+> If any hooks fail with a non-zero exit code, the job will be aborted at that point.
+> The behavior for most `post-*` hooks above, however, is a little different. If a `command` is the core of a job, the `command` job would have already run and completed before the `post-*` hook is run. Therefore, a `post-*` hook's success or failure has no impact on stopping the core `command` job from running.
+> The exception to this is the `post-checkout` hook, which runs after checkout but before the core `command` job runs. In such circumstances, if the `post-checkout` hook fails, then the `command` job will not run.
```
thejcannon, line 153: The wording here is weird, since you say "if any hooks fail ..." which sounds quite global. Then next sentence contradicts it.

This also gives a bit too much context in my opinion. Most people care about the what/how and not the _why_.

So maybe something like:
```suggestion
> For most hooks, if they fail with a non-zero exit code, the job will be aborted at that point.
> For `post-*` hooks (except for `post-checkout`), if they fail with a non-zero exit code, the job has no impact.
```

gilesgas: Thanks for your feedback @thejcannon .

I agree with you in re-writing this note/callout to front-load the _what/how_ explanation, since in this context, it's the most important explanation about 'Hook failure behavior'.
However, I would also argue that most people who visit documentation _are_ interested in the _why_ too, and that people consult documentation to look out for these explanations, so that they don't have to invest extra time in experimenting with a product to work out the 'why' through trial and error.

I will re-write this note/callout to front-load your suggestion, but I will retain the explanation for the benefit of other readers who'd like a more detailed explanation on why.

gilesgas: Here's a preview of the final result:
https://2883--bk-docs-preview.netlify.app/docs/agent/v3/hooks#hook-failure-behavior

pages/agent/v3/hooks.md
```diff
@@ -147,6 +147,14 @@ they are run as part of each job:
 | `pre-exit`      | <span class="add-icon-agent">Agent</span><br /><span class="add-icon-repository">Repository</span><br /><span class="add-icon-plugin">Plugin (non-vendored)</span><br /><span class="add-icon-plugin">Plugin (vendored)</span> | Runs before the job finishes. Useful for performing cleanup tasks. |
 {: class="table table--no-wrap"}
 
+Each `command` job defined in a pipeline's `pipeline.yml` file is independent and could execute concurrently. Therefore, a given hook will run for every one of these `command` jobs.
```
patrobinson, line 150:
```suggestion
Each `command` job defined in a pipeline's `pipeline.yml` file executes independently, therefore, a given hook will run for every one of these `command` jobs.
```

patrobinson: The concurrency thing might be confusing and not helpful

patrobinson (approved):

jradtilbrook (approved):

pages/agent/v3/hooks.md
```diff
@@ -147,6 +147,14 @@ they are run as part of each job:
...
 {: class="table table--no-wrap"}
 
+Each `command` job defined in a pipeline's `pipeline.yml` file executes independently, and therefore, a given hook will run for every one of these `command` jobs.
+
+When defining multiple commands in a step using the `commands` attribute, such as the `pipeline.yml` example described in [Command step attributes](/docs/pipelines/command-step#command-step-attributes), then each sub-command is concatenated with `&&` and effectively runs as a single job. Therefore, a given hook will only run once for a given `commands` job, consisting of multiple sub-commands.
+
+> 📘 Hook failure behavior
+> In the table above, if any of the hooks above `command` (from `pre-bootstrap` to `pre-command`, inclusive) fails with a non-zero exit code, then the command job will not run.
+> Since all the hooks below `command` (from `post-command` to `pre-exit`, inclusive) run after the command job, then any non-zero exit code failure in these hooks has no impact on the running of the command job.
```
moskyb, line 156: i think the wording here is making things a little bit confusing - we're referring to `command jobs` as the individual commands that get run during the command phase of a given job, but a "command job" is generally what we refer the whole thing - all the phases, all the hooks - as.

the core of this is that hooks after the command hook _will_ affect the command job, because they'll cause the job to fail with whatever status code they exit with. they won't, however, prevent the command itself from running - it's a bit of a nuanced difference, but it's important, i think.

moskyb (changes requested):

pages/agent/v3/hooks.md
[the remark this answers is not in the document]
moskyb: @patrobinson beat me to this one

pages/agent/v3/hooks.md
gilesgas: Aha! Thanks for explaining this subtlety! This is the distinction I was looking for before! However, prior to this, I did have trouble understanding exactly what constituted a 'command job'.

I'll re-write this now!

moskyb: my figma diagram that i made but decided not to post: the green box is the job, each purple box is a phase, and each reddish postit is a hook

moskyb: my figma diagram that i made but decided not to post: the green box is the job, each purple box is a phase, and each reddish postit is a hook

Untitled

gilesgas: Thanks @moskyb ! Very helpful.
Hopefully what I've re-written and commited now/since makes sense, but I'm now wondering (from your diagram), if a `post-command` fails, does this also cause the `command` phase to fail too?

moskyb: there's no real concept of a phase failing - they're more conceptual groupings than anything else. the job will fail if any hook returns a nonzero status code, but there's no intermediate "phase failure" that we record or measure.

gilesgas: So from a practical perspective @moskyb , is there anything a user can do with a successfully run '`command` phase' but with its subsequent hook execution failing, which would then cause the entire job to fail? I.e. is there anything they can do with they could do with the output of this '`command` phase' execution?

I'm just wanting to know if there's any practical value from a user's perspective in keeping the **Hook failure behavior** note/callout in the docs?

gilesgas: If the answer to my question above is 'no', then perhaps we should reword the note/callout to:

> 📘 Hook failure behavior
> If any of the hooks above fails with a non-zero exit code, the entire job will fail.

?

Leaving this for @mbelton-buildkite to finish off and merge in as I will be on leave for the next 3 weeks.

gilesgas: Sorry @moskyb , I was in a bit of a rush when I wrote my previous comment above and I missed the significance of what you'd written above - i.e.

> they won't, however, prevent the command itself from running - it's a bit of a nuanced difference, but it's important, i think.

Therefore, what I can see that what I'd written above:

> Any actions conducted by commands and scripts that ran successfully during the job's execution, prior to the point of its failure, will also be aborted.

is now incorrect.

Therefore, I'm committing an update to correctly clarify this callout, and also change it into a heading instead.

moskyb (approved):

pages/agent/v3/hooks.md
gilesgas: Thanks heaps for confirming the updates @moskyb !
cc @mbelton-buildkite

PR #2884, opened 2024-07-12
Add test splitting setup page to the docs.

wooly:
We have recently released a test splitting client called test-splitter. The current documentation for the test-splitter is in-tool on a page in the Test Suites section of the app, making it hard for people to discover the configuration instructions.

This PR adds a page to the public docs site with instructions on how to configure test-splitter in your pipeline.


gilesgas:
I've started reviewing this PR @wooly , and have so far tweaked the intro paragraph. I hope this still makes sense? Please correct anything that isn't quite correct.

In the meantime, what is the difference between the terms 'test splitting' and 'orchestration'?
To avoid customer confusion, it'd be great to stick to a just single term in the docs (and the product UI and elsewhere) for a (i.e. any) given product concept.

pages/test_analytics/test_splitting.md
```diff
@@ -0,0 +1,65 @@
+# Configuring test splitting
+
+Test splitting is the process of partitioning test suites to run in parallel across multiple Buildkite agents. Buildkite maintains the open source [test-splitter](https://github.com/buildkite/test-splitter) tool. This tool uses your Buildkite Test Analytics test suite data to intelligently partition tests throughout your test suites into multiple sets, such that each set of tests runs in parallel across your agents. This process is known as _orchestration_. Currently, the test-splitter tool only supports RSpec.
+
+## Dependencies
+
+Test splitting relies on execution timing data captured by the Buildkite test collectors to partition your tests evenly, so you will need to configure the [Ruby test collector](./ruby-collectors) for your test suite.
```
gilesgas, line 7: Is the execution timing data obtained from previous builds?

wooly: It is yeah

wooly:
> what is the difference between the terms 'test splitting' and 'orchestration'?

I think they're probably different terms for the same thing. I have no strong feelings on which one we lean towards. Might be a better question for @niceking and @jameshill?

niceking:
Hmmm great question @gilesgas, and I'll give you my 2c but I think @jameshill might have some further thoughts too. 

In my mind, "Test Splitting" is one aspect of "Orchestration". 

Orchestration is a general collection of tools and features that help customers orchestrate their test suite. So far we have the ability to split it up, but in the future we're planning on adding test selection, test quarantine, test burn-in etc, providing a richer set of tools that a customer can use to determine the execution of their test suite. 

I guess because we've only built the single Orchestration feature so far of Test Splitting, they're currently quite synonymous with each other. I'd lean towards "Orchestration" as the product concept though if we're thinking about how to communicate that externally!

gilesgas (approved):
Sorry for my delay on this @wooly !

This is now ready for you to merge in.

I have made some edits which you might like to scan over before merging this in.

Other than some minor wording tweaks and simplifying 'Buildkite build' to just 'build', I thought it important to add a definition (near the top of the page) to explain what a _test plan_ is, because it's not immediately obvious to users what this is.

PR #2892, opened 2024-07-17
Improve build id documentation

patrobinson:
We had an escalation from a customer asking:

> we're trying to understand what happened with one of our jobs where it seems to have skipped a build number.

We made a change to our system some time ago which means we can no longer guarantee a gapless sequence of build IDs.

This change updates the documentation to replace the language such as  "sequence" and "This number increases by 1 with every build" with "monotonically increasing" to more accurately reflect the current state.

I had to take a side quest to get the GraphQL schema docs updating.


gilesgas (approved):
LGTM! 🎉

PR #2894, opened 2024-07-17
Fix bad link

thejcannon:


mbelton-buildkite (approved):
Thanks for catching this and fixing it, @thejcannon! I'll run the linting build and merge for you when it passes. 😊

PR #2898, opened 2024-07-18
make it clear that agent config is a dir not a file

pbuckley:


pbuckley:
Just some more context, I was talking to someone who used our docs and thought this area could use more clarification that the param takes a directory path not a file path. Trying with a file didn't work and caused some delay of checking "is my config file alright? permissions? on the host? etc.

mbelton-buildkite (approved):
Thanks for making this clearer!

PR #2905, opened 2024-07-22
remove fragment from when we didn't offer custom images

pbuckley:
maybe this was just a leftover from a prior cleanup? it doesn't seem to fit in the doc anymore


gilesgas (approved):
Whoops! Thanks for picking up on this @pbuckley !

The config for hosted agents is now done through the UI.

I've approved this PR, so please feel free to merge this in when convient...

PR #2909, opened 2024-07-26
Messaging around possible differences in state expectations

mcncl:
## Why
The possible build states are not the states that will be returned by `state` in the API.

Eg, a blocked build could return `state` as `passed`, but would return `blocked` as `true`.


pages/apis/rest_api/builds.md
```diff
@@ -338,6 +338,8 @@ curl -H "Authorization: Bearer $TOKEN" \
 > 📘 Webhook URL
 > The response only includes a webhook URL in `pipeline.provider.webhook_url` if the user has edit permissions for the pipeline. Otherwise, the field returns with an empty string.
 
+The `state` of a build will not return `blocked`, as is the case with [notifications](/docs/pipelines/notifications#build-states). Instead, it will return `true` for the value of `blocked`.
```
gilesgas, line 341: Thanks @mcncl for this contribution!

Two things pushed up directly:

- Remove the callout formatting as the web renderer doesn't support 2 consecutive callouts together.
- Start a new sentence from 'Instead...'.

Query, what is 'it' that returns a value of `true`, since `state` doesn't appear to have a valid value of `true`?

Using the word 'it' for the subject of a sentence is inherently ambiguous, unless it relates directly to the object of the previous sentence or clause, when the meaning of 'it' in such contexts is more obvious.

gilesgas (changes requested):
Thanks @mcncl for this contribution!

Two things pushed up directly:

- Remove the callout formatting as the web renderer for the docs site currently doesn't support 2 consecutive callouts together.
- Start a new sentence from 'Instead...'.

My main question is, in:

> Instead, it will return `true` for the value of `blocked`.

what is 'it' that returns a value of `true`, since `state` doesn't appear to have a valid value of `true`?

In general, using the word 'it' for the subject of a sentence (or clause) is inherently ambiguous, unless the 'it' relates more obviously to the object of the previous sentence/clause. When I read the initial sentence, I implied that 'it' referred to `state`, which I don't think is correct.

gilesgas:
Ah! I should ready your PR descriptions before launching into a diatribe!

You mean `blocked` 🤦 which is kinda what I suspected. I'll amend your PR

gilesgas:
Hi again @mcncl ,

In order to improve the discoverability of this information, I've added this info to the existing callout in the reused descriptions throughout other relevant areas of the docs (i.e. the [**Notifications**](https://buildkite.com/docs/pipelines/notifications) and [**Defining steps**](https://buildkite.com/docs/pipelines/defining-steps) pages within the **Configuring pipelines** section of the **Pipelines** docs), since there are links to these sections where surfacing this idiosyncratic behaviour would be useful.

Anyway, I shall approve and merge this PR now, but if you have any concerns about these additional changes, please do let me know!

gilesgas (approved):

PR #2914, opened 2024-07-29
Update google_workspace_saml.md

karensawrey:
There are two "Step 3" in there right now. Adding Step 4.


mbelton-buildkite:
Do you mind fixing the next heading, too? It looks like this takes us from having a double step 3 to a double step 4

karensawrey:
Done and done, @mbelton-buildkite! Let's :shipit:!

mbelton-buildkite (approved):

PR #2916, opened 2024-07-30
Updating PATCH /pipelines REST API Documentation to Include Optional `slug` Parameter 

123sarahj123:
### Description
Implementation to expose an optional slug parameter to `PATCH /pipelines`

This PR is to update the REST API documentation to include the optional `slug` paramenter for `docs/apis/rest-api/pipelines` for updating properties of an existing pipeline. 

A user can now update the pipeline `slug` name using `PATCH /pipelines` when they provide a `slug` parameter. 

It is important to note that if a user updates the `name` of a pipeline _without_ providing a `slug` parameter, a new `slug` will be generated for that pipeline that is derived from the updated name. If a user updates the `name` and forgot to add the `slug` parameter (because they wanted to preserve their pipeline `slug`) they can update the `slug` name in isolation.

Under "Update a pipeline" in the "Optional request body properties" table
image

image

### DIFF
--- a/pages/apis/rest_api/pipelines.md
+++ b/pages/apis/rest_api/pipelines.md
@@ -876,7 +876,10 @@ Optional [request body properties](/docs/api#request-body-properties):
   </tr>
   <tr>
     <th><code>name</code></th>
-    <td>The name of the pipeline. If you provide a new name without a <code>slug</code> parameter, the slug will be automatically updated to match the new name.<p class="Docs__api-param-eg"><em>Example:</em> <code>"New Pipeline"</code></p></td>
+    <td>
+      <p>The name of the pipeline. If you provide a new name without a <code>slug</code> parameter, the slug will be automatically updated to match the new name.</p>
+      <p class="Docs__api-param-eg"><em>Example:</em> <code>"New Pipeline"</code></p>
+    </td>
   </tr>
   <tr>
     <th><code>pipeline_template_uuid</code></th>
@@ -900,7 +903,9 @@ Optional [request body properties](/docs/api#request-body-properties):
   </tr>
   <tr>
     <th><code>slug</code></th>
-    <td>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. If not provided when the pipeline name is updated, the slug will be automatically generated from the new pipeline name. If the pipeline name is not updated, the existing slug will remain unchanged.
+    <td>
+      <p>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. If not provided when the pipeline name is updated, the slug will be automatically generated from the new pipeline name. If the pipeline name is not updated, the existing slug will remain unchanged.<br>
+      The slug may only contain alphanumeric characters or dashes, and cannot begin with a dash.</p>
       <p><em>Example:</em> <code>"my-custom-pipeline-slug"</code></p>
     </td>
   </tr>
--- a/pages/apis/rest_api/pipelines.md
+++ b/pages/apis/rest_api/pipelines.md
@@ -904,8 +904,8 @@ Optional [request body properties](/docs/api#request-body-properties):
   <tr>
     <th><code>slug</code></th>
     <td>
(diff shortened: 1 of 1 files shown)


pages/apis/rest_api/pipelines.md
```diff
@@ -876,7 +876,7 @@ Optional [request body properties](/docs/api#request-body-properties):
   </tr>
   <tr>
     <th><code>name</code></th>
-    <td>The name of the pipeline.<p class="Docs__api-param-eg"><em>Example:</em> <code>"New Pipeline"</code></p></td>
+    <td>The name of the pipeline. If you provide a new name without a <code>slug</code> parameter, the slug will be automatically updated to match the new name.<p class="Docs__api-param-eg"><em>Example:</em> <code>"New Pipeline"</code></p></td>
```
mhaylock, line 879: It seems to be inconsistent whether or not we wrap these cells in a paragraph, I suggest we do so with these changes:
```suggestion
    <td>
      <p>The name of the pipeline. If you provide a new name without a <code>slug</code> parameter, the slug will be automatically updated to match the new name.</p>
      <p class="Docs__api-param-eg"><em>Example:</em> <code>"New Pipeline"</code></p>
    </td>
```

pages/apis/rest_api/pipelines.md
```diff
@@ -898,6 +898,12 @@ Optional [request body properties](/docs/api#request-body-properties):
     <th><code>skip_queued_branch_builds_filter</code></th>
     <td>A <a href="/docs/pipelines/branch-configuration#branch-pattern-examples">branch filter pattern</a> to limit which branches intermediate build skipping applies to. <p class="Docs__api-param-eg"><em>Example:</em> <code>"!main"</code><br><em>Default:</em> <code>null</code></p></td>
   </tr>
+  <tr>
+    <th><code>slug</code></th>
+    <td>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. If not provided when the pipeline name is updated, the slug will be automatically generated from the new pipeline name. If the pipeline name is not updated, the existing slug will remain unchanged.
+      <p><em>Example:</em> <code>"my-custom-pipeline-slug"</code></p>
+    </td>
```
mhaylock, line 905: Suggesting wrapping in a paragraph and adding a line about the required format of slugs:
```suggestion
    <td>
      <p>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. If not provided when the pipeline name is updated, the slug will be automatically generated from the new pipeline name. If the pipeline name is not updated, the existing slug will remain unchanged.<br>
      The slug may only contain alphanumeric characters or dashes, and cannot begin with a dash.</p>
      <p><em>Example:</em> <code>"my-custom-pipeline-slug"</code></p>
    </td>
```

mhaylock (commented):
Making a couple of minor suggestions

pages/apis/rest_api/pipelines.md
```diff
@@ -898,6 +901,14 @@ Optional [request body properties](/docs/api#request-body-properties):
     <th><code>skip_queued_branch_builds_filter</code></th>
     <td>A <a href="/docs/pipelines/branch-configuration#branch-pattern-examples">branch filter pattern</a> to limit which branches intermediate build skipping applies to. <p class="Docs__api-param-eg"><em>Example:</em> <code>"!main"</code><br><em>Default:</em> <code>null</code></p></td>
   </tr>
+  <tr>
+    <th><code>slug</code></th>
+    <td>
+      <p>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. If not provided when the pipeline name is updated, the slug will be automatically generated from the new pipeline name. If the pipeline name is not updated, the existing slug will remain unchanged.<br>
+      The slug may only contain alphanumeric characters or dashes, and cannot begin with a dash.</p>
```
mbelton-buildkite, line 908: I think this could be a little clearer. What do you think of rearranging some of the info like so:

```suggestion
      <p>A custom identifier for the pipeline. This slug will be used as the pipeline's URL path. It can only contain alphanumeric characters or dashes and cannot begin with a dash.<br>
      The slug updates whenever the pipeline name changes. If you don't provide a slug when you update the pipeline name, the slug will be automatically generated from the new pipeline name.</p>
```

123sarahj123: Thank you :) I have committed your suggestion.

mbelton-buildkite (approved):
Looks good to me :)

PR #2935, opened 2024-08-16
Fix invalid Test Event Payload

turingbeing:
The existing Payload for `Build Started` is invalid and fails Eventbridge Test Pattern checks. 

The proposed update is valid.

PR #2942, opened 2024-08-21
Add standard eventbridge fields to the example payloads

amu-g:
The example payloads for eventbridge events only showed fields that buildkite varies. As raised in https://github.com/buildkite/docs/pull/2935, copy and pasting this into the eventbridge sandbox as-is results in an error message that the sample event is not valid.
Putting the full payload in the examples will make it easier for users to test patterns using the eventbridge sandbox.

### DIFF
--- a/pages/integrations/amazon_eventbridge.md
+++ b/pages/integrations/amazon_eventbridge.md
@@ -171,7 +171,14 @@ AWS EventBridge has strict limits on the size of the payload as documented in [A
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


amu-g:
yes, it does @gilesgas

gilesgas (approved):
Thanks for confirming @amu-g !

Ready for you to merge this in :-) (I just pushed up a fix to one of the headings on the page.)

PR #2944, opened 2024-08-23
Review of new Files and Helm registry pages.

gilesgas:
This PR provides a PR review for the newish Files and Helm registry pages.

This PR also adds newer features, which includes:

- The 'single command' (with short-lived token) publish option for each registry type.


gilesgas:
This is being done as part of another PR, hence closing this PR.

PR #2951, opened 2024-08-29
Ignore CVE-2024-37371 and CVE-2024-37370 in docker image

yob:
These CVEs are in krb5, a library that handles kerbeos authentication. We don't any kerbeos in production, but also the issues are  fixed in 1.20.1-2+deb12u2 which is available in the debian repos and the docker build logs show is being installed. This seems to be a false positive by ECR


gilesgas (approved):
LGTM

PR #2961, opened 2024-09-06
Job expiry process delays

lizrabuya:
A customer has asked why there was delays in expiring jobs in their org. After clarification from the pipelines team, it would be good to have this information available for the customers to know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,5 @@
   "dependencies": {
     "@docsearch/js": "3",
     "@hotwired/turbo-rails": "^7.3.0"
-  },
-  "version": "0.0.0"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -21,5 +21,6 @@
   "dependencies": {
     "@docsearch/js": "3",
     "@hotwired/turbo-rails": "^7.3.0"
-  }
+  },
+  "version": "0.0.0"
 }
--- a/pages/pipelines/build_timeouts.md
+++ b/pages/pipelines/build_timeouts.md
@@ -40,4 +40,4 @@ You can override the default by setting a shorter value in your organization's [
 Scheduled job limits should not be confused with [scheduled builds](/docs/pipelines/scheduled-builds). A scheduled build's jobs will still go through the [build states](/docs/pipelines/defining-steps#build-states), and the timeout will apply once its individual jobs are in the scheduled state waiting for agents.
 
 > 📘 Delays in expiring a job
-> A job's expiry process is run hourly at 5 minutes past. When the expiry process runs and the job's scheduled expiry was not over at that hour, it will only be expired until the next hour when the process is executed again. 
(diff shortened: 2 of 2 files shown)


gilesgas:
Thanks for your contribution @lizrabuya ! 🎉

Just had a quick question or two...

package.json
```diff
@@ -21,5 +21,6 @@
   "dependencies": {
     "@docsearch/js": "3",
     "@hotwired/turbo-rails": "^7.3.0"
-  }
+  },
+  "version": "0.0.0"
```
gilesgas, line 25: What's the purpose of this `package.json`? If this was unintentional, could it please be removed?

pages/pipelines/build_timeouts.md
```diff
@@ -38,3 +38,6 @@ By default, jobs are canceled when not picked up for 30 days. This will cause th
 You can override the default by setting a shorter value in your organization's [**Pipeline Settings**](https://buildkite.com/organizations/~/pipeline-settings) page.
 
 Scheduled job limits should not be confused with [scheduled builds](/docs/pipelines/scheduled-builds). A scheduled build's jobs will still go through the [build states](/docs/pipelines/defining-steps#build-states), and the timeout will apply once its individual jobs are in the scheduled state waiting for agents.
+
+> 📘 Delays in expiring a job
+> A job's expiry process is run hourly at 5 minutes past. When the expiry process runs and the job's scheduled expiry was not over at that hour, it will only be expired until the next hour when the process is executed again.
```
gilesgas, line 43: In [US English](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#english-flavor), the word 'expiration' is used instead of 'expiry'.

Since we use US English throughout our docs as part of our styleguide, then:

```suggestion
> A job's expiration process is run hourly at 5 minutes past. When the expiration process runs and the job's scheduled expiration was not over at that hour, it will only be expired until the next hour when the process is executed again.
```

Also, since I can't easily do this through this PR, could you please update the other 4 remaining 'expiry' references on this page to 'expiration' please?
There are no internal docs links to the 'Scheduled job expiry' heading, and therefore, changing this to 'Scheduled job expiration' should be fine.

gilesgas, line 43: In US English, the word 'expiration' is used instead of 'expiry'.

Since we use [US English throughout our docs as part of our styleguide](https://github.com/buildkite/docs/blob/main/styleguides/writing-style.md#english-flavor), then:

```suggestion
> 📘 Delays in job expiration
> A job's expiration process is run hourly at 5 minutes past. When the expiration process runs and the job's scheduled expiration was not over at that hour, it will only be expired until the next hour when the process is executed again.
```

Also, since I can't easily do this through this PR, could you please update the other 4 remaining 'expiry' references on this page to 'expiration' please?
There are no internal docs links to the 'Scheduled job expiry' heading, and therefore, changing this to 'Scheduled job expiration' should be fine.

package.json
lizrabuya: this change is unintentional and should not be part of the PR

gilesgas:
Remove package.json thing before merging.

PR #2964, opened 2024-09-09
chore: add git mirror copy

dimiexplorer:
- Add copy around new Git mirror caching for mac
- Update mac agent image software list


pages/pipelines/hosted_agents/mac.md
```diff
@@ -149,3 +152,32 @@ Updated Xcode versions will be available one week after Apple offers them for do
 
 - libpq
 - GMP
+
+## Git mirror cache
+
+The Git mirror cache is a specialized type of cache volume designed to accelerate Git operation by caching the Git repository between builds. This is useful for large repositories that are slow to clone. These volumes are attached on a best-effort basis depending on their locality, expiration and current usage, and therefore, should not be relied upon as durable data storage. By default, Git mirror cache is scoped to a pipeline and is shared between all steps in the pipeline.
```
matthewborden, line 158:
```suggestion
The Git mirror cache is a specialized type of cache volume designed to accelerate Git operation by caching the Git repository between builds. This is useful for large repositories that are slow to clone. These volumes are attached on a best-effort basis depending on their locality, expiration and current usage, and therefore, should not be relied upon as durable data storage. By default, Git mirror cache is per respository and is shared between all steps in the pipeline.
```

gilesgas:
Thanks very much for preparing this PR @dimiexplorer !

[Like the Linux-hosted agents](https://buildkite.com/docs/pipelines/hosted-agents/linux#cache-volumes-container-cache), it looks like container caching is also available for Mac too. Is that right and if so, we should probably add a small section to this Mac page about this feature (for consistency).

gilesgas (approved):

PR #2965, opened 2024-09-11
Standalone Deploy Pipeline

matthewborden:
Currently the deploy pipeline and the testing pipeline are uploaded within the same build. This means there is a dependency on the rspec tests passing before the deployment proceeds. I'd like to remove this dependency so we can move the deploy steps to their own pipeline. In this pull request, I've created a copy of the deployment pipeline that can be run in parallel to the existing pipeline, with a view to removing the duplication with urgency.


gilesgas:
Hi @matthewborden , is this PR still valid?

PR #2968, opened 2024-09-11
[PKG-7403] - API docs for registries create

ensvo:
image
image

### DIFF
--- a/pages/apis/rest_api/packages/registries.md
+++ b/pages/apis/rest_api/packages/registries.md
@@ -48,7 +48,7 @@ Optional [request body properties](/docs/api#request-body-properties):
   </tbody>
 </table>
 
-Required scope: `create_registries`
+Required scope: `write_registries`
 
 Success response: `200 OK`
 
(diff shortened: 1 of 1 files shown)


pages/apis/rest_api/packages/registries.md
```diff
@@ -2,6 +2,56 @@
...
+
+Optional [request body properties](/docs/api#request-body-properties):
+
+<table class="responsive-table">
+  <tbody>
+    <tr><th><code>description</code></th><td>Description of the registry.<br><em>Default value:</em> <code>null</code>.</td></tr>
+  </tbody>
+</table>
+
+Required scope: `create_registries`
```
swebb, line 51: `create_registries` doesn't exist. I think this should be [write_registries](https://github.com/buildkite/buildkite/blob/6577aa6bf6e59191bf00a0a6495205a392406a36/app/models/api/access_token/scopes.rb#L48).

swebb (approved):
One change but basically looks good 👍🏻

pages/apis/rest_api/packages/registries.md
ensvo: Thank you for catching this!

gilesgas (approved):

PR #2971, opened 2024-09-11
`getRules` GraphQL Rules cookbook indentation

james2791:
Noticed that the `getRules` GraphQL cookbook had a slight indentation added to the example - which I've brought it back to be in-line with the rest of the examples.


gilesgas:
Thanks for picking this up @james2791 !

This is ready to be merged in...

gilesgas (approved):

PR #2972, opened 2024-09-11
Fix --spawn tag not in code tag

JuanitoFatas:
Simple fix for the missing code tag wrapped for `--spawn` in below screenshot

CleanShot 2024-09-12 at 08 08 56@2x


gilesgas:
Thanks @JuanitoFatas , but I doubt this fix will last long, as this page's content is automatically re-generated from the [relevant agent source code](https://github.com/buildkite/agent/blob/main/clicommand/agent_start.go#L331), which converts the forward ticks (in the agent source code) to backticks in the resulting Markdown/HTML of the docs.

We can't really change this to HTML tags in the agent source because these descriptions are used in the command line help options (i.e. `buildkite-agent start --help`), where these tags are likely to be rendered literally as is.

It might be worthwhile investigating [if the relevant script](https://github.com/buildkite/docs/blob/main/scripts/cli2md.rb#L17) that does this conversion for the docs can change the pairs of forward ticks to `<code></code>` tags in the resulting Markdown (for the HTML).

JuanitoFatas:
@gilesgas You’re right I fixed in the wrong place 🤦🏻‍♂️

gilesgas:
No worries @JuanitoFatas , this can't be fixed until the relevant `cli2md.rb` script has been updated to handle forward ticks, which I also just discovered are called 'prime' symbols. Thanks again for caring!

PR #3000, opened 2024-10-28
fix(tablet): set rows grid to 0 if it's empty (e.g. no toc section)

hannahcancode:
- on tablet, the screen is split into 4 grid areas: if there's no automatic TOC generated, which takes up the top right corner, and the content doesn't fill up the screen, the content ends up in the bottom left corner. If we add a `minmax` to the grid rows, with a min of 0 and a max of auto, it should sort itself out better.

**After:**
Screenshot 2024-10-28 at 1 53 18 PM

**Before** Reported by @zhming0:
Screenshot 2024-10-28 at 1 23 23 PM

### DIFF
--- a/app/frontend/css/components/_page.scss
+++ b/app/frontend/css/components/_page.scss
@@ -10,7 +10,7 @@ $lg-screen-margin-top: map-get($header-height, "one-row");
 
 .Page {
   display: grid;
-  grid-template-rows: auto;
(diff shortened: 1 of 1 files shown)


buildkite-systems:
Preview URL: https://f54a71181fe66bb4df3e9e5018b8ede93815833f--bk-docs-preview.netlify.app

CoreyGinnivan (approved):

PR #3005, opened 2024-11-13
Add documentation for Slack Workspace

eleanorakh:
We're in the process of rolling out some changes to the Slack Integration. This PR addresses:

- [ ] Adding new Slack Workspace integration
- [x] User @ mentions
- [x] Notifications in private channels
- [x] Adding Slack privacy policy as per their required guidelines
- [x] New failing state 
- [x] Add screenshots


pages/integrations/slack.md
```diff
@@ -7,7 +7,21 @@ Configuring a Slack notification service will authorize access for your desired
 
 Setting up a Notification Service requires Buildkite organization admin access.
 
-## Adding a notification service
+## Adding notification services
+There are two ways to add notification services: workspace-wide and channel-specific. We reccomend using a workspace. Using a workspace only requires a single authorization for the entire workspace, allowing you to connect channels via YAML rather than multiple notifications services. The migration to a Workspace will not impact current integration implementation.
```
dannymidnight, line 11: I reckon it's also worth calling out that the workspace integration adds support for Slack mentions, specifically because we request an additional scope to read user email addresses.

dannymidnight, line 11:
```suggestion
There are two ways to add notification services: workspace-wide and channel-specific. We recommend using a workspace. Using a workspace only requires a single authorization for the entire workspace, allowing you to connect channels via YAML rather than multiple notifications services. The migration to a Workspace will not impact current integration implementation.
```

dannymidnight (commented):
The following is no longer true with the new method:

https://github.com/buildkite/docs/blob/4d7c39ac2c720aa0c02b99fc00d707935d6bd4c7/pages/integrations/slack.md?plain=1#L6

I wonder if it's worth breaking this whole integration into a separate page? 

I also wonder if we should remove the section on "Legacy slack" now that we've added yet another version!

PR #3010, opened 2024-11-19
Document `text` field in InputStep:TextInput

thejcannon:
The examples show the field, but it is undocumented.

Description/Example ripped from https://github.com/buildkite/pipeline-schema


gilesgas:
Thanks very much for your contribution @thejcannon !

We'll get this one merged in soon

PR #3018, opened 2024-12-17
Fix incorrect test runner env var name

malclocke:
Currently documented ENV var name [is incorrect](https://github.com/buildkite/test-engine-client/blob/3f045ed4a69a569913b007ca3aecc96a0c541e00/internal/config/env.go#L52).


buildkite-systems:
Preview URL: https://1e7e13597bbcf9cbe4c7ae5e0066d658cdbbe724--bk-docs-preview.netlify.app

nprizal (approved):
Good catch! Thanks for fixing this.

PR #3026, opened 2025-01-08
Make `cluster_id` just required (not optional as well)

thejcannon:
This property is listed in both required/optional.


gilesgas:
Thanks very much again @thejcannon ! 🎉
This is great and much appreciated :-)

PR #3027, opened 2025-01-08
`cluster_id` is the UUID ID, not the b64 encoded one

thejcannon:
(title)


gilesgas:
Many thanks again @thejcannon !

PR #3063, opened 2025-06-11
PIPE-1119: tiny fix on k8s stack installation instruction

zhming0:
This is a 1 LoC change 🙂 

Context: PIPE-1119

### DIFF
--- a/pages/agent/v3/agent_stack_k8s/installation.md
+++ b/pages/agent/v3/agent_stack_k8s/installation.md
@@ -58,6 +58,7 @@ If you prefer to self-manage a Kubernetes Secret containing the agent token inst
 Here is how a custom secret can be created:
 
 ```bash
+kubectl create namespace buildkite
 kubectl create secret generic <kubernetes-secret-name> -n buildkite \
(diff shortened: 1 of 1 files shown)


sj26 (approved):
Nice!

Separate wondering: I wonder if it's worth omitting the namespace from our instructions. Would folks generally expect them to be present, or left as an exercise to the reader?

DrJosh9000:
> Would folks generally expect them to be present, or left as an exercise to the reader?

That's a good question. 

I encountered namespaces fairly early on when learning Kubernetes, but I think that was more because our install instructions included namespace setup, and I kept forgetting the namespace flag, so e.g. `kubectl` commands would not work with namespaced things.

It's called out in the Kubernetes docs as generally unnecessary to bother with namespaces. But for things like our stack, which can create zillions of jobs, I can imagine it would be easier to clean up a deployment gone wrong if all the jobs are confined to a namespace.

zhming0:
I am supposed to create this in the private doc repo. closing this down and re-creating it in private

PR #3086, opened 2025-10-30
fix: specify that the average build stat uses default branch

liamstevens:
This PR adds a small clarification about the "Builds per week" measurement.

This is something that one of our engineers ran across when doing some investigations and was unsure about until we inspected the number of builds actually run ourselves. 

Feel free to reject this PR if this is documented elsewhere. I haven't found it though!

### DIFF
--- a/pages/pipelines/dashboard_walkthrough.md
+++ b/pages/pipelines/dashboard_walkthrough.md
@@ -30,7 +30,7 @@ The reliability of your pipeline is a calculation based on passing vs failing bu
 
 ## Builds per week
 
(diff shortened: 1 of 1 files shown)


gilesgas:
Thanks very much for your contribution @liamstevens ! 🎉

I'm just getting this double-checked at the moment and if all's good, will get this merged in within the next few hours.

gilesgas:
FYI @liamstevens , thanks again for this contribution, which has been accepted and will be merged in.

I also added a note to indicate that if the pipeline's default branch setting is left blank (i.e. set to `None`), then the metric is calculated on all branches of the repository.

PR #3094, opened 2025-11-21
Fix Terraform module reference link

jkburges:
The previous link was 404ing.


gilesgas:
Thanks very much @jkburges for picking up on this! Much appreciated and we'll integrate your fix now :-)

PR #3095, opened 2025-11-21
adding exception for Hosted Agents

quinn-diesel:
We found out that hosted agents doesn't respect the priority list.

We should get updated in the docs so customers know for the future.

I've poorly written a line about this but wanted to get it directly into the docs.

This is the linear with more information
https://linear.app/buildkite/issue/MDC-776/hosted-agents-not-respecting-job-priority-in-buildkite

### DIFF
--- a/pages/pipelines/configure/workflows/managing_priorities.md
+++ b/pages/pipelines/configure/workflows/managing_priorities.md
@@ -2,7 +2,7 @@
 
 By default, jobs are dispatched (taken from the queue and assigned to an agent) on a first-in-first-out basis. However, job priority and pipeline upload time can affect that order.
 
-This is not the case for Hosted Agents. Jobs are assigned and run based on the time they are triggered.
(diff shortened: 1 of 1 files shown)


buildkite-systems:
Preview URL: https://c1377b458f30eaa93da4b77f956e0d8650820009--bk-docs-preview.netlify.app

pages/pipelines/configure/workflows/managing_priorities.md
```diff
@@ -2,6 +2,8 @@
 
 By default, jobs are dispatched (taken from the queue and assigned to an agent) on a first-in-first-out basis. However, job priority and pipeline upload time can affect that order.
 
+This is not the case for Hosted Agents. Jobs are assigned and run based on the time they are triggered.
```
gilesgas: Add a link to the Buildkite hosted agents overview page + reword slightly.

```suggestion
This is not the case for [Buildkite hosted agents](/docs/pipelines/hosted-agents), where jobs are assigned and dispatched at the time they are executed.
```

PR #3126, opened 2026-04-27
Define Step Uploads action before talking about it

epssy:
As a newbie to the Buildkite platform I got tripped up by the sudden appearance of the Step Upload concept with no context.  Further up the page there is some discussion of uploads though it isn't clear how the two are related.  It was doubly confusing as immediately prior to its usage in the warning was an exhaustive enumeration of Step Types that did not mention Uploads.  

Adding a quick unobtrusive note to link the concepts of step uploads and dynamic pipelines would have helped me bridge the understanding gap and resolve my issue so adding one in to help future explorers.

### DIFF
--- a/pages/pipelines/configure/defining_steps.md
+++ b/pages/pipelines/configure/defining_steps.md
@@ -124,6 +124,9 @@ For more example steps and detailed configuration options, see the example `pipe
 
 If your pipeline has more than one step and you have multiple agents available to run them, they will automatically run at the same time. If your steps rely on running in sequence, you can separate them with [wait steps](/docs/pipelines/configure/step-types/wait-step). This will ensure that any steps before the 'wait' are completed before steps after the 'wait' are run.
 
+> 📘
+> It is also possible to upload additional steps while a build is in progress. This isn't a step type, instead it is an action performed by an agent within the context of a job. See [Dynamic pipelines](/docs/pipelines/configure/dynamic_pipelines) for more information.
+
 >🚧 Explicit dependencies in uploaded steps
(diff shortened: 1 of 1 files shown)


pages/pipelines/configure/defining_steps.md
```diff
@@ -124,6 +124,9 @@ For more example steps and detailed configuration options, see the example `pipe
 
 If your pipeline has more than one step and you have multiple agents available to run them, they will automatically run at the same time. If your steps rely on running in sequence, you can separate them with [wait steps](/docs/pipelines/configure/step-types/wait-step). This will ensure that any steps before the 'wait' are completed before steps after the 'wait' are run.
 
+> 📘
+> It is also possible to upload additional steps while a build is in progress. This isn't a step type, instead it is an action performed by an agent or user. See [Dynamic pipelines](/docs/pipelines/configure/dynamic_pipelines.md) for more information.
```
dabarrell: Step uploads are always done by an agent, not by a user

```suggestion
> It is also possible to upload additional steps while a build is in progress. This isn't a step type, instead it is an action performed by an agent within the context of a job. See [Dynamic pipelines](/docs/pipelines/configure/dynamic_pipelines.md) for more information.
```

epssy: Oh that's interesting.  Is it "within" or "with"? Like does it have to be an agent currently running some step or is it any agent running anywhere that is observing the state of a job.

dabarrell: oops - yeah, should be "within". Has to be done using a job token, which is only issued when an agent accepts a job, and is tied to the lifetime of that job

karensawrey:
Thanks so much for rainisng this, @epssy!
Addressed this via a different PR following an internal discussion. Closing this PR.

PR #3154, opened 2026-08-14
[PB-2846] Temporarily hide Cursor Origin GraphQL types

hannahcancode:
## Summary

- temporarily remove the three Cursor Origin types from the checked-in GraphQL schema
- regenerate the GraphQL reference pages and navigation without those types

## Why

This keeps the types out of the public GraphQL documentation while the automatic schema filtering changes are reviewed separately. It does not change the public GraphQL API.

## Testing

- `mise exec -- bundle exec rake graphql:generate`
- `mise exec -- bundle exec rspec spec/scripts/graphql_api_content` (30 examples, 0 failures)


hannahcancode:
@buildsworth-bk

